Issue #28, opened 2020-01-03
index: chain reorgs


jsign:
Decide what is the best strategy to handle chain reorgs when using `ChainNotify`.

Issue #19, opened 2019-12-31
define Reputation API


jsign:
Create some first proposal version of this module.

PR #48, opened 2020-01-09
reputation: index info, reputation module, metrics, and more fixes

### LINKED ISSUE #27: index/miner: metrics
Add metrics for monitoring.

### LINKED ISSUE #30: index/slashing: metrics
Add metrics to monitor this index life.

### LINKED ISSUE #38: index/miners: add region information
Add region information about miners.

### LINKED ISSUE #45: index/miner: add user-agent information
Add user-agent information of libp2p client.
The user-agent information is the name and version of the FC client used.
Should not be trusted blindly. (Like most things you gather from p2p peers)

jsign:
Quite a big PR again. Things evolve and will continue to evolve so review with a grain of salt. I'll comment mostly on what's central in the PR, since other changes are more internal changes/improvements.

---

A new package `fchost` which connects to the private network of Lotus. For the moment, it allows to connect to other libp2p peers in the network, such as miners, resolve in DHT, etc. May serve to run directly protocols if necessary too. For the moment, is leveraged to resolve in DHT and peer metadata explained right below.

---

The miner index now collects two new pieces of information:
- `UserAgent`: its a tag set by libp2p (or custom). Currently, lotus set up with name-version, like `lotus-0.1.6+git716b2a4b`. May serve as a filter criterion if interested in excluding some lotus client versions.
- Resolves the country and latitude/longitude of the miner. This by directly connecting to the DHT and resolving the peerID multiaddrs, and using them in an offline database for resolution. I preferred an offline database to make resolution quick, not depend on third-party apis, and its a free one. It seems to work reasonably well for some tests I made. This feature would allow to clients filter miners from certain countries. Recall that if relays o vpns are used, that's another story.

The database is from `ip2location`. It has multiple versions regarding how much info you want to resolve. Currently, I selected the one with country and lat/long info, it's ~65mb. The one with only countries its much smaller, I think ~15mb (so we may change it). 

Currently, isn't embedded but that might be a good idea to have a single binary for installation (#49).

---

Now all indexes have metrics and were included in grafana. Unfortunately, now the testnet is down so I can't add a screenshot here. But they basically show information about refresh times, gauges with current progress in refreshing indexes (10%, 20%, ..., 100%) since they can take a long time and are hard to imagine what's happening, some summary information in each case (in miners index: how many on-chain, how many online, etc), etc).

--- 

The `reputation` package now has some code that calculates a score for each miner considering all sources of information. This was the main purpose from which the indexes where built; to be sources of information to make decisions or build more abstract data (in this case, scores).

It may also serve as a central point to make rich queries that unify all known information (local or external), eg: _Give me the top 20 miners sorted by price, that have never been slashed, and aren't in China, and happened to be online for at least 10 weeks_. That kind of query is mostly possible now. Yet to know if that's appropriate to do.

---

Unfortunately these last days the testnet is down, so I couldn't test that much some big changes in indexes... so there might be bugs present here. On Monday it will be reset and working again. As soon as its up again, I'll run all the thing again to check if some bugfixes are needed.

---

Enhances #19 
Closes #30  
Closes #27 
Closes #45 
Closes #38 
Towards #28


index/miner/miner.go
```diff
@@ -33,41 +39,51 @@ type API interface {
...
+	Static Static
+	Power  Power
 }
 
-type PowerInfo struct {
-	Power      uint64
-	TotalPower uint64
+type Power struct {
+	Power    uint64
+	Relative float64
```
jsign: `Power`: power of miner in the network
`Relative`: relative power compared to total network power.

index/miner/miner.go
```diff
@@ -5,26 +5,32 @@ import (
 	"encoding/binary"
```
jsign: Summary:
- New static information about miners: user-agent, country, lat/long.
- Refactors/improvements
- Metrics

index/miner/miner.go
```diff
@@ -256,13 +307,43 @@ func (mi *MinerIndex) Close() error {
...
-		TotalPower: mp.TotalPower.Uint64(),
+	p := mp.MinerPower.Uint64()
+	tp := mp.TotalPower
+	return Power{
+		Power:    p,
+		Relative: float64(p) / float64(tp.Uint64()),
 	}, nil
 }
+
+func getStaticInfo(ctx context.Context, c API, h *fchost.FilecoinHost, lr iplocation.LocationResolver, addr string) (Static, error) {
```
jsign: Here's where the static info is gathered.
We connect to the miner byt its peerid. The real addr is resolved by our `FilecoinHost` instance... which is a peer in the FC network connected to the DHT. So all this resolution is magical from this side.
Regarding geo information, is resolved by `LocationResolver` abstraction... which the thing using the offline database for resolving multiaddrs.

index/miner/miner.go
```diff
@@ -256,13 +307,43 @@ func (mi *MinerIndex) Close() error {
...
+	si := Static{}
+	pid, err := c.StateMinerPeerID(ctx, addr, nil)
+	if err != nil {
+		return si, err
+	}
+	_, err = h.NewStream(ctx, pid, identify.ID)
+	if err != nil {
+		return si, err
+	}
+	si.Online = true
```
jsign: If we could connect to the peer, is online.
This is used for the grafana dashboard, but can also be important as a reputation filter... since we could avoid others to fail connecting offline peers.
This may have to be tuned to scale... if the amount of miners is very big, we might relax the info as _last online time_ if we can't hold so many open connections. (just thinking out loud)

index/slashing/metrics.go
```diff
@@ -0,0 +1,39 @@
+package slashing
```
jsign: Metrics for slashing index.

index/slashing/slashing.go
```diff
@@ -2,23 +2,22 @@ package slashing
```
jsign: Summary:
- Metrics
- Refactor/improvements, renames, etc

Most _Consensus_ name referencing was deleted considering yesterday's talk with FC guys. The `slashedAt` is used both for consensus faults and deal faults, so it can't be distinguished. So now its just named _History_.

index/slashing/slashing.go
```diff
@@ -36,52 +35,59 @@ type SlashingIndex struct {
...
-	LastUpdated uint64
-	// History has slashed height History of storage miners
-	History map[string][]uint64
+type Index struct {
+	Tipset types.TipSetKey
+	Miners map[string]Info
+}
+
+type Info struct {
+	History []uint64
```
jsign: `Height` was changed to `Tipset`.

The thing here is that `Height` isn't a correct description of the point we last updated the index, since the chain may fork. So, if we say `Last sync was at height XXX` we can't guarantee that is the same tipset of the current chain (if it was reorged by some heavier fork).

So... the correct description is the fingerprint of the tipset. This will become more apparent when #28  is done since is quite important. Is related with the talk with FC guys about `ChainNotify` receiving a tipset checkpoint.

This has some big implications... since if when the index resumes syncing it discovers that the last tipset isn't part of the current chain (because it was reorged), we should kind of do a full refresh. A better option is having some checkpoints of index state at lower heights to rebuild from there. I have some things in mind, but that's further work.

iplocation/ip2location/iplocation.go
```diff
@@ -0,0 +1,61 @@
+package ip2location
```
jsign: The thing we use to resolve location from ip addrs.
TBH, its quite simple. The client SDK is very poor in design: it uses globals and that sorts of things... but it works.

Clients rely on an interface that satisfies `Resolve(mas []multiaddr.Multiaddr) (iplocation.Location, error)`, so we can change this to other implementation or whatever easily if necessary.

iplocation/iplocation.go
```diff
@@ -0,0 +1,21 @@
+package iplocation
```
jsign: The package that clients depend upon. It serves as an abstraction if we want to change the resolving underlying library or similar.

lotus/client.go
```diff
@@ -23,24 +24,25 @@ var (
 
 type API struct {
```
jsign: New APIs that call Lotus.

lotus/client.go
```diff
@@ -77,58 +79,137 @@ func New(maddr ma.Multiaddr, authToken string) (*API, func(), error) {
 }
```
jsign: Decided to add logging at this level, to have a better understanding on errs in packages. As a boundary between lotus client messages and our code.

lotus/jsonrpc/handler.go
```diff
@@ -58,63 +58,11 @@ type response struct {
 	Error   *respError  `json:"error,omitempty"`
```
jsign: Unused code.

tests/auth.go
```diff
@@ -45,14 +45,17 @@ func ClientConfigMA() (ma.Multiaddr, string) {
...
 	}
-	home, err := os.UserHomeDir()
-	if err != nil {
-		panic(err)
-	}
-	path := filepath.Join(home, ".lotus")
-	authToken, err := GetLotusToken(path)
-	if err != nil {
-		panic(err)
+	authToken, ok := os.LookupEnv("TEXTILE_LOTUS_TOKEN")
```
jsign: Also looks at an env var, used in docker-compose. 
We might have a more clear policy to name env vars. For the moment, choose this.

reputation/reputation.go
```diff
@@ -2,12 +2,18 @@ package reputation
```
jsign: This package now leverage the work of the created indexes.

reputation/reputation.go
```diff
@@ -19,39 +25,165 @@ type ReputationModule struct {
...
+	go rm.indexBuilder()
 
 	return rm
 }
 
+func (rm *ReputationModule) AddSource(id string, maddr ma.Multiaddr) error {
+	return rm.sources.Add(source.Source{Id: id, Maddr: maddr})
+}
+
+func (rm *ReputationModule) GetTopMiners(n int) ([]MinerScore, error) {
```
jsign: For the moment, a method that returns the top N miners with best score.
We could make more rich queries, etc. Limited by imagination.

reputation/reputation.go
```diff
@@ -19,39 +25,165 @@ type ReputationModule struct {
...
+}
+
 func (rm *ReputationModule) Close() error {
 	rm.cancel()
 	return nil
 }
 
-func (rm *ReputationModule) AddSource(id string, maddr ma.Multiaddr) error {
-	return rm.sources.Add(source.Source{Id: id, Maddr: maddr})
+func (rm *ReputationModule) subscribeIndexes() {
```
jsign: This is a long running goroutine that listen for all indexes.
When some index say _Hey, I have new info_, it signals to recompute scores.

reputation/reputation.go
```diff
@@ -19,39 +25,165 @@ type ReputationModule struct {
...
+			return
+		}
+		rm.lockIndex.Lock()
+		minerIndex := rm.mIndex
+		slashIndex := rm.sIndex
+		askIndex := rm.aIndex
+		rm.lockIndex.Unlock()
+
+		scores := make([]MinerScore, 0, len(minerIndex.Miners))
+		for addr := range minerIndex.Miners {
```
jsign: For each miner, compute score.

reputation/reputation.go
```diff
@@ -19,39 +25,165 @@ type ReputationModule struct {
...
+			score := calculateScore(addr, minerIndex, slashIndex, askIndex, sources)
+			scores = append(scores, score)
+		}
+		sort.Slice(scores, func(i, j int) bool {
```
jsign: Sort them by score.

reputation/reputation.go
```diff
@@ -19,39 +25,165 @@ type ReputationModule struct {
...
+	var askScore float64
+	if a, ok := ai.Miners[addr]; ok && a.Price < ai.MedianPrice {
+		askScore = 1
+	}
+
+	score := 50*slashScore + 20*powerScore + 20*externalScore + 10*askScore
+	return MinerScore{
+		Addr:  addr,
+		Score: int(score),
+	}
```
jsign: Ok, this is just an attempt to do _something_. It doesn't mean is a meaningful tuned thing.
It uses information from the different indexes to compute a weighted average to come up with a number:
`score := 50*slashScore + 20*powerScore + 20*externalScore + 10*askScore`
In that formula, slashes have more weight than a good ask score, etc.

Pretty hard to know now what would be a good formula. The main idea is showing that all this reputation logic can be abstracted from all the Lotus logic and just gather all information availabe in local indexes and external `Source`s, to come up with some sorted list.

jsign: There're more information in indexes, and it will have more the future. I just used something here in the formulas.

index/ask/ask.go
```diff
@@ -95,130 +118,153 @@ func (d *AskIndex) AvailableAsks(q Query) ([]StorageAsk, error) {
...
-
-	var buf bytes.Buffer
-	encoder := json.NewEncoder(&buf)
-	for _, ask := range asks {
-		buf.Reset()
-		if err := encoder.Encode(ask); err != nil {
+	for _, ask := range ai.index.Miners {
+		b, err := cbor.DumpObject(ask)
+		if err != nil {
+			panic(err)
```
sanderpick: panic intended?

reputation/reputation.go
sanderpick: Slick

reputation/reputation.go
sanderpick: 🎉

reputation/reputation.go
```diff
@@ -19,39 +25,165 @@ type ReputationModule struct {
...
+}
+
+func calculateScore(addr string, mi miner.Index, si slashing.Index, ai ask.Index, ss []source.Source) MinerScore {
+	minfo := mi.Miners[addr]
+	powerScore := minfo.Power.Relative
+
+	var slashScore float64
+	if si, ok := si.Miners[addr]; ok {
+		slashScore = 1 / math.Pow(2, float64(len(si.History)))
```
sanderpick: So, slash score non-linearly decreases with count... harsh! :) I have zero intuitions here.

reputation/reputation.go
sanderpick: Sounds like a great starting place... maybe PL will have some ideas about what makes for a good score algo.

index/ask/ask.go
jsign: Ups, nope. Will change.

reputation/reputation.go
jsign: Yep, it feels to me that shouldn't be linear.
I think a better thing would be to also consider how many deals has on-chain.
If a miner has 3 deals, or 10000 deals... feels like it should have to be normalized or something similar. 
So maybe first normalize, then being exponentially penalized :P

jsign:
Thanks @sanderpick  for the review :+1:, made the panic change.

reputation/reputation.go
sanderpick: Ah yeah, makes sense.

PR #60, opened 2020-01-20
Deals and Wallet CLI

asutula:
Adds cli for deals and wallet related things. It currently runs against a stub version of the filecoin client. ~~Was spinning my wheels a bit figuring out the correct way to integrate Viper config and settled on putting all viper bindings in the root command. That might be heavy handed, but it works for now. Things get weird if you bind different command flags to the same Viper value, so this avoids that situation all together.~~


api/client/client.go
```diff
@@ -16,13 +15,14 @@ type Client struct {
...
+func NewClient(target string, creds credentials.TransportCredentials) (*Client, error) {
+	var opts []grpc.DialOption
+	if creds != nil {
+		opts = append(opts, grpc.WithTransportCredentials(creds))
+	} else {
+		opts = append(opts, grpc.WithInsecure())
 	}
-	// ToDo: Support secure connection
-	conn, err := grpc.Dial(addr, grpc.WithInsecure())
+	conn, err := grpc.Dial(target, opts...)
```
asutula: Moved the grpc connection code to the way @sanderpick started doing it in `textile`.

api/server/server.go
```diff
@@ -44,7 +43,8 @@ type Server struct {
 type Config struct {
 	LotusAddress    ma.Multiaddr
 	LotusAuthToken  string
-	GrpcHostAddress ma.Multiaddr
+	GrpcHostNetwork string
+	GrpcHostAddress string
```
asutula: Just decided to make this configurable because you can actually set it up to use uds or other things for testing etc

api/server/server.go
```diff
@@ -44,7 +43,8 @@ type Server struct {
 type Config struct {
 	LotusAddress    ma.Multiaddr
 	LotusAuthToken  string
-	GrpcHostAddress ma.Multiaddr
+	GrpcHostNetwork string
```
asutula: Just decided to make this configurable because you can actually set it up to use uds or other things for testing etc

exe/cli/cmd/deal.go
```diff
@@ -0,0 +1,169 @@
...
+
+func init() {
+	dealCmd.Flags().StringP("file", "f", "", "Path to the file to store")
+	dealCmd.Flags().IntP("limit", "l", -1, "limit the number of results")
+	dealCmd.Flags().IntP("offset", "o", -1, "offset of results")
+
+	rootCmd.AddCommand(dealCmd)
+}
+
+var dealCmd = &cobra.Command{
```
asutula: This is the fully interactive deal ux

exe/cli/cmd/deals.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(dealsCmd)
+}
+
+var dealsCmd = &cobra.Command{
```
asutula: The sub commands of this one are the individual calls to use the different parts of the deals module (asks, store, watch)

exe/cli/cmd/root.go
```diff
@@ -0,0 +1,98 @@
...
+	err = viper.BindPFlag("pieceSize", rootCmd.PersistentFlags().Lookup("pieceSize"))
+	checkErr(err)
+	err = viper.BindPFlag("wallets", rootCmd.PersistentFlags().Lookup("wallets"))
+	checkErr(err)
+
+	viper.SetDefault("address", "")
+	viper.SetDefault("duration", 100)
+	viper.SetDefault("maxPrice", 200)
+	viper.SetDefault("pieceSize", 1024)
+	viper.SetDefault("wallets", []string{})
```
asutula: Moved all viper bound flags into the root command here and made them persistent flags. Might be a better way to do this.

asutula: Looks like the confusion about this is documented here https://github.com/spf13/viper/issues/233

A good solution is to bind flags to viper using `PreRun` or `RunIn`.

asutula: Went ahead and updated to do all viper binding in `PreRun`... much nicer this way.

exe/cli/cmd/deal.go
```diff
@@ -0,0 +1,174 @@
...
+		for i, ask := range asks {
+			choices[i] = fmt.Sprintf("Price of %v with min piece size %v expiring %v", ask.Price, ask.MinPieceSize, ask.Expiry)
+		}
+
+		selectedAsks := []int{}
+		prompt := &survey.MultiSelect{
+			Message: "Select from available asks:",
+			Options: choices,
+		}
+		survey.AskOne(prompt, &selectedAsks, survey.WithValidator(survey.Required))
```
jsign: Cool library.

exe/cli/cmd/deal.go
```diff
@@ -0,0 +1,174 @@
...
+
+		watchCtx, watchCancel := context.WithTimeout(context.Background(), cmdTimeout)
+		defer watchCancel()
+
+		ch, err := fcClient.Deals.Watch(watchCtx, succeeded)
+		checkErr(err)
+
+		for {
+			event, ok := <-ch
+			if ok == false {
```
jsign: Nit: `!ok`.
Also, I wonder if this should also output some error message since isn't expected (being a timeout or an error)

exe/cli/cmd/deals_store.go
```diff
@@ -0,0 +1,106 @@
...
+	storeCmd.Flags().StringP("file", "f", "", "Path to the file to store")
+	storeCmd.Flags().StringSliceP("miners", "m", []string{}, "miner ids of the deals to execute")
+	storeCmd.Flags().IntSliceP("prices", "p", []int{}, "prices of the deals to execute")
+	storeCmd.Flags().StringP("address", "a", "", "wallet address used to store the data")
+	storeCmd.Flags().Uint64P("duration", "d", 0, "duration to store the data for")
+
+	dealsCmd.AddCommand(storeCmd)
+}
+
+var storeCmd = &cobra.Command{
```
jsign: So this command is similar to the fully UX one, but not interactive?

exe/cli/cmd/deals_watch.go
```diff
@@ -0,0 +1,125 @@
...
+// var DealStates = []string{
+// 	"DealUnknown",
+// 	"DealRejected",
+// 	"DealAccepted",
+// 	"DealStaged",
+// 	"DealSealing",
+// 	"DealFailed",
+// 	"DealComplete",
+// 	"DealError",
+// }
```
jsign: Nit: delete?

exe/cli/cmd/deals_watch.go
```diff
@@ -0,0 +1,125 @@
...
+			event, ok := <-ch
+			if ok == false {
+				break
+			}
+			state[event.Deal.ProposalCid.String()] = &event
+			updateOutput(writer, state)
+			if isComplete(state) {
+				break
+			}
+		}
```
jsign: Maybe this code could be extracted and be used by other commands that call `Watch`, like in the interactive cmd.
If that doesn't sound like may worth it; same comments as in the other case (`!ok` and error msg)

exe/cli/cmd/deals_watch.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func isComplete(state map[string]*client.WatchEvent) bool {
+	for _, value := range state {
+		processing := false
+		if value == nil ||
+			value.Deal.StateName == "DealUnknown" ||
+			value.Deal.StateName == "DealAccepted" ||
+			value.Deal.StateName == "DealStaged" ||
+			value.Deal.StateName == "DealSealing" {
```
jsign: Would be better to use constants defined at the API level?

exe/cli/cmd/wallet_balance.go
```diff
@@ -0,0 +1,45 @@
...
+func init() {
+	balanceCmd.Flags().StringP("address", "a", "", "wallet address to check the balance of")
+
+	walletCmd.AddCommand(balanceCmd)
+}
+
+var balanceCmd = &cobra.Command{
+	Use:   "balance",
+	Short: "Print the balance of the specified wallet",
+	Long:  `Print the balance of the specified wallet`,
```
jsign: s/`specified wallet`/`specified wallet address`?

exe/cli/cmd/wallet_new.go
```diff
@@ -0,0 +1,61 @@
...
+	newCmd.Flags().StringP("type", "t", "bls", "specifies the wallet type, either bls or secp256k1. Defaults to bls.")
+	newCmd.Flags().BoolP("switch", "s", false, "whether or not to switch to the newly created wallet")
+
+	walletCmd.AddCommand(newCmd)
+}
+
+var newCmd = &cobra.Command{
+	Use:   "new",
+	Short: "Create a new filecoin wallet",
+	Long:  `Create a new filecoin wallet`,
```
jsign: `s/`wallet`/`wallet address`

Or maybe we should say simply _address_ and avoid _wallet_?
Or maybe wallet is OK despite isn't what is happening under the hoods. A single-address wallet.

api/client/client.go
sanderpick: Cool! In the threads client(s), I started just passing raw opts to give the user max flexibility... not sure if needed here.

exe/cli/cmd/helpers.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		v.SetEnvPrefix("TXTL")
+		v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+		v.AutomaticEnv()
+		_ = v.ReadInConfig()
+	}
+}
+
+func BindFlags(v *viper.Viper, root *cobra.Command, flags map[string]Flag) error {
```
sanderpick: I don't see this being used... should we wire it up?

exe/cli/cmd/root.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.filecoin.yaml)")
+	rootCmd.PersistentFlags().String("serverAddress", "127.0.0.1:5002", "address of the filecoin service api")
+}
+
+func initConfig() {
```
sanderpick: Dup from helpers?

exe/cli/cmd/deals_asks.go
```diff
@@ -0,0 +1,82 @@
...
+			for i, a := range asks {
+				data[i] = []string{
+					a.Miner,
+					strconv.Itoa(int(a.Price)),
+					strconv.Itoa(int(a.MinPieceSize)),
+					strconv.Itoa(int(a.Timestamp)),
+					strconv.Itoa(int(a.Expiry)),
+				}
+			}
+			RenderTable(os.Stdout, []string{"miner", "price", "min piece size", "timestamp", "expiry"}, data)
```
andrewxhill: this seems to need a line return before rendering the table

Screen Shot 2020-01-21 at 10 51 38 AM

api/client/client.go
asutula: Makes sense, will do that here

exe/cli/cmd/deal.go
asutula: Cool, fixed the nit. `!ok` isn't necessarily and error, right? It happens when the channel is closed because no more data will be sent over it.

asutula: Also, this is a channel of `client.WatchEvent`:
```
type WatchEvent struct {
	Deal deals.DealInfo
	Err  error
}
```
So we explicitly pass any error along as part of a channel event.

asutula: In the case of an error coming through in a channel event, the error is displayed in the cli output next to the corresponding deal id.

exe/cli/cmd/deals_asks.go
asutula: Strange, works for me. The previous output of `Querying network for available storage asks...` should also disappear. Do you get the same result when running from go?
```
go run exe/cli/main.go deals asks
```

exe/cli/cmd/deals_store.go
asutula: Right. It is meant to be used stand alone. In theory, you'd use the other stand alone commands (`deals asks` and `deals watch`) before and after `deals store`.

exe/cli/cmd/deals_watch.go
asutula: Will do

exe/cli/cmd/deals_watch.go
asutula: That would be better. I guess we'd do that at the deal module level?

exe/cli/cmd/deals_watch.go
asutula: Extracted it.

exe/cli/cmd/helpers.go
asutula: Copy pasta. Not sure if I want to wire it up. I've been explicit about what flags are defined where and bind viper as needed. Please correct me if I'm wrong, but my reasoning is that using something like this `BindFlags` tends to cause you to overload the root command with persistent flags that don't actually all apply to every child command. My goal was to keep the auto generated command help accurate in the sense that global persistent flags actually apply to every child and that child flags accurately represent the flags available for that command.

asutula: I'll delete the things in this file that aren't being used.

exe/cli/cmd/root.go
asutula: Yep. Will de-dup.

exe/cli/cmd/deals_watch.go
asutula: Created this issue to track it https://github.com/textileio/fil-tools/issues/71

exe/cli/cmd/deal.go
asutula: I think our explicit handling of any error in the event is good, going to resolve this conversation. Feel free to chime in again if you don't agree.

asutula:
@sanderpick re: the `exe/cli` directory structure... Yea I'm open to doing this however makes sense. My thinking with the way it is now is that we don't just have one executable program, there are two (daemon and cli). I wanted to use the `cmd` dir and package for the actual cli command files. But whatever, I can change all that.

exe/cli/cmd/deal.go
jsign: > Cool, fixed the nit. `!ok` isn't necessarily and error, right? It happens when the channel is closed because no more data will be sent over it.

Oh ok.. if the items are that struct, makes sense.

jsign: If that's the case you can use `for e := range `, so it terminates automatically on channel close.

exe/cli/cmd/root.go
asutula: For now I just went with the existing code that is being used. It's simple. Removed all the duplicated, unused code here and made an issue to look at implementing a more robust config file location mechanism in the future if we want https://github.com/textileio/fil-tools/issues/72

exe/cli/cmd/deals_asks.go
asutula: Also works for me when using the `go install`ed version. What shell are you using?

andrewxhill: Screen Shot 2020-01-28 at 2 24 58 PM

lgtm

PR #66, opened 2020-01-22
fixes & improvements

### LINKED ISSUE #49: iplocation: consider embedding database

jsign:
A bunch of fixes and improvements based on running the dashboard, not much of this bugs tracked since discovered and fixed right away. Some more details in the comments.

The only tracked one was Closes #49 , which seems no to be necessary since I'm copying it in the Dockerfile, so it's kind of embedded (not in the binary, but in the running image).

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -84,7 +84,7 @@ func NewServer(conf Config) (*Server, error) {
 	}
 	dm := deals.New(txndstr.Wrap(ds, "dealmodule"), c)
 
-	ip2l := ip2location.New([]string{"./ip2location-ip4.bin"}) // ToDo: Flags or embed
+	ip2l := ip2location.New([]string{"./ip2location-ip4.bin"})
 	mi, err := miner.New(txndstr.Wrap(ds, "index/miner"), c, fchost, ip2l)
 	if err != nil {
 		return nil, fmt.Errorf("error when creating miner index: %s", err)
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -37,12 +37,13 @@ var (
 
 // Server represents the configured lotus client and filecoin grpc server
 type Server struct {
-	ds datastore.TxnDatastore
-	dm *deals.Module
-	ai *ask.AskIndex
-	wm *wallet.Module
-	si *slashing.SlashingIndex
-	mi *miner.MinerIndex
+	ds   datastore.TxnDatastore
+	dm   *deals.Module
+	ai   *ask.AskIndex
+	wm   *wallet.Module
+	si   *slashing.SlashingIndex
+	mi   *miner.MinerIndex
+	ip2l *ip2location.IP2Location
 
 	rpc           *grpc.Server
 	dealsService  *deals.Service
@@ -83,7 +84,7 @@ func NewServer(conf Config) (*Server, error) {
 	}
 	dm := deals.New(txndstr.Wrap(ds, "dealmodule"), c)
 
-	ip2l := ip2location.New([]string{"ip2location-ip4.bin"}) // ToDo: Flags or embed
+	ip2l := ip2location.New([]string{"./ip2location-ip4.bin"}) // ToDo: Flags or embed
 	mi, err := miner.New(txndstr.Wrap(ds, "index/miner"), c, fchost, ip2l)
 	if err != nil {
 		return nil, fmt.Errorf("error when creating miner index: %s", err)
@@ -112,6 +113,7 @@ func NewServer(conf Config) (*Server, error) {
 		wm:            wm,
 		mi:            mi,
 		si:            si,
+		ip2l:          ip2l,
 		dealsService:  dealsService,
 		walletService: walletService,
 		closeLotus:    cls,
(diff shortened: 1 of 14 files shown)


api/server/server.go
```diff
@@ -83,7 +84,7 @@ func NewServer(conf Config) (*Server, error) {
 	}
 	dm := deals.New(txndstr.Wrap(ds, "dealmodule"), c)
 
-	ip2l := ip2location.New([]string{"ip2location-ip4.bin"}) // ToDo: Flags or embed
+	ip2l := ip2location.New([]string{"./ip2location-ip4.bin"})
```
jsign: The _ip2location-ip4.bin_ file is `COPY`ed in the Dockerfile in the same directory that the entroypoint.

api/server/server.go
```diff
@@ -184,4 +186,5 @@ func (s *Server) Close() {
 		log.Errorf("error when closing datastore: %s", err)
 	}
 	s.closeLotus()
+	s.ip2l.Close()
```
jsign: Should close the `ip2l` service.

docker/Makefile
```diff
@@ -1,4 +1,5 @@
 LT = $(shell cat ~/.lotus/token)
+#LT = $(shell docker exec lotustest cat /root/.lotus/token)
```
jsign: Leaved this since in remote instance the Lotus node is running in docker.

docker/grafana/provisioning/dashboards/TextileFC.json
```diff
@@ -513,17 +513,17 @@
         "showThresholdLabels": false,
```
jsign: Grafana fixes/improvements.

fchost/fchost.go
```diff
@@ -32,7 +36,7 @@ func New() (*FilecoinHost, error) {
 		return nil, err
 	}
 
-	dht, err := dht.New(ctx, h)
+	dht, err := dht.New(ctx, h, dhtopts.Protocols("/lotus/kad/1.0.0"))
```
jsign: Seems that now that libp2p hosts aren't running in a private network as before, they have their ProtocolID for DHT.
Took some time to figure out why DHT wasn't working...

fchost/fchost.go
```diff
@@ -19,8 +22,9 @@ var (
 
 // FilecoinHost is a libp2p host connected to the FC network
 type FilecoinHost struct {
-	host.Host
-	dht *dht.IpfsDHT
+	ping *ping.PingService
+	h    host.Host
+	dht  *dht.IpfsDHT
```
jsign: `host.Host` isn't embedded anymore, so less power for users of `FilecoinHost`.
Preferred this way and implement as public methods whatever they need.

fchost/fchost.go
```diff
@@ -57,6 +63,25 @@ func (fh *FilecoinHost) Bootstrap() error {
 	return nil
 }
 
+func (fc *FilecoinHost) Ping(ctx context.Context, pid peer.ID) bool {
```
jsign: Used so the Metadata Index of Miners can discover if a miner responds to the Ping protocol, to decide if it's online or not.

fchost/fchost.go
```diff
@@ -57,6 +63,25 @@ func (fh *FilecoinHost) Bootstrap() error {
...
+	r := <-fc.ping.Ping(ctx, pid)
+	return r.Error == nil
+}
+
+func (fc *FilecoinHost) GetAgentVersion(pid peer.ID) string {
```
jsign: Abstracted AgentVersion fetching info. 
In case we're interested in an extra thing, it could receive the key as param and be generic.

index/miner/meta.go
```diff
@@ -2,11 +2,12 @@ package miner
 
 import (
```
jsign: Mostly enabling some capped parallelization to discover metadata about miners. Capped at 100 max at a time, could be played bit.

index/miner/metrics.go
```diff
@@ -7,15 +7,22 @@ import (
 )
 
 var (
```
jsign: Yet some more accurate renamings.

index/slashing/slashing.go
```diff
@@ -243,9 +243,8 @@ func epochPatch(ctx context.Context, c API, pts *types.TipSet, ts *types.TipSet)
 			}
 			mas, ok := as.State.(map[string]interface{})
 			if !ok {
-				panic("read state should be a map interface result")
+				log.Debugf("read state should be a map interface result: %#v", as.State)
```
jsign: For some reason this case started happening. Opted to not be fatal, log, and move on.
Basically is returning `nil`, but not a typed `map[string]interface{}` nil or empty thing.

iplocation/ip2location/ip2location.go
```diff
@@ -41,18 +45,25 @@ func New(paths []string) *IP2Location {
...
+			continue
+		}
+		ip, _, err := net.SplitHostPort(ipport)
 		if err != nil {
+			log.Debugf("error parsing ip/port from %s: %s", ipport, err)
 			continue
 		}
 		r := ip2location.Get_all(ip)
-		if r.Country_long != "" && r.Latitude != 0 && r.Longitude != 0 {
+		if r.Country_long != "" || (r.Latitude != 0 && r.Longitude != 0) {
```
jsign: Any of the country or coords should be enough to return something.

PR #68, opened 2020-01-24
jsonrpc autoconnect, indexes client recovery, docker volumes, bugfixes

### LINKED ISSUE #41: lotus/client: reconnect on failure
If the connection with the lotus client fails (e.g: lotus is restarted), things should keep moving forward retrying or similar.

### LINKED ISSUE #67: docker: user map repo to volume

jsign:
While running our reputation index, which is a long-running thing, the lotus client was kind of fatal when the websocket connection was lost. This was a known thing for a while in the `jsonrpc` from lotus, so I jumped to help us both in that (https://github.com/filecoin-project/lotus/pull/1155).
The current solution that is in this PR, is more in line with a previous PR than the mentioned before... so it may change again with my new solution, but the current one is enough for our interests.

While installing and leaving running the service in the cloud instance, I discovered a set of issues or needed improvements to keep it more: reliable on lotus or self crashes, tested the persistence of indexed data and recovery after restarting, and some extra tunes to dockerized things that run on the cloud (and I also use locally to test things, since helps fast iteration with the many wires to connect).

I'll make some comments, but that's pretty much the summary of this.

Closes #41 
Closes #67

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -1,5 +1,8 @@
 version: '3.7'
 
+volumes:
+  textile-fc-data:
+
 services:
   textile-fc:
     network_mode: host
@@ -10,7 +13,10 @@ services:
       - TEXTILE_LOTUS_TOKEN=${TEXTILE_LOTUS_TOKEN}
     ports:
       - 8888:8888
+      - 6060:6060
     restart: always
+    volumes:
+      - textile-fc-data:/root/.texfc
     
   prometheus:
     network_mode: host
--- a/docker/grafana/provisioning/dashboards/TextileFC.json
+++ b/docker/grafana/provisioning/dashboards/TextileFC.json
@@ -523,7 +523,7 @@
       ],
       "timeFrom": null,
       "timeShift": null,
-      "title": "On-Chain Progress",
+      "title": "On-Chain Refresh",
       "type": "gauge"
     },
     {
@@ -597,7 +597,7 @@
       "thresholds": "",
       "timeFrom": null,
       "timeShift": null,
-      "title": "Height",
+      "title": "Index Height",
       "type": "singlestat",
       "valueFontSize": "80%",
       "valueMaps": [
@@ -684,7 +684,7 @@
       "thresholds": "",
       "timeFrom": null,
       "timeShift": null,
-      "title": "Avg Full-Refresh Duration",
+      "title": "Full-Refresh Duration",
       "type": "singlestat",
       "valueFontSize": "80%",
       "valueMaps": [
@@ -768,7 +768,7 @@
       "thresholds": "",
       "timeFrom": null,
       "timeShift": null,
-      "title": "Avg Delta-Refresh Duration",
+      "title": "Delta-Refresh Duration",
       "type": "singlestat",
       "valueFontSize": "80%",
       "valueMaps": [
@@ -828,7 +828,7 @@
       ],
       "timeFrom": null,
       "timeShift": null,
-      "title": "Meta Progress",
+      "title": "Meta Refresh",
       "type": "gauge"
     },
     {
@@ -884,7 +884,7 @@
       ],
       "timeFrom": null,
       "timeShift": null,
-      "title": "Current #Miners on-chain",
+      "title": "#Miners Ping Reply",
       "type": "bargauge"
     },
     {
@@ -948,7 +948,7 @@
       ],
       "timeFrom": null,
       "timeShift": null,
-      "title": "Full refresh progress",
+      "title": "Full Refresh",
       "type": "gauge"
     },
     {
@@ -1022,7 +1022,7 @@
       "thresholds": "",
       "timeFrom": null,
       "timeShift": null,
-      "title": "Last Refresh Duration",
+      "title": "Refresh Duration",
       "type": "singlestat",
       "valueFontSize": "80%",
       "valueMaps": [
@@ -1032,7 +1032,7 @@
           "value": "null"
         }
       ],
-      "valueName": "current"
+      "valueName": "avg"
     },
     {
       "cacheTimeout": null,
@@ -1082,7 +1082,7 @@
       ],
       "timeFrom": null,
       "timeShift": null,
-      "title": "Last Query-Ask Response",
+      "title": "Last Query-Ask Reply",
       "type": "bargauge"
     },
     {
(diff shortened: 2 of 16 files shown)


cmd/server/main.go
```diff
@@ -78,3 +81,9 @@ func instrumentationSetup() {
 		}
 	}()
 }
+
+func pprofSetup() {
```
jsign: Enabled pprof. I wanted to inspect some things about a memprofile of the service. 
All good. Left it here anyways, is binded to `localhost` so not public.

docker/docker-compose.yaml
```diff
@@ -10,7 +13,10 @@ services:
       - TEXTILE_LOTUS_TOKEN=${TEXTILE_LOTUS_TOKEN}
     ports:
       - 8888:8888
+      - 6060:6060
     restart: always
+    volumes:
+      - textile-fc-data:/root/.texfc
```
jsign: Made a change in the docker-compose used to launch the server, prometheus, grafana, etc.
The `textile-fc` (our service) uses a volume to map the repo folder to a Docker volume. Mostly for saving progress while restarting the compose environment, and tests index persistence.

docker/grafana/provisioning/dashboards/TextileFC.json
```diff
@@ -523,7 +523,7 @@
       ],
```
jsign: Some improvements in clarity and new boxes that helped discover some things.

index/miner/meta.go
```diff
@@ -151,33 +150,3 @@ func (mi *MinerIndex) persistMetaIndex(index MetaIndex) error {
 	}
 	return nil
 }
-
```
jsign: Moved to `miners.go` since loading from store is a general thing.

index/miner/miner.go
```diff
@@ -153,10 +154,11 @@ func (mi *MinerIndex) Close() error {
 // miners.
 func (mi *MinerIndex) start() {
 	defer func() { mi.finished <- struct{}{} }()
-	n, err := mi.api.ChainNotify(mi.ctx)
```
jsign: Stop using `ChainNotify` for _listening_ to the chain.
I've been thinking about this for quite some time since puts a lot of stress. It notifies about every possible forks that might happen in 10-20s time frame... we're only interested in the heaviest tipset known and let time (and lotus) decide which are the heaviest ones.

Switched to waiting the avg tipset time and just asking for the heaviest current tipset (which is what we need). Mini forks or possible future forks aren't interesting.  We don't need to be at the bleeding-edge of that.

index/miner/onchain.go
```diff
@@ -12,28 +12,48 @@ import (
...
+	hOffset = uint64(5)
 )
 
-// updateOnChainIndex updates on-chain index information when a new tipset
-func (mi *MinerIndex) updateOnChainIndex(new *types.TipSet) error {
+// updateOnChainIndex updates on-chain index information in the direction of heaviest tipset
+// with some height offset to reduce sensibility to reorgs
+func (mi *MinerIndex) updateOnChainIndex() error {
 	log.Info("updating on-chain index...")
+	heaviest, err := mi.api.ChainHead(mi.ctx)
```
jsign: Since we don't use `ChainNotify` anymore, we call `ChainHead` which returns the current accepted heaviest tipset known (what's considered _the truth_ now).

index/miner/onchain.go
```diff
@@ -12,28 +12,48 @@ import (
...
 	ts, err := mi.store.LoadAndPrune(mi.ctx, newtsk, &chainIndex)
 	if err != nil {
-		return err
+		return fmt.Errorf("error getting last saved from store: %s", err)
 	}
 	if chainIndex.Power == nil {
 		chainIndex.Power = make(map[string]Power)
 	}
+	hdiff := new.Height - chainIndex.LastUpdated
+	if hdiff == 0 {
```
jsign: A simple check to avoid doing some unnecessary work if the heaviest things is the same as the last known one.

index/slashing/slashing.go
```diff
@@ -123,42 +129,45 @@ func (s *SlashingIndex) Close() error {
 // start is a long running job that keeps the index up to date with chain updates
 func (s *SlashingIndex) start() {
 	defer close(s.finished)
-	n, err := s.api.ChainNotify(s.ctx)
```
jsign: Same idea of changing `ChainNotify` with `ChainHead` in this index.

index/slashing/slashing.go
```diff
@@ -200,15 +210,12 @@ func updateFromPath(ctx context.Context, api API, index *Index, path []*types.Ti
...
-			if !ok {
-				info = Slashes{}
-				index.Miners[addr] = info
-			}
+			info := index.Miners[addr]
 			if len(info.Epochs) > 0 && info.Epochs[len(info.Epochs)-1] == patch[addr] {
 				continue
 			}
 			info.Epochs = append(info.Epochs, patch[addr])
+			index.Miners[addr] = info
```
jsign: Bug fix. `info` was edited but not assigned to current index (so lost changes).

lotus/client.go
```diff
@@ -61,7 +61,7 @@ func New(maddr ma.Multiaddr, authToken string) (*API, func(), error) {
 	closer, err := jsonrpc.NewMergeClient("ws://"+addr+"/rpc/v0", "Filecoin",
 		[]interface{}{
 			&api.Internal,
-		}, headers)
+		}, headers, jsonrpc.WithReconnect(true, time.Second*3, 0))
```
jsign: Original PR that I did for lotus. You can configure a interval time to wait so that the websocket connection is recovered, adn the client can still use it as "nothing happened".

In the newer PR that will get merged, you can annotate with `retry:"true"` methods of the API. If that annotation is detected, the client retries itself the call until having success on a recovered connection. This makes the client even do less work since shouldn't do any kind of retry logic.

I may eventually consider switching to the new solution.

lotus/client.go
```diff
@@ -242,16 +242,10 @@ func monitorLotusSync(ctx context.Context, c *API) {
 }
 
 func refreshHeightMetric(ctx context.Context, c *API) {
-	var h uint64
-	state, err := c.SyncState(ctx)
+	heaviest, err := c.ChainHead(ctx)
```
jsign: Simplify `curent height` metric in grafana dashboard to the current heaviest tipset... not the newer _potential_ new tipset that we discovered recently.

lotus/jsonrpc/client.go
```diff
@@ -85,12 +85,21 @@ type client struct {
```
jsign: `lotus/jsonrpc` package is from Lotus. Has the changes proposed for autoconnect feature.
Can be dismissed for review... but if like crazy things might be fun.

util/util.go
```diff
@@ -2,10 +2,15 @@ package util
 
 import (
 	"fmt"
+	"time"
 
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+const (
+	AvgBlockTime = time.Second * 30
```
jsign: In testnet is 45s. But in mainet will be 30s as mentioned in the spec.
Leaving 30s here since updating more often doesn't hurt; worst thing that can happen is doing nothing.

jsign:
Will DCO after review just in case that might not invalidate review comments.

index/miner/meta.go
```diff
@@ -50,6 +49,7 @@ func (mi *MinerIndex) metaWorker() {
 			newIndex, err := updateMetaIndex(mi.ctx, mi.api, mi.h, mi.lr, addrs)
 			if err != nil {
 				log.Errorf("error when updating meta index: %s", err)
+				break
```
jsign: If for some reason generating a new index fails, log the error and break here. If we continue we would override the current-good-state-index with an empty one and lost things. Better to keep it outdated than to lose it.

index/miner/miner.go
```diff
@@ -100,7 +101,9 @@ func (mi *MinerIndex) Get() Index {
 	defer mi.lock.Unlock()
 	ii := Index{
 		Meta: MetaIndex{
-			Info: make(map[string]Meta, len(mi.index.Meta.Info)),
+			Online:  mi.index.Meta.Online,
+			Offline: mi.index.Meta.Offline,
```
jsign: These copies were missing so didn't appear correctly in the output.

jsign:
Rebased.

PR #70, opened 2020-01-28
grafana: enable anonymous readonly auth & cleaning & impr

jsign:
Will wait for #68  first, since some other stuff happened there in these files.


jsign:
Thanks!

PR #79, opened 2020-01-30
using native lotus types & ldevnet for tests

jsign:
Leverages a mixture of _TestDealFlow_ and _TestRPC_ plus some hacks to be able to run tests targeting a Lotus client API.


deals/deals_test.go
```diff
@@ -0,0 +1,73 @@
...
+	if len(cids) != len(dcfgs) {
+		t.Fatalf("some deal cids are missing, got %d, expected %d", len(cids), len(dcfgs))
+	}
+	time.Sleep(time.Second)
+	for {
+		di, err := dnet.Client.ClientGetDealInfo(ctx, cids[0])
+		checkErr(t, err)
+		if di.State == api.DealComplete {
+			break
+		}
```
jsign: :tada:

ldevnet/ldevnet.go
```diff
@@ -0,0 +1,326 @@
...
+	ld.cancel()
+
+	for i := 0; i < ld.numMiners; i++ {
+		<-ld.done
+	}
+	close(ld.done)
+	ld.closer()
+}
+
+func New(t *testing.T, numMiners int) (*LocalDevnet, error) {
```
jsign: This package `ldevnet` is for building the temp FIL devnet. 
Mostly based on _TestDealFlow_ test from Lotus, mixed with other test to enable API... so to make a FIL network with a client Lotus node, and let our code communicate via API (not internal things). So, to simulate what we're really going to do.

I stretched a bit for supporting more than 1 miner compared to original impl... seems to work but not 100% sure, have to play more with this to be sure.

.github/workflows/test.yml
```diff
@@ -15,6 +15,13 @@ jobs:
         uses: actions/setup-go@v1
         with:
           go-version: 1.13
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - name: OpenCL dependencies
+        run: sudo apt-get install mesa-opencl-icd ocl-icd-opencl-dev
```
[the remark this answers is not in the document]
sanderpick: 👍

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "extern/filecoin-ffi"]
```
sanderpick: Good call on the sub modules

jsign:
Thanks!

PR #97, opened 2020-02-03
Run gRPC web proxy as part of server process

asutula:
This will allow connections from javascript gRPC clients

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -1,13 +1,16 @@
 package server
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net"
 	"net/http"
 	"os"
 	"path/filepath"
+	"time"
 
+	"github.com/improbable-eng/grpc-web/go/grpcweb"
 	"github.com/ipfs/go-datastore"
 	badger "github.com/ipfs/go-ds-badger2"
 	logging "github.com/ipfs/go-log"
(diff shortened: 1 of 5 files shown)


api/server/server.go
```diff
@@ -59,18 +62,21 @@ type Server struct {
...
-	LotusAddress    ma.Multiaddr
-	LotusAuthToken  string
-	GrpcHostNetwork string
-	GrpcHostAddress string
-	RepoPath        string
+	LotusAddress        ma.Multiaddr
+	LotusAuthToken      string
+	GrpcHostNetwork     string
+	GrpcHostAddress     string
+	GrpcServerOpts      []grpc.ServerOption
```
asutula: Bubbling these config options all the way up so was can easily add tls configuration at some point

api/server/server.go
```diff
@@ -59,18 +62,21 @@ type Server struct {
...
+	GrpcWebProxyAddress string
```
asutula: Address to run the grpc web proxy server on

api/server/server.go
```diff
@@ -125,6 +131,30 @@ func NewServer(conf Config) (*Server, error) {
...
+		if wrappedServer.IsGrpcWebRequest(r) ||
+			wrappedServer.IsAcceptableGrpcCorsRequest(r) ||
+			wrappedServer.IsGrpcWebSocketRequest(r) {
+			wrappedServer.ServeHTTP(w, r)
+		}
+	})
+	grpcWebProxy := &http.Server{
+		Addr:    conf.GrpcWebProxyAddress,
+		Handler: handler,
+	}
```
asutula: Wrap the grpc service in the web proxy per https://github.com/improbable-eng/grpc-web/tree/master/go/grpcweb and consistent with how we do it in `go-threads`

exe/server/main.go
```diff
@@ -7,15 +7,17 @@ import (
...
 	logging "github.com/ipfs/go-log"
 	"github.com/textileio/fil-tools/api/server"
 	"github.com/textileio/fil-tools/tests"
-	_ "net/http/pprof"
 )
 
 var (
-	grpcHostAddr = "127.0.0.1:5002"
+	grpcHostAddr     = "0.0.0.0:5002"
+	grpcWebProxyAddr = "0.0.0.0:6002"
```
asutula: Updated to `0.0.0.0` so we can accept connection over any network interface.

asutula: Also @jsign when this gets merged and deployed to azure, can you open up port 6002?

jsign: Sure, will do.

PR #101, opened 2020-02-04
retrieval api, ldevnet multiple miners, renaming

### LINKED ISSUE #94: ldevnet: see if can add more miners
Would be nice to see if more miners could be included in ldevnets.

### LINKED ISSUE #100: retrieve api
Make a minimal API to retrieve stored data.
Minimal since current Lotus API might change soon, so don't hold tight on the current one.

jsign:
Closes #94 
Closes #100

In this PR the `ldevnet` package now _supports_ multiple miners; generally less than 6 without starting to have some problems. Most of the work needed for supporting this was done in the previous PR but it needed a special tweak (after lots of testing), and was only allowing one miner to mine. This is like simulating that there's a very powerful miner in the network. This isn't a problem for our needs since we only need that all miners accept deals and do their stuff. We aren't interested in them earning block rewards or similar. 

Regarding tests:
- Extended the `TestStore` for deals to test for 1, 3 and 5 miners; works.
- Created a new `Retrieve` API in the _Deals Module_ to have a test to check the full cycle of storing something, and checking that the retrieval matches what was stored. 

Unfortunately, the current code in Lotus doesn't accept fetching the data from multiple miners where we store; it only _remembers_ the last miner where the data was stored. So, even we can store things in multiple miners, the retrieval tests can't test each of them, only the last one. I left the code generic since this will be eventually implemented, but the test case is configured to store and retrieve only from one miner (since storing in extra ones cant be retrieved so it's meaningless).

Other changes:
- Renamed the `DealConfig` to a better name since there will be also `RetreievalDeals`.
- The retrieval of data is with the `dataCID` which was stored, not with the `ProposalCids`. Since this is crucial information to retrieve what was stored, I added it to the return values of `Store` and to all the path backward of gRPC types and on.


api/client/deals.go
```diff
@@ -27,7 +27,7 @@ type WatchEvent struct {
 
 // Store creates a proposal deal for data using wallet addr to all miners indicated
 // by dealConfigs for duration epochs
-func (d *Deals) Store(ctx context.Context, addr string, data io.Reader, dealConfigs []deals.DealConfig, duration uint64) ([]cid.Cid, []deals.DealConfig, error) {
+func (d *Deals) Store(ctx context.Context, addr string, data io.Reader, dealConfigs []deals.StorageDealConfig, duration uint64) ([]cid.Cid, []deals.StorageDealConfig, error) {
```
jsign: All changes regarding `StorageDealConfig` are just that, a rename.

api/client/deals.go
```diff
@@ -70,22 +70,22 @@ func (d *Deals) Store(ctx context.Context, addr string, data io.Reader, dealConf
 		return nil, nil, err
 	}
 
-	cids := make([]cid.Cid, len(reply.GetCids()))
-	for i, replyCid := range reply.GetCids() {
+	cids := make([]cid.Cid, len(reply.GetProposalCids()))
```
jsign: Renamed to `GetProposalCids()` since now there's a new attribute for the CID of the data stored, so to avoid confusions.

deals/deals.go
```diff
@@ -57,26 +62,26 @@ func New(api API, opts ...Option) (*Module, error) {
 
 // Store creates a proposal deal for data using wallet addr to all miners indicated
 // by dealConfigs for duration epochs
-func (m *Module) Store(ctx context.Context, waddr string, data io.Reader, dcfgs []DealConfig, dur uint64) ([]cid.Cid, []DealConfig, error) {
+func (m *Module) Store(ctx context.Context, waddr string, data io.Reader, dcfgs []StorageDealConfig, dur uint64) (cid.Cid, []cid.Cid, []StorageDealConfig, error) {
```
jsign: Now the `Store` API also returns the CID of the data, which is needed in the new API for retrieving.

deals/deals.go
```diff
@@ -92,7 +97,34 @@ func (m *Module) Store(ctx context.Context, waddr string, data io.Reader, dcfgs
 		}
 		proposals = append(proposals, *proposal)
 	}
-	return proposals, failed, nil
+	return dataCid, proposals, failed, nil
+}
+
+func (m *Module) Retrieve(ctx context.Context, waddr string, cid cid.Cid) (io.ReadCloser, error) {
```
jsign: Quite simple `Retrieve` API: using a specified wallet, it tries to retrieve the data with `cid`.
The API fetches the data from any miner that has it. Unfortunately, not much complex things can be done *now* since Lotus only holds the last miner for which *we* stored the data. 

The bold part is important. No feature yet to discover if other Lotus node stored that cid... most probably will be implemented by Lotus in the near future. When that happens, I think this method would be refactored since we would like to have more fine-grained control in which _Retrieval Ask_ we want to use to fetch the data.

deals/deals_test.go
```diff
@@ -24,9 +25,43 @@ func TestMain(m *testing.M) {
...
+			defer close()
+			m, err := New(dnet.Client, WithImportPath(filepath.Join(os.TempDir(), "imports")))
+			checkErr(t, err)
+			_, err = storeMultiMiner(m, dnet, nm, randomBytes(1000))
+			checkErr(t, err)
+		})
+	}
+}
+
+func TestRetrieve(t *testing.T) {
```
jsign: A new test that tests whole cycle of storing something, retriving it, and checking all things went as expected.

deals/deals_test.go
```diff
@@ -71,56 +106,73 @@ func TestWatchStore(t *testing.T) {
...
+	if err := waitForDealComplete(dnet, pcids); err != nil {
+		return cid.Undef, fmt.Errorf("error waiting for deal to complete: %s", err)
+	}
+	return dcid, nil
 }
 
 func waitForDealComplete(dnet *ldevnet.LocalDevnet, deals []cid.Cid) error {
 	ctx := context.Background()
-	for {
+	finished := make(map[cid.Cid]struct{})
```
jsign: Bugfix. Should check that *all* proposals are complete before returning.

deals/pb/deals.proto
```diff
@@ -38,8 +38,9 @@ message StoreRequest {
 }
 
 message StoreReply {
-    repeated string cids = 1;
-    repeated DealConfig failedDeals = 2;
+    string dataCid = 1;
```
jsign: Now there's an extra return attr with the `dataCid`, will be used for retrieval.

ldevnet/ldevnet.go
```diff
@@ -78,11 +78,11 @@ func New(t *testing.T, numMiners int) (*LocalDevnet, error) {
 		}
 		time.Sleep(time.Second)
 
-		mine := true
+		mine := i == 0
```
jsign: A magic-trick to make a single miner miner blocks in a multi-miner setup.
This avoids problems in chain syncing between them since blocks are mined fast and I suspect many forks happen. Since mining from different miners isn't important for us, we let the chain be mined by only one of them. What we're interested is that miners store information, post it on chain (which doesn't require mining), and offer data retrieval.

.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   test:
     name: Test
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
```
jsign: Runs in a Github Actions Runner in Azure.

.github/workflows/test.yml
```diff
@@ -24,13 +24,5 @@ jobs:
         run: sudo apt-get update && sudo apt-get install mesa-opencl-icd ocl-icd-opencl-dev
       - name: Check out code
         uses: actions/checkout@v1
-      - name: Cache dependencies
```
jsign: No need for this, since the self-hosted runner already has all dependencies cached.

Makefile
```diff
@@ -26,5 +26,5 @@ test: build
 	cat build/proof-params/parameters.json | jq 'keys[]' | xargs touch
 	mv -n v20* /var/tmp/filecoin-proof-parameters
 	rm v20* || true
-	go test ./... -short
+	go test -short -p 1 -v -count 1 ./...
```
jsign: Be easy on the VM to minimize flakyness. No parallel, and force always run all tests (since successfull ones may be cached).

api/client/deals.go
asutula: Been wondering if we should just stop using the word "deal". It could just be "storage" and "retrieval". I have a feeling we started with "deal" just because of words in the grant proposal.

asutula: But this is fine for now.

api/client/deals.go
asutula: Makes sense.

deals/deals.go
asutula: Cool!

deals/deals.go
```diff
@@ -92,7 +98,34 @@ func (m *Module) Store(ctx context.Context, waddr string, data io.Reader, dcfgs
...
+	offers, err := m.api.ClientFindData(ctx, cid)
+	if err != nil {
+		return nil, err
+	}
+	if len(offers) == 0 {
+		return nil, ErrNoOffersAvailable
+	}
+	for _, o := range offers {
+		log.Debugf("trying to retrieve data from %s", o.Miner)
+		if err := m.api.ClientRetrieve(ctx, o.Order(addr), f.Name()); err != nil {
```
asutula: Will be interesting to see how this should be integrated into a client code. Looks like this is simply writing to a file and blocking until that work is done. If that is the case, maybe we should write the client code in a way that parallelizes calls like this. Just a thought.

jsign: Actually I think that most of the current impl of `Store` and `Retrieve` will change quite a bit.

I think this code is quite minimal since we hadn't the chance to really test it. Now that we can run more e2e tests, I think we can start making all this code more meaningful. I'm mostly refering to more interesting configurations such as replication.

I still feel there are some gaps in which responsibility this module should have. I hope that clears when we build something that needs this module for its use-case.

api/client/deals.go
jsign: SGTM

jsign:
Thanks!

deals/deals.go
asutula: Sounds good

sanderpick:
I had a look too... Nice work! Tricky stuff.

PR #103, opened 2020-02-05
Implement retrieve service and client

asutula:


api/client/deals.go
```diff
@@ -136,3 +136,44 @@ func (d *Deals) Watch(ctx context.Context, proposals []cid.Cid) (<-chan WatchEve
...
+	reader, writer := io.Pipe()
+
+	go func() {
+		for {
+			// Check if the context has been canceled, and if so, close the writer an finish
+			_, ok := <-ctx.Done()
+			if !ok {
+				_ = writer.Close()
+				break
+			}
```
asutula: Trying to think of the best way to respond to the context being canceled. This just checks each time around the loop. Is there a better way?

deals/service.go
```diff
@@ -137,3 +137,34 @@ func (s *Service) Watch(req *pb.WatchRequest, srv pb.API_WatchServer) error {
...
+			return nil
+		}
+		sendErr := srv.Send(&pb.RetrieveReply{Chunk: buffer[:bytesRead]})
+		if sendErr != nil {
+			return sendErr
+		}
+		if err == io.EOF {
+			break
+		}
+	}
```
asutula: Hmmm I suppose we need to react to the `srv.Context` being canceled, and in that case stop reading the `reader` and return?

asutula: Can apply any feedback about the best way to do that in the client here as well

asutula:
Could use some help figuring out what's up with the failed tests @jsign

jsign:
> Could use some help figuring out what's up with the failed tests @jsign

Huh... for some reason running the test for some heavy setup of a network with 5 miners seems too much to ask. Even when using the custom self-hosted runner...

Can you verify that `make test` runs OK on your laptop? I don't have any high-end one and this wasn't never a problem.

jsign:
Including something regarding that in  #104 with some comments.

asutula:
`make test` runs fine on my laptop

deals/service.go
asutula: Have an idea here, about to push and will get feedback

api/client/deals.go
jsign: You can do `if ctx.Err() == nil`, if that's the case the context isn't cancelled.

asutula: Yea? Err gets set if you cancel?

asutula: Ah I see that in the docs.

jsign: Yep.
Here's some detailed info:
```
    // If Done is not yet closed, Err returns nil.
    // If Done is closed, Err returns a non-nil error explaining why:
    // Canceled if the context was canceled
    // or DeadlineExceeded if the context's deadline passed.
    // After Err returns a non-nil error, successive calls to Err return the same error.
```

deals/service.go
```diff
@@ -137,3 +137,34 @@ func (s *Service) Watch(req *pb.WatchRequest, srv pb.API_WatchServer) error {
...
+	defer reader.Close()
+	if err != nil {
+		return err
+	}
+
+	buffer := make([]byte, 1024*32) // 32KB
+	for {
+		bytesRead, err := reader.Read(buffer)
+		if err != nil && err != io.EOF {
+			return nil
```
jsign: Returning nil was intentional?

asutula: No that's a bug, fixed locally already.

deals/service.go
```diff
@@ -137,3 +137,34 @@ func (s *Service) Watch(req *pb.WatchRequest, srv pb.API_WatchServer) error {
...
+		}
+		sendErr := srv.Send(&pb.RetrieveReply{Chunk: buffer[:bytesRead]})
+		if sendErr != nil {
+			return sendErr
+		}
```
jsign: Nit: Could be inlined if you like: `if err := ... ; err != nil {`.

api/client/deals.go
asutula: Maybe I'm being too picky here, but my thinking about why it sucks to just check for canceled context on each time around the loop is that... if the context is canceled while you're blocked on receiving the stream, you have to wait until the stream is done being received and processed before you check the context state again. I'm working on a way where you'd get notified of the cancelation faster. Not sure how important this distinction is.

jsign: `stream.Rcv()` shouldn't fail when `ctx` is canceled?
Are you trying to handle that particular case for a reason?

jsign: I mean, it seems that if it was canceled closing with error would be OK.

asutula: The use case is the user of the api cancels the context that they passed in to `Retrieve()`. In that case, we should cancel all the work we're doing and return from the function.

jsign: I see, but I mean in that use-case, wouldn't `reply, err := stream.Recv()` fail with an error and quite with already existing code? Maybe I'm missing something about how `stream` is connected with the API call that returned it.

asutula: Ah good call. Yes, I think you're right. Because we're passing that user context into `stream, err := d.client.Retrieve(ctx, req)`.

asutula: Ok so in that case, I don't even need to check if the context is canceled because `stream.Recv()` should just return an error if it is canceled.

jsign: Yep!

asutula:
Ok all cleaned up @jsign

PR #104, opened 2020-02-05
use viper for configuration

### LINKED ISSUE #58: cmd/server: flags/envs
Currently, configs like: lotus addr, lotus token, grpc host addr, repo path, are somewhat fixed.
This eventually should be configurable by flags/envs.

jsign:
Closes #58 
Use `viper` for flags/envs.

### DIFF
--- a/deals/deals_test.go
+++ b/deals/deals_test.go
@@ -25,7 +25,7 @@ func TestMain(m *testing.M) {
 }
 
 func TestStore(t *testing.T) {
-	numMiners := []int{1, 3, 5}
+	numMiners := []int{1, 3}
 	for _, nm := range numMiners {
 		t.Run(fmt.Sprintf("CantMiners%d", nm), func(t *testing.T) {
 			dnet, _, _, close := tests.CreateLocalDevnet(t, nm)
--- a/exe/server/Dockerfile
+++ b/exe/server/Dockerfile
@@ -6,12 +6,12 @@ COPY extern extern
 COPY go.mod go.sum ./
 RUN go mod download
 COPY . . 
-RUN GOOS=linux go build -o main exe/server/main.go && \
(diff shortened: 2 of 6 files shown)


deals/deals_test.go
```diff
@@ -25,7 +25,7 @@ func TestMain(m *testing.M) {
 }
 
 func TestStore(t *testing.T) {
-	numMiners := []int{1, 3, 5}
+	numMiners := []int{1, 3}
```
jsign: Reducing heaviness despite in my laptop this never seemed to be a problem.
I'll take a look if there's something extra to do CI-wise... I don't want to set-up a self-hosted costly thing for running this thing, seems like too much. On the other hand, don't like the idea of relying on us to make heavier tests...

exe/server/Dockerfile
```diff
@@ -6,12 +6,12 @@ COPY extern extern
 COPY go.mod go.sum ./
 RUN go mod download
 COPY . . 
-RUN GOOS=linux go build -o main exe/server/main.go && \
-go run github.com/GeertJohan/go.rice/rice append --exec main -i ./build
+RUN GOOS=linux go build -o fil-tools exe/server/main.go && \
```
asutula: how about `fil-toolsd` since in some cases you'll be running the server and cli (presumably called `fil-tools`) on the same computer?

asutula: If this only applies to docker, maybe it doesn't matter, don't know.

jsign: Yeah, actually mostly applies to docker since the binary isn't use elsewhere. 
Despite that, I like the convention. Will change.

jsign:
Changed the binary suffix aligned with usual conventions.

exe/server/Dockerfile
asutula: cool

PR #105, opened 2020-02-05
Cli command for retrieve

asutula:
Simple, writes a file


exe/fil-tools/cmd/deals_retrieve.go
```diff
@@ -0,0 +1,84 @@
...
+	"path"
+
+	"github.com/caarlos0/spin"
+	"github.com/ipfs/go-cid"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func init() {
+	retrieveCmd.Flags().StringP("address", "a", "", "wallet address of the file to fetch")
```
jsign: Nit: maybe confusing descripton? Alternative: `wallet address to use funds for retrieval`.

exe/fil-tools/cmd/deals_retrieve.go
```diff
@@ -0,0 +1,84 @@
...
+	retrieveCmd.Flags().StringP("cid", "c", "", "cid of the file to fetch")
```
jsign: Nit: maybe te concept of _file_ can be avoided, in general. Not sure though.

exe/fil-tools/cmd/deals_retrieve.go
asutula: cool, updated.

exe/fil-tools/cmd/deals_retrieve.go
asutula: Sure. Changed to "data".

PR #123, opened 2020-02-11
Isolate types

asutula:
This may fix the transitive dependency @jsign mentioned.

### DIFF
--- a/api/client/asks.go
+++ b/api/client/asks.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"time"
 
-	"github.com/textileio/fil-tools/index/ask"
 	pb "github.com/textileio/fil-tools/index/ask/pb"
+	"github.com/textileio/fil-tools/index/ask/types"
 )
 
 // Asks provides an API for viewing asks data
@@ -14,17 +14,17 @@ type Asks struct {
 }
 
 // Get returns the current index of available asks
-func (a *Asks) Get(ctx context.Context) (*ask.Index, error) {
+func (a *Asks) Get(ctx context.Context) (*types.Index, error) {
 	reply, err := a.client.Get(ctx, &pb.GetRequest{})
 	if err != nil {
 		return nil, err
 	}
 	lastUpdated := time.Unix(reply.GetIndex().GetLastUpdated(), 0)
-	storage := make(map[string]ask.StorageAsk, len(reply.GetIndex().GetStorage()))
+	storage := make(map[string]types.StorageAsk, len(reply.GetIndex().GetStorage()))
 	for key, val := range reply.GetIndex().GetStorage() {
 		storage[key] = askFromPbAsk(val)
 	}
-	return &ask.Index{
+	return &types.Index{
 		LastUpdated:        lastUpdated,
 		StorageMedianPrice: reply.GetIndex().StorageMedianPrice,
 		Storage:            storage,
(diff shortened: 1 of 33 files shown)


asutula:
There is one file where I'm not sure about how it will effect the transitive dependency situation for a program that imports the `client` package. It is `test_utils.go` in the `client` package. It contains this code:
```
package client

import (
	"context"
	"testing"

	"github.com/textileio/fil-tools/api/server"
	"github.com/textileio/fil-tools/tests"
	"google.golang.org/grpc"
)
```
This is mysterious to me because I don't know if a program that imports the `client` package actually imports the test files as well. If so, the imports of
```
"github.com/textileio/fil-tools/api/server"
"github.com/textileio/fil-tools/tests"
```
may branch out into all the extra dependencies we're trying to avoid.

Other than that, the `client` package now only imports the various simple `types` package and `pb` package included in each index/module package.

jsign:
Did you try to make a simple project and import it now and see if it resulted?

jsign:
> There is one file where I'm not sure about how it will effect the transitive dependency situation for a program that imports the `client` package. It is `test_utils.go` in the `client` package. It contains this code:
> 
> ```
> package client
> 
> import (
> 	"context"
> 	"testing"
> 
> 	"github.com/textileio/fil-tools/api/server"
> 	"github.com/textileio/fil-tools/tests"
> 	"google.golang.org/grpc"
> )
> ```
> 
> This is mysterious to me because I don't know if a program that imports the `client` package actually imports the test files as well. If so, the imports of
> 
> ```
> "github.com/textileio/fil-tools/api/server"
> "github.com/textileio/fil-tools/tests"
> ```
> 
> may branch out into all the extra dependencies we're trying to avoid.
> 
> Other than that, the `client` package now only imports the various simple `types` package and `pb` package included in each index/module package.

Test files in general should be on the form `*_test.go` so the compiler knows that shouldn't be included on main compilation. So, that should help.

jsign:
I'm doing heavy work on the next PR and have some realization about something related to this.

Currently, the tests of the CLI assume that some things are already running in the host. (that's the reason these tests are the only ones flagged with the `-short` condition in `go test`).

Now, there are two options to fix that:
1- Spin them up in the test itself.
2- Spin them up in the test but without needing to run the code to do it.

Option 1 should be kind of discarded. If the _server_ is instantiated in the test code itself, it will inevitably have the dependencies we don't want to have. If the tests have those dependencies, so will do the code of the client.

Option 2 sounds better and it's the way I'm testing things for the Fast API in my next PR. I'm spinning up dependencies by running a docker container. That would mean that we could run a _server_ with docker, run the test, and then kill it. This way the code of the test (and thus CLI) never would have an import that we don't want (all is abstracted in the docker container). 

---

On the other side, if deleting/commenting that test would avoid moving the client to a separate repo (considering the changes you've already made in this PR), I think it is worth it. So you may want to try that.

jsign:
I did some extra research about the current behavior of Go Modules, and it doesn't matter if the import graph doesn't include other things. All dependencies of the Go Module that contains that subpackage will be considered in the project _go.mod_. 

Someone is working in a proposal for Go Module dependency pruning to solve exactly this problem. If the import dependency graph doesn't include dependencies of the module, not force them in the module dependency graph. So, best-case scenario on 1.15 (so, to minor version ahead) will this work.


To be 100% sure that now won't work, I did an even simpler _client_, simple _main.go_:
```
package main

import "github.com/textileio/fil-tools/util"

func main() {
	_, err := util.TCPAddrFromMultiAddr(nil)
	if err != nil {
		panic(err)
	}
}
```

This simple client uses the `util` subpackage (`textileio/fil-tools/util`) which only imports:
```
import (
	"fmt"
	"time"

	ma "github.com/multiformats/go-multiaddr"
)
```
End result: annoying things in `textileio/fil-tools` _go.mod_ are considered in this program. So, even if we make the work of _cleaning_ the import graph of `fil-tools/api/client`, that still isn't enough.

Conclusion: for how things works now in Go, we should extract the client to another repo. 
In fact, the nasty dependencies that we want to avoid for the client, are in _fil-lotus_ for the same reason. We're interested in some subpackages of Lotus, but get all the deps bag _for free_.

asutula:
Ok thanks for the research. Closing.

PR #131, opened 2020-02-14
fastapi: first milestone

jsign:
Closes #121 
Closes #110 
Closes #89 
Closes #112 
Closes #126 
Closes #127 
Closes #129 
Closes #128 
Closes #119
Closes #120 
Closes #130

Some code comments to help... this PR got a bit massive for my original intentions.

PR #142, opened 2020-02-21
index/miners: keep old non-empty location data

### LINKED ISSUE #139: index/meta: don't override location if miner is offline

jsign:
Closes #139 

Depends on merging #131 

cc @andrewxhill might be interested in this fix when merged and deployed.


andrewxhill:
nice!

PR #153, opened 2020-02-28
proxy: handles ListenServe error in api proxy

andrewxhill:
@sanderpick caught this one in my recent textile pr so thought I'd nip it here too

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -264,7 +264,7 @@ func NewServer(conf Config) (*Server, error) {
 	}()
 
 	go func() {
-		if err := grpcWebProxy.ListenAndServe(); err != nil {
+		if err := grpcWebProxy.ListenAndServe(); err != nil && err != http.ErrServerClosed {
 			log.Errorf("error starting proxy: %v", err)
 		}
 	}()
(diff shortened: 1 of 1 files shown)


api/server/server.go
```diff
@@ -236,8 +236,10 @@ func NewServer(conf Config) (*Server, error) {
 		grpcServer.Serve(listener)
 	}()
 
+	errc := make(chan error)
 	go func() {
-		grpcWebProxy.ListenAndServe()
+		errc <- grpcWebProxy.ListenAndServe()
+		close(errc)
```
jsign: I wonder if a channel is used for a particular reason?
Compared to: `if err := grpcWebProxy.ListenAndServe(); err != nil ` and log error.

cc @sanderpick

sanderpick: Hmm, good question. I don't remember. Yeah, don't need the channel.

andrewxhill:
fixed that

api/server/server.go
```diff
@@ -237,7 +237,9 @@ func NewServer(conf Config) (*Server, error) {
 	}()
 
 	go func() {
-		grpcWebProxy.ListenAndServe()
+		if err := grpcWebProxy.ListenAndServe(); err != nil {
+			log.Errorf("error starting proxy: %v", err)
```
jsign: Nit: there's an edge case which isn't an error.
```
if err == http.ErrServerClosed {
					// Not really an error, but Shutdown() or Close() was called.
} 
```

jsign:
Merge away!

PR #155, opened 2020-02-28
FPA add file

### LINKED ISSUE #154: Add file to FPA
You should be able to add data to FPA from a local file. This entails sending the data to the server, adding it to IPFS and then pinning it and storing it in filecoin.

This flow should be represented in the gRPC service and cli.

asutula:
Adds ability to add data to FPA from a local file.
Covers everything from the server side to client, cli and tests

closes #154


asutula:
Some extra noise in here from renaming things from "store" and "put" to "add"

exe/fil-tools/cmd/fpa_add.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+func init() {
+	fpaAddCmd.Flags().StringP("token", "t", "", "FPA access token")
+
+	fpaCmd.AddCommand(fpaAddCmd)
+}
+
+var fpaAddCmd = &cobra.Command{
+	Use:   "add [cid|path]",
```
asutula: Decided to combine adding by cid and path into this single command. Also using args now for the cid/path instead of flags.

exe/fil-tools/cmd/helpers.go
```diff
@@ -42,7 +42,7 @@ func RenderTable(writer io.Writer, header []string, data [][]string) {
 	table.SetBorder(false)
 	table.SetTablePadding("\t")
 	table.SetNoWhiteSpace(true)
-	headersColors := make([]tablewriter.Colors, len(data[0]))
+	headersColors := make([]tablewriter.Colors, len(header))
```
asutula: General bug

fpa/fastapi/fastapi_add.go
```diff
@@ -23,19 +24,39 @@ var (
 	ErrNotStored     = errors.New("cid not stored")
 )
 
-func (i *Instance) Put(ctx context.Context, c cid.Cid) error {
+func (i *Instance) AddFile(ctx context.Context, reader io.Reader) (*cid.Cid, error) {
 	ar := i.auditor.Start(ctx, i.info.ID.String())
-	ar.Close()
-	if err := i.put(ctx, ar, c); err != nil {
+	defer ar.Close()
```
asutula: Bug fix, pretty sure this should be deferred.

fpa/fastapi/fastapi_add.go
```diff
@@ -23,19 +24,39 @@ var (
...
+		ar.Errored(err)
+		return nil, err
+	}
+	ar.Success()
+	return cid, nil
+}
+
+func (i *Instance) AddCid(ctx context.Context, c cid.Cid) error {
+	ar := i.auditor.Start(ctx, i.info.ID.String())
+	defer ar.Close()
```
asutula: Bug fix, pretty sure this should be deferred.

fpa/service.go
```diff
@@ -104,10 +104,46 @@ func (s *Service) Store(ctx context.Context, req *pb.StoreRequest) (*pb.StoreRep
...
+	if err != nil {
+		return err
+	}
+
+	reader, writer := io.Pipe()
+	defer reader.Close()
+
+	go receiveFile(srv, writer)
+
+	cid, err := i.AddFile(srv.Context(), reader)
```
asutula: @sanderpick I ended up letting the call to `AddFile` block and spun the data processing out in a goroutine... much cleaner, thx for the suggestion.

api/client/fpa.go
```diff
@@ -32,6 +32,42 @@ func (f *fpa) Store(ctx context.Context, c cid.Cid) error {
...
+	for {
+		bytesRead, err := data.Read(buffer)
+		if err != nil && err != io.EOF {
+			return nil, err
+		}
+		sendErr := stream.Send(&pb.AddFileRequest{Chunk: buffer[:bytesRead]})
+		if sendErr != nil {
+			if sendErr == io.EOF {
+				var noOp interface{}
+				return nil, stream.RecvMsg(noOp)
```
asutula: Oh, this is important. I was thrown off by this for 24 hrs. Turns out that if the server returns an error in the middle of client streaming requests, you get back error `io.EOF`. If you want to see the actual error from the server, you have to get it with `stream.RecvMsg()`. Will get this updated elsewhere in the code base.

exe/fil-tools/cmd/fpa_add.go
```diff
@@ -0,0 +1,76 @@
...
+	PreRun: func(cmd *cobra.Command, args []string) {
+		err := viper.BindPFlags(cmd.Flags())
+		checkErr(err)
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
+		defer cancel()
+
+		if len(args) != 1 {
+			Fatal(errors.New("you must provide a valid cid or file path"))
```
jsign: Nit: text error in this msg is the same as in line 73. Maybe should be diferent? (say, delete the `valid` word in this msg).

fpa/fastapi/fastapi_add.go
jsign: Yeah, actually the `auditor` concept is still quite undefined... but since it has a `Close()` better to do this.

api/client/fpa.go
sanderpick: Good catch!

exe/fil-tools/cmd/fpa_add.go
sanderpick: Would be nice to handle stdin at some point. I was thinking the same for textile.

fpa/fastapi/fastapi_add.go
```diff
@@ -110,6 +131,15 @@ func (i *Instance) storeInFIL(ctx context.Context, c cid.Cid) (ColdInfo, error)
 	return ci, nil
 }
 
+func (i *Instance) addToHotLayer(ctx context.Context, reader io.Reader) (*cid.Cid, error) {
+	path, err := i.ipfs.Unixfs().Add(ctx, files.NewReaderFile(reader), options.Unixfs.Pin(false))
```
sanderpick: You guys might consider trying to combine the pinning and adding. I'm guessing the same pin will be added in a different step.

fpa/service.go
sanderpick: 👍

PR #159, opened 2020-03-04
Simple gateway web app

### LINKED ISSUE #158: Simple gateway web ui
Idea here is to make a very simple web app that will show some data from fil-tools. After getting fil-tools up and running, a user could use this to verify things are working.

asutula:
This simply shows the data in the various indices and reputation module. We'll figure out how this data could be represented in the Textile gateway separately.

Closes #158


asutula:
Screen Shot 2020-03-04 at 11 30 03 AM
Looks like that

reputation/reputation.go
```diff
@@ -85,8 +85,11 @@ func (rm *Module) GetTopMiners(n int) ([]MinerScore, error) {
 	if n < 1 {
 		return nil, fmt.Errorf("the number of miners should be greater than zero")
 	}
-	mr := make([]MinerScore, 0, n)
 	rm.lockScores.Lock()
+	if n > len(rm.scores) {
+		n = len(rm.scores)
+	}
+	mr := make([]MinerScore, 0, n)
```
asutula: Ran across this bug.

gateway/gateway.go
```diff
@@ -0,0 +1,366 @@
...
+		"Rows":      rows,
+	})
+}
+
+func (g *Gateway) reputationHandler(c *gin.Context) {
+	menuItems := makeMenuItems(3)
+
+	topMiners, err := g.reputationModule.GetTopMiners(30)
+	if err != nil {
+		g.renderError(c, http.StatusInternalServerError, err)
```
jsign: Nit: maybe make `30` a const, and use it on line 282 to make the text future-proof.

gateway/gateway.go
```diff
@@ -0,0 +1,366 @@
...
+	}()
+	go func() {
+		for {
+			select {
+			case err, ok := <-errc:
+				if err != nil {
+					if err == http.ErrServerClosed {
+						return
+					}
+					log.Errorf("gateway error: %s", err)
```
jsign: Should a `return` be added after this line? I think that even if that isn't the case, the next for loop will enter in the `if !ok` and exit indirectly, but that sounds like a lucky consequence than on purpose?

gateway/gateway.go
asutula: cool, will do

gateway/gateway.go
asutula: Yea I think you're right, added the `return`.

PR #188, opened 2020-03-13
powergate: enable rich configurations for Hot and Cold storages

### LINKED ISSUE #176: Add default Cid config per API
Considering that now storing _Cids_ have rich configurations such as duration, country filtering, renewal flags, hot/cold storage enabling, etc; it becomes a burden to indicate all that for each stored Cid in the instance.

Implement a new feature _Default API config_ which defines a default configuration for _Cids_ which will be used unless an explicit configuration is also indicated when storing a particular _Cid_.

### LINKED ISSUE #177: Add RepFactor to Cid config
Create a new feature that allows storing a Cid in a number N of miners, which also comply with other filters from the configuration (e.g: not in blacklist, countries, etc).

### LINKED ISSUE #178: Add Timeout to HotStorage Cid config
Retrieving data from _HotStorage_ can be cancelled/timedout by using idiomatic _context.Context_ since it is a synchronous operation.

Storing a new Cid is an async operation, and _Hot Storage_ involves fetching some data from the IPFS network. Allow a new configuration attribute _Hot.Ipfs.AddTimeout_ which caps the number of seconds allowed to search for that data in IPFS before failing.

### LINKED ISSUE #179: Add Duration in ColdLayer config
Create a new configuration item that indicates the duration of _new_ Deals in the Filecoin network. This means, every time a deal is made for this Cid it will use the configured duration.

### LINKED ISSUE #180: Add Country filtering ColStorage configuration
Create a new configuration item that allows specifying countries codes, which miners selected for new deals much comply. Countries should be specified by long-ISO names, e.g: China, Uruguay.

We could support in the future also other ISO names such as short ones: _USA_, _URY_, etc.

### LINKED ISSUE #181: Add miner addr backlisting in ColdStorage config
Enable the client decide on a black-list of storage providers (miners) which won't be used not matter what other rules are configured.

e.g: if the blacklist is `["t042"]`, then miner _t042_ will never be selected for making deals even if complies to all other specified filters.

### LINKED ISSUE #182: Add Enabled config for Hot and Cold storage
The _Enabled_ flag will simply indicate if the type of storage is disabled for storing the data.

For example, if _Enabled_ is false for _Hot Storage_ and _true_ for _Cold Storage_, the Cid will only be stored in the _Cold Storage_. If the client wants to retrieve that Cid, it should _unfreeze_ the data, which means, set to _true_ the _Enabled_ flag in the _Hot Storage_.

Similarly, the client can opt to only store data in the _Hot Storage_

### LINKED ISSUE #184: Add Renewal to ColdStorage config
Add _Renewal_ feature that makes new deals on a certain threshold when deals are about to expire.

There's an interesting case if the renewal implies fetching the data from cold layer, which may incur a cost. Support to that is separated to #198

### LINKED ISSUE #189: Unfreeze: from Cold to Hot
If a Cid was stored with the _Hot Layer_ _Enabled_ flag set to _false_, then we call the data to be _frozen_ (aka: only in the _Cold Storage).

Since data retrieval can only happen from the _Hot Storage_, there should be a mechanism to fetch this Cid from the _Cold Storage_ to the _Hot Storage_ to enable retrieval. This can be indicated simply by setting _Enabled_ from _Hot Storage_ to _true_. The _Scheduler_ should simply comply to that change, fetching the data from the only available layer (_Cold_) and saving it in the _Hot Storage_.

Saying it differently, _unfreezing data_ despite being a complex process for the _Scheduler_, from the client POV is just setting the _Hot Storage_ _Enable_ flag to _true_. 

If the client calls _GetFromHot()_ and the data is _frozen_, it will get an _ErrShouldUnfreeze_.
Since _unfreezing_ the data is just pushing a new changed Cid configuration, the client watch that _Job_ state to know when data is available for retrieval.

### LINKED ISSUE #199: Use self-hosted GH runner
The _integration tests_ run for FFS which cover all features seems to be too much work for the free-tier of GH Actions. Set-up a cloud VM to host a GH runner with more power.

### LINKED ISSUE #200: deals: include DealID and ActivationEpoch
Include _DealID_ and _ActivationEpoch_ information to successful deal results.

The _DealID_ is an identifier of the Deal to be used as a reference in other APIs.

The _ActivationEpoch_ is the epoch in which the Deal became fully active on-chain, which only the chain knows when the miner provides the first proof that it's storing the data properly. This precise information is needed to correctly run the deal-renewal algorithm since we need the _ActivationEpoch_ plus the _Duration_ to take appropriate actions.

jsign:
This PR targets a bunch of new features for FFS and other lateral refactors and improvements.

---

**Features**:
Regarding _new features_, they're mapped to the following issues which have further details about them:
- Closes #176
- Closes #178
- Closes #177
- Closes #179
- Closes #180
- Closes #181
- Closes #182
- Closes #189
- Closes #184

---
**Other changes**:
Here're other non-feature changes:
- Closes #199
- Closes #200
- Renamed _FastAPI_ or _Instance_ concepts to simply _API_.
- _Api_ persistence was abstracted to a separate component interface _InstanceStore_.
- _Scheduler_ persistence was abstracted to the following component interfaces:
   - _CidInfoStore_
   - _JobStore_
   - _PushConfigStore_ 
- Documentation was updated.
- Renamed _PayloadCid_ to _DataCid_, which refers to the Cid of the data stored on Filecoin. Remember that the original data is transformed in a CAR file, and then Lotus makes another tree for storage-proofs, so the final stored data has other Cid compared to the original data Cid. 

---

To get a nice taste of how all these new features work, it may be a good idea to take a look at `ffs/integrationtest/integration_test.go`, which has test for each individual feature running with a real IPFS Node and a Filecoin Devnet. 

Also, all `ffs` packages have complete exported comments to help documentation.


api/server/server.go
```diff
@@ -22,14 +22,18 @@ import (
 	"github.com/textileio/powergate/deals"
```
jsign: This file is mostly a bootstrapping point for creating the server. Apart from creating some extra things due to new features, I started extracting a bit of the logic of creating things to improve readability.

I wouldn't say it's in a perfect state, but better than before. I think what would enable a cleaner next step is grouping dependencies in the `Sever` struct instead of being flat, that would allow continuing creating new methods that would create that sub-parts of the Sever... which can have a `Close()` method to a more reasonable close-flow.

ffs/scheduler/scheduler.go
```diff
@@ -120,53 +142,188 @@ func (s *Scheduler) run() {
...
+		if err := s.evaluateRenewal(ctx, a); err != nil {
+			log.Errorf("renweal of %s: %s", a.Config.Cid, err)
+		}
+	}
+}
+
+func (s *Scheduler) evaluateRenewal(ctx context.Context, a ffs.PushConfigAction) error {
+	inf, err := s.cis.Get(a.Config.Cid)
+	if err == ErrNotFound {
+		log.Infof("skip renewal evaluation for %s since hasn't cid info", a.Config.Cid)
```
sanderpick: Nit: "since don't have cid info"?

ffs/integrationtest/integration_test.go
```diff
@@ -3,6 +3,7 @@ package integrationtest
 import (
```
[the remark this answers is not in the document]
sanderpick: Very nice.

jsign:
I landed some extra commits with things I found running the server in embedded mode... which uses the reptop instead of fixed _MinerSelector_... since now the _Reputation_ module was asking for more data in miners, that was too much to ask for the miner-host in the embed network.
Anyway, all good now.

Thanks for the reviews!

Issue #204, opened 2020-03-24
Cannot build pow


sebastiendan:
The README lists the following command to build `pow` but such path doesn't exist:

```
go build -i -o pow exe/cli/main.go 
```

Tried with both paths under /exe with no success.

jsign:
You're right! It seems that on some renamings that folder got incorrectly renamed.
I created & merged #209, which fixes that and also refreshed a bit the _Readme_ which was stale. It's still pretty dry, but much better than before.

I'm currently working on #186  which will make building and running easier since I will take out all CGO related stuff.

jsign:
@sebastiendan, a lot has happened since your report. The Readme has some new information about building PG. The recommended way is running `make build-cli` for building the CLI. If you want to run the server (and lotus, and ipfs) to use the cli, I'd recommend running `make fresh` standing in the `/docker` folder. Recall that the Lotus node should be synced to be able to use FFS.

jsign:
I'll be closing this issue. Please re-open one if you find some problems using the new workflow.
Thanks for the report!

Issue #186, opened 2020-03-12
Extract devnet to other repo & avoid CGO


jsign:
Importing `Lotus` as a dependency comes with a heavy bag of things that aren't nice:
- git-submodule of a Rust library
- Need of CGO
- Include some assets in the binary
- A ton of dependencies in the module
- Since the gRPC client is in this repo, anyone that tries to use the client will carry this heavy bag. This is because the way Go calculates dependency graphs.

Most of the above pain points are already solved in automated make-rules, Dockerfiles, etc. So, they aren't painful in the day-to-day DX. I'd say the worst thing is having to extract the client gRPC to another repo (a thing that might be good by other reasons). Also, having weird things in the Makefile and in Dockerfiles, despite shouldn't be changed, are simply weird and not related to Powergate.

Lotus it's imported because we leverage:
- Method signatures and types of the API.
- Needed things to run the devnets for tests and embedded mode.

The only way to avoid items of the first list without sacrificing the second list is:
- Moving the devnet creation to another repo. Make docker-image releases, which can be used in tests in this repo. This way, no imports are needed to have a devnet running.
- Using that same docker-image to run the _embedded_ mode, but at the docker-compose level.
- All types and signatures of the Lotus API, should be _copy,pasted_. Importing them means importing the Lotus Go Module and thus get again to the original problem.

So... would be good to dedicate some time to give a try to do all this.

PR #209, opened 2020-03-24
rename cli folder & update readme

jsign:
It seems with some renaming the folder containing the CLI got an unwanted rename, so getting back to its original name.

Motivated by #204, I quickly updated the readme with updated explanations about building stuff and using some powerful features (such as _embedded_ mode). Most prob, this readme will change again when #186 lands, which will make some steps simpler.


jsign:
I'll merge this since it has no code changes, and allow refreshed Readme right away!

PR #210, opened 2020-03-24
Make Powergate CGO free & use testnet/3

### LINKED ISSUE #186: Extract devnet to other repo & avoid CGO

### LINKED ISSUE #211: use testnet/3
Considering `lotus@testnet/3` will be`master` soon, take a step ahead and move there.
It may provide APIs to unblock other work too.

### LINKED ISSUE #212: add building CLI and server binaries to Makefile
Considering #210  landed, make `make build` just build the cli and server binaries for convenience.

### LINKED ISSUE #213: ffs/manager: Make master address configurable
Currently, the _master address_ from which new funds are sent when a new _API_ instance is created is inferred by _the default wallet addr in the Lotus node_. 

Allow configuring the address by a flag/env, that's a much safer way.

jsign:
To have some context and motivation see #186.

Most of the deleted code wasn't really unnecessary but moved to:
https://github.com/textileio/lotus-client
https://github.com/textileio/lotus-devnet

Both repos have some work on that I pushed indiscriminately. From now on I'll stick to PRs to have a clearer history. Updates to those repos will be triggered by updating to a newer version of _Lotus_. 

Also, while doing `lotus-client` I realized that with some PR and some OK from Lotus-team, might be reasonable to make direct-importing of the API part from Lotus. Now isn't possible since some dependency graph ends in the CGO world. (Which is what I fixed by building that almost-similar repo). Eventually, it would be nice to do that PR to have one less repo to maintain.

The `lotus-devnet`seems more useful in the long-run, so I dropped some words in the readme on how to use it.

The idea now is that _Powergate_ uses `lotus-client`, which is a CGO free Lotus client. And use the docker-containers auto-generated by `lotus-devnet` to:
- Use in the docker-file of the embedded mode. The network isn't run by `powd` anymore, but just a new container in the docker-compose.
- In integration tests.

The above means _Powergate_ isn't tied to any CGO thing anymore, so the `Makefile`, docker image building, and cross-platform compiling are much simplified.

---

While doing the CGO stuff, we decided to switch to `testnet/3` to be a step ahead of what will happen. That took some extra work to get things working since the API was quite different compared to `master`.

The only thing that isn't good is that it seems that the devnet doesn't allow making more than one deal per run. I had to set an _skip_ to the _Renewal_ test that was working on `master` since, of course, that test implies making a second deal to renew the first one. 
I'll be tracking an issue I created [here](https://github.com/filecoin-project/lotus/issues/1455) to see if that can be fixed.

Apart from that, the migration went fine with enough work.

---

Closes #186 
Closes #211 
Closes #212 
Closes #213


api/server/server.go
```diff
@@ -93,6 +92,7 @@ type Config struct {
 	IpfsApiAddr         ma.Multiaddr
 	LotusAddress        ma.Multiaddr
 	LotusAuthToken      string
+	LotusMasterAddr     string
```
jsign: #213

api/server/server.go
```diff
@@ -104,25 +104,21 @@ type Config struct {
...
+
 	if conf.Embedded {
-		c, cls, err = lotus.NewEmbedded()
-		if err != nil {
-			return nil, fmt.Errorf("creating the embedded network: %s", err)
-		}
-		masterAddr, err = c.WalletDefaultAddress(context.Background())
-		if err != nil {
-			return nil, fmt.Errorf("getting default address: %s", err)
+		if masterAddr, err = c.WalletDefaultAddress(context.Background()); err != nil {
```
jsign: In _embedded_ mode we get the _master address_ as the default, like in the _old_ days.
This because that whole devnet is ephemeral.

api/server/server.go
```diff
@@ -174,7 +170,7 @@ func NewServer(conf Config) (*Server, error) {
 	lchain := lotuschain.New(c)
 	var ms ffs.MinerSelector
 	if conf.Embedded {
-		ms = fixed.New([]fixed.Miner{{Addr: "t0300", EpochPrice: 4000000}})
+		ms = fixed.New([]fixed.Miner{{Addr: "t01000", EpochPrice: 1000000}})
```
jsign: Necessary consequences of switching to `testnet/3`. Similar changes are in the wild.

api/stub/asks.go
```diff
@@ -19,29 +19,29 @@ func (a *Asks) Get(ctx context.Context) (*ask.Index, error) {
 			Miner:        "miner1",
 			Price:        5001,
 			MinPieceSize: 1004,
-			Timestamp:    uint64(time.Now().Unix()),
-			Expiry:       uint64(time.Now().Unix()),
+			Timestamp:    1,
+			Expiry:       100,
```
jsign: `Timestamp` and `Expiry` are now of type `ChainEpoch`, which ultimately is an integer.
That integer refers to the height at which things happen (or will).

chainsync/chainsync.go
```diff
@@ -4,23 +4,18 @@ import (
...
 	"github.com/filecoin-project/lotus/chain/types"
+	"github.com/textileio/lotus-client/api/apistruct"
+	"github.com/textileio/lotus-client/chain/store"
 )
 
-// API provides an abstraction to a Filecoin full-node
-type API interface {
-	ChainGetPath(context.Context, types.TipSetKey, types.TipSetKey) ([]*store.HeadChange, error)
-	ChainGetGenesis(context.Context) (*types.TipSet, error)
-}
```
jsign: Decided to drop these "abstractions" from modules. In the end they are always referring to concrete Lotus APIs.
Originally, I thought it would be useful to just narrow down _which_ APIs where used here.

Feels defining this API might be very much related in plans of supporting multiple clients or similar. It will take some works to build some reasonable-coherent generic API. But this is another story.

deals/deals.go
```diff
@@ -183,15 +190,15 @@ func pushNewChanges(ctx context.Context, api API, currState map[cid.Cid]*api.Dea
 				Size:          dinfo.Size,
 				PricePerEpoch: dinfo.PricePerEpoch.Uint64(),
 				Duration:      dinfo.Duration,
-				DealID:        dinfo.DealID,
+				DealID:        uint64(dinfo.DealID),
```
jsign: `DealID` didn't exist in `testnet/3`, actually I was the one that included it in `master` to use here.
Did the same PR yesterday but in this branch: https://github.com/filecoin-project/lotus/pull/1452

deals/deals.go
```diff
@@ -183,15 +190,15 @@ func pushNewChanges(ctx context.Context, api API, currState map[cid.Cid]*api.Dea
...
-			if dinfo.State == storagemarket.DealComplete {
-				ocd, err := api.StateMarketStorageDeal(ctx, dinfo.DealID, types.EmptyTSK)
+			if dinfo.State == storagemarket.StorageDealActive {
+				ocd, err := client.StateMarketStorageDeal(ctx, dinfo.DealID, types.EmptyTSK)
 				if err != nil {
 					log.Errorf("getting on-chain deal info: %s", err)
 					continue
 				}
-				newState.ActivationEpoch = ocd.ActivationEpoch
+				newState.ActivationEpoch = int64(ocd.State.SectorStartEpoch)
```
jsign: `tesntet/3` changes

deals/deals_test.go
```diff
@@ -28,27 +35,24 @@ func TestStore(t *testing.T) {
...
+			_, err = storeMultiMiner(m, client, nm, randomBytes(1600))
 			checkErr(t, err)
 		})
 	}
 }
-
 func TestRetrieve(t *testing.T) {
 	numMiners := []int{1} // go-fil-markets: doesn't support remembering more than 1 miner
-	data := randomBytes(1000)
+	data := randomBytes(1600)
```
jsign: Magic number changes for `testnet/3` sectorbuilder-mock. 
Nothing too much important since this data is random... the size shouldn't be meaningful.

docker/docker-compose-embedded.yaml
```diff
@@ -1,5 +1,8 @@
 version: '3.7'
 
+volumes:
```
jsign: Tricky part with elegant solution. 
In any real setup, the Lotus node and Powergate are in the same host.
Powergate saves the file locally and pass a file-path to Lotus to store data. So they have to share a docker-volume to make them appear running in the same host.

docker/docker-compose-embedded.yaml
```diff
@@ -14,14 +17,25 @@ services:
       - 6002:6002
     depends_on:
       - ipfs
+      - lotus
     environment:
       - TEXPOWERGATE_EMBEDDED=true
+      - TEXPOWERGATE_LOTUSHOST=/dns4/lotus/tcp/7777
```
jsign: Now, a run of Powergate with _embedded_ flag is _almost_ like a real-case... the Lotus node it's _outside_ and calls the APIs _as usual_. It still keeps making ephemeral Powergate repos for convenience.

docker/docker-compose-embedded.yaml
```diff
@@ -14,14 +17,25 @@ services:
...
-    restart: unless-stopped    
+
+  lotus:
+    image: textile/lotus-devnet:lotus-testnet3-d41aeb
+    ports:
+      - 7777:7777
+    environment:
+      - TEXLOTUSDEVNET_SPEED=1500
+    volumes:
+     - pow-embed-shared:/tmp/powergate
```
jsign: So, some thousands code lines less are resumed here.
The `textile/lotus-devnet` is running the devnet, exposing in 7777 the Lotus API.
Refer to that repo readme for the flags/envs possible.

exe/server/Dockerfile
```diff
@@ -1,33 +1,14 @@
 FROM golang:1.14 as builder
```
jsign: So, greatly simplified since we don't need Rust for building, or OpenCL and other things.
Now, `CGO_ENABLED=0` can be used, and also `alpine` as the base final image for running.

ffs/integrationtest/integration_test.go
```diff
@@ -40,9 +39,16 @@ import (
 	"github.com/textileio/powergate/wallet"
 )
 
+const (
+	tmpDir = "/tmp/powergate"
```
jsign: This is fixed since it's important for the mapping of the docker-volume with the _"should be local"_ lotus node.

ffs/integrationtest/integration_test.go
```diff
@@ -139,15 +147,15 @@ func TestInfo(t *testing.T) {
...
+	n := 1
 	for i := 0; i < n; i++ {
 		cid, _ := addRandomFile(t, r, ipfs)
 		jid, err := fapi.PushConfig(cid)
 		require.Nil(t, err)
 		requireJobState(t, fapi, jid, ffs.Success)
 	}
 
-	t.Run("WithThreeAdds", func(t *testing.T) {
+	t.Run("WithOneAdd", func(t *testing.T) {
```
jsign: Related with not being able to run more than one deal mentioned before and in PR description.

ffs/integrationtest/integration_test.go
```diff
@@ -454,12 +461,14 @@ func TestUnfreeze(t *testing.T) {
 }
 
 func TestRenew(t *testing.T) {
+	// ToDo: unskip when testnet/3 sectormock allows more than one deal
+	// See https://github.com/filecoin-project/lotus/issues/1455
+	t.SkipNow()
```
jsign: The only test that was _skipped_ in the testnet/3 migration. A bit annoying, hope fix soon.

ldevnet/ldevnet.go
```diff
@@ -1,374 +0,0 @@
-package ldevnet
```
jsign: Most of here is now in https://github.com/textileio/lotus-devnet
This changed since this worked on `master`, so I had to migrate to `testnet/3`.

lotus/client.go
```diff
@@ -5,12 +5,10 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/filecoin-project/lotus/api/apistruct"
-	"github.com/filecoin-project/lotus/build"
+	"github.com/filecoin-project/lotus/lib/jsonrpc"
```
jsign: Notice we use directly the `jsonrpc` impl of Lotus and not having our own (based on them). So one less thing to mantain.

lotus/client.go
```diff
@@ -35,7 +33,8 @@ func New(maddr ma.Multiaddr, authToken string) (*apistruct.FullNodeStruct, func(
 	closer, err := jsonrpc.NewMergeClient("ws://"+addr+"/rpc/v0", "Filecoin",
 		[]interface{}{
 			&api.Internal,
-		}, headers, jsonrpc.WithReconnect(true, time.Second*3, 0))
```
jsign: The reconnect thing is still enabled. I did a PR some months ago in Lotus with this feature.
https://github.com/filecoin-project/lotus/pull/1155

So, it's baked in now.

lotus/client.go
```diff
@@ -54,16 +53,6 @@ func New(maddr ma.Multiaddr, authToken string) (*apistruct.FullNodeStruct, func(
 
 }
 
-func NewEmbedded() (*apistruct.FullNodeStruct, func(), error) {
```
jsign: 👋

tests/ldevnet.go
```diff
@@ -2,34 +2,51 @@ package tests
...
+	"github.com/textileio/powergate/util"
 )
 
-func init() {
-	build.InsecurePoStValidation = true
-}
-
-func CreateLocalDevnet(t *testing.T, numMiners int) (*ldevnet.LocalDevnet, address.Address, []address.Address, func()) {
-	dnet, err := ldevnet.New(numMiners, ldevnet.DefaultDuration)
+func CreateLocalDevnet(t *testing.T, numMiners int) (*apistruct.FullNodeStruct, address.Address, []address.Address) {
```
jsign: So now `CreateLocalDevnet` spinup a docker container instead of running the devnet per-se.
The good part is that all code called this method, so changing this was indistinguishable for _the clients_.

wallet/wallet.go
```diff
@@ -35,7 +30,16 @@ func New(api API, maddr *address.Address, iam big.Int) (*Module, error) {
...
 func (m *Module) NewAddress(ctx context.Context, typ string) (string, error) {
-	addr, err := m.api.WalletNew(ctx, typ)
+	ty := crypto.SigTypeUnknown
+	if "bls" == typ {
+		ty = crypto.SigTypeBLS
+	} else if "secp256k1" == typ {
+		ty = crypto.SigTypeSecp256k1
+	} else {
+		return "", fmt.Errorf("unkown wallet type %s", typ)
+	}
```
jsign: A quick fix for the `testnet/3` change since now it's typed, not a string.
I won't imagine the types of wallet to change much, so isn't that bad... but don't fully like this kind of ifs.

lotus/client.go
asutula: That, plus not having to maintain all those API structs is much cleaner!

lotus/client.go
asutula: 🥂

deals/deals.go
jsign: Forgot to mention, `DealID` was necessary for a reason... it is needed in other API for getting the Epoch at which the deal became active.

PR #220, opened 2020-03-31
Push new Cid Configs & CAR flag deal and retrieval

### LINKED ISSUE #116: fps: change pinned configuration
Hot <--> Cold change on existing pin.

### LINKED ISSUE #187: Direct CAR file import in Lotus
The current API in Lotus has assumptions about re-encoding the Cid data into UnixFS.

In our case, we already make a CAR transformation before storing the data, so our goal is to store the data _exactly as sent_ to Lotus. Saying it differently, we don't want Lotus to do any other data transformation for the Deal.

Currently, seems that this isn't possible in Lotus _master_, but in a feature branch of interop there're new APIs. We should have a chat and see if this could be done at _master_ (or see if gets eventually merged).

jsign:
This PR addresses two things, pushing newer `CidConfigs`, and new CAR support from Lotus API. I'll drop some extra words about the first thing, which is important.

---

Recall that `CidConfig` is the desired storing state of a Cid reg. Hot and Cold storage. So, pushing new desired states can be separated in two cases:

*Changes in Hot Storage config*
This case's basically enabling or disabling storage. This isn't a problem because it basically maps to pinning or unpinning the data in the IPFS node. The IPFS node storing state is easily changeable.

*Changes in Cold Storage*
Ok, here there are more things that could be changed: country filters, excluded miners, deals duration, rep factor, etc.
Now, when you come and think about it the only thing that should make some _possible immediate action_ when a new _CidConfig_ is pushed is a change in the _RepFactor_. 

If _RepFactor_ increased, *some* new deals should be made. The new deals should select miners considering the other configurations reg. country filters, excluded miners, duration selected, etc. e.g: if rep factor is 1 and is increased to 3, then 2 new deals satisfying all other conditions should be created. (not 3, since we already have 1).

If _RepFactor_ decreased, pushing this config doesn't execute any real action despite saving the new desired state. This because existing deals can't be undone, so if the _RepFactor_ decreased from 2 to 1, then there will be 2 deals active until some of them expire. The important part here is that, if the _CidConfig_ has _Renweal_ activated, the _Scheduler_ will renew only the last deal before becoming expired. Saying it differently, the _renew_ logic isn't blindly renewing all active deals that are about to expire, but it also considers _repfactor_ now.

I hope that makes sense... actually, it's basically still guiding all thoughts about enforcing what the client tells us to do. Doing the reasonable stuff to make the _desired state_ real.

In other changes like changing _country filtering_, _excluded Miners_, _deal durations_, etc; they never trigger an immediate action. That information is saved to be used when the next deal, for whatever reason (renewal, or increased repfactor). This sounds like the most reasonable way to think about it: the current desired parameters to consider when making *new* deals.

---

One last thing to mention. Since the logic of pushing a new `CidConfig` basically mean _do whatever possible to ensure this state_, that has similar semantics of the _repair_ feature that is pending. That's no coincidence since this design was considering that _repair_ feature.

So the _repair logic_ could be also understood as an indirect execution of the current `CidConfig`. For the _Scheduler_, both things have the same intention: ensure that state is true.

Of course, auto-pushing the current _CidConfig_ to simulate a _Repair_ is a bit weird since the client is the one that pushes configs. But my point is that most of the work necessary is already done under the covers. Most of the repair part will be introducing a cron-style job (similar to the renew feature) that will kind-of do that: mostly execute again the last known _CidConfig_, that will achieve the goal.

---

Closes #116 
Closes #187 
Required a [PR](https://github.com/filecoin-project/lotus/pull/1472#pullrequestreview-385078408) in Lotus which got merged
Implied doing and merging https://github.com/textileio/lotus-client/pull/4, and https://github.com/textileio/lotus-devnet/pull/8


api/server/server.go
```diff
@@ -134,7 +134,9 @@ func NewServer(conf Config) (*Server, error) {
 		return nil, fmt.Errorf("creating repo folder: %s", err)
 	}
 
-	ds, err := badger.NewDatastore(path, &badger.DefaultOptions)
+	opts := &badger.DefaultOptions
+	opts.NumVersionsToKeep = 0
+	ds, err := badger.NewDatastore(path, opts)
```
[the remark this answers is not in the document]
sanderpick: Oh, did they make a new release?

ffs/coreipfs/coreipfs.go
```diff
@@ -43,6 +43,29 @@ func (ci *CoreIpfs) Put(ctx context.Context, b blocks.Block) error {
...
+}
+
+// IsStored return if a particular Cid is stored.
+func (ci *CoreIpfs) IsStored(ctx context.Context, c cid.Cid) (bool, error) {
+	ci.ipfs.Block()
+	pins, err := ci.ipfs.Pin().Ls(ctx)
+	if err != nil {
+		return false, fmt.Errorf("getting pins from IPFS: %s", err)
+	}
+	for _, p := range pins {
```
sanderpick, line 61: It could be worth keeping a cache of pins as a map/index? Pretty wild you have loop through the pin set.

ffs/interfaces.go
```diff
@@ -54,9 +54,11 @@ type Scheduler interface {
 // data or Cids.
 type HotStorage interface {
 	Add(context.Context, io.Reader) (cid.Cid, error)
+	Remove(context.Context, cid.Cid) error
 	Get(context.Context, cid.Cid) (io.Reader, error)
 	Pin(context.Context, cid.Cid) (int, error)
 	Put(context.Context, blocks.Block) error
+	IsStored(context.Context, cid.Cid) (bool, error)
```
sanderpick, line 61: Nit: `Pin` related to `IsStored`, so maybe `IsStored` -> `Pinned` or `IsPinned`, or since the API is called `HotStorage`, `Pin` -> `Store`?

ffs/integrationtest/integration_test.go
```diff
@@ -256,36 +266,70 @@ func TestRepFactor(t *testing.T) {
...
 
-	t.Run("IncreaseBy1", func(t *testing.T) {
-		t.SkipNow()
-		ipfsAPI, fapi, cls := newAPI(t, 2)
-		defer cls()
-		cid, _ := addRandomFile(t, r, ipfsAPI)
-		jid, err := fapi.PushConfig(cid)
-		require.Nil(t, err)
-		requireJobState(t, fapi, jid, ffs.Success)
+func TestRepFactorIncrease(t *testing.T) {
```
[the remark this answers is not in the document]
sanderpick: Awesome to see the scheduler doing its thing!

ffs/integrationtest/integration_test.go
```diff
@@ -256,36 +266,70 @@ func TestRepFactor(t *testing.T) {
...
+	require.Nil(t, err)
+	requireJobState(t, fapi, jid, ffs.Success)
+	requireCidConfig(t, fapi, cid, &config)
+	cinfo, err = fapi.Show(cid)
+	require.Nil(t, err)
+	require.Equal(t, 2, len(cinfo.Cold.Filecoin.Proposals))
+	require.Contains(t, cinfo.Cold.Filecoin.Proposals, firstProposal)
+}
+
+func TestRepFactorDecrease(t *testing.T) {
```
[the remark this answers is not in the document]
sanderpick: Nice, makes sense 👍

api/server/server.go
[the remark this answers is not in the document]
jsign: Nah, I've been following discussions about the Badger size problem... they keep appearing and some tweaks in the configuration are suggested. Seeing if this one changes something! I don't have much hope, but we can try.

ffs/coreipfs/coreipfs.go
jsign: Sounds like a good idea since nobody else should be touching the IPFS node. 
Maybe only refreshing it after some time or similar. Will make an issue!

ffs/interfaces.go
jsign: Yeah, I didn't want to use a `Pin` concept because `HotStorage` should be a higher-level concept than IPFS.
Say, a `HotStorage` could be saving things in S3 (or in both IPFS and S3, who knows).

asutula: If I remember right, `Pin` is the method to store data that is already in IFPS, take the existing cid and _pin_ it here. But you can also `Add` (and maybe `Put`, I'm not sure), raw data. So there are two ways (at least) of storing data. `IsStored` is how you check if data stored in any of those ways is stored.

sanderpick: > Yeah, I didn't want to use a `Pin` concept because `HotStorage` should be a higher-level concept than IPFS.
> Say, a `HotStorage` could be saving things in S3 (or in both IPFS and S3, who knows).

Agreed. That's why I was suggesting the option of not using the word "pin" in the API.

sanderpick: Taking the buckets case, the only reason to not auto-pin when you add is because it's more efficient to _update_ the existing pin (if it exists). Ideally, that would be handled internally in add/put:
1. Search for a pin for the given cid/block, if exists, add w/o auto-pinning
2. Add/put
3. If there was an existing pin, update it

sanderpick: (Definitely not blocking this PR, just thinking out loud)

jsign: Sorry @sanderpick , I misread your original comment; now I see what you mean.
First, I agree that `Pin()` should be changed to `Store` to be aligned with `IsStored`. I'll add this change to the next PR.

Regarding `Put()`, the only use of that API is for CAR importing. The CAR library needs a `Store` to store blocks when the CAR file is processed:
```go
type Store interface {
	Put(blocks.Block) error
}
```
After all that is loaded, then `Pin(..)` is called with the root of the DAG, which pins recursively.

I preferred not pinning in that implementation because if the CAR importing fails in the middle of the process, that would mean that half of the DAG was pinned without its root. And pinning blocks without the proper root would mean that will be pinned forever, since unpinning is with the root.

To be clear, that API is called on individual blocks of the DAG by the CAR lib, not only with root Blocks. That's why this is different than _buckets_. 
Also, at the FFS level the concept of _Updating_ a pin doesn't _make sense_ since the API pins Cids, not _Files_ or higher-level concepts that mutate. 

For some time I've been thinking about the need of a new API to `Swap` or `Replace` or a similar idea to simplify some client use-cases as _Buckets_ or similar. They can still add and remove Cids, but native support for `Swap`ping Cids makes the Pin update at the IPFS layer more efficient.

Makes sense?

sanderpick: Okay, cool! Makes sense.

PR #234, opened 2020-04-08
Update gPRC and Go Client

### LINKED ISSUE #201: ffs/grpc: support all new rich configurations and async style
Make appropriate changes in the gRPC service to support rich configurations and the async style of the API.

### LINKED ISSUE #233: Update gRPC and Go Client to latest API
Related to #201

asutula:
This get the gRPC layer and Go client completely up to date with the latest FFS API.

closes #233
closes #201


api/client/ffs.go
```diff
@@ -0,0 +1,244 @@
+package client
```
asutula, line 1: Renamed this file and unfortunately GitHub is showing it as new. Will try to comment on new/important things below.

api/client/ffs.go
```diff
@@ -0,0 +1,244 @@
...
+
+type ffs struct {
+	client pb.APIClient
+}
+
+// JobEvent represents an event for Watching a job
+type JobEvent struct {
+	Job ff.Job
+	Err error
+}
```
asutula, line 22: Type to pass updates, including any error, to job listener

api/client/ffs.go
```diff
@@ -0,0 +1,244 @@
...
+	return ff.ApiID(resp.ID), nil
+}
+
+func (f *ffs) WalletAddr(ctx context.Context) (string, error) {
+	resp, err := f.client.WalletAddr(ctx, &pb.WalletAddrRequest{})
+	if err != nil {
+		return "", err
+	}
+	return resp.Addr, nil
+}
```
asutula, line 46: `ID` and `WalletAddr` added for completeness, some overlap with output from `Info`

api/client/ffs.go
asutula: Actually, the API changes will be visible in the `.proto` diff. Just know they're all implemented in here for the client.

api/client/ffs_test.go
```diff
@@ -0,0 +1,27 @@
+package client
+
+import (
+	"testing"
+
+	pb "github.com/textileio/powergate/ffs/pb"
+)
```
asutula, line 7: Wanted to start creating tests for the client, but realized all the client tests need to be updated to use whatever docker image method we use in other tests to get the node up and running. Will make and issue and work on that later.

ffs/pb/fps.proto
```diff
@@ -6,92 +6,227 @@ option java_package = "io.textile.filecoin.ffs.pb";
...
+}
+
+message AddToHotReply {
+  string cid = 1;
+}
+
+
+
 service API {
-    rpc AddCid(AddCidRequest) returns (AddCidReply) {}
```
asutula, line 91: `AddCid` is no longer needed now that `PushConfig` exists.

ffs/pb/fps.proto
```diff
@@ -6,92 +6,227 @@ option java_package = "io.textile.filecoin.ffs.pb";
...
+   rpc GetDefaultCidConfig(GetDefaultCidConfigRequest) returns (GetDefaultCidConfigReply) {}
+   rpc GetCidConfig(GetCidConfigRequest) returns (GetCidConfigReply) {}
+   rpc SetDefaultCidConfig(SetDefaultCidConfigRequest) returns (SetDefaultCidConfigReply) {}
+   rpc Show(ShowRequest) returns (ShowReply) {}
+   rpc Info(InfoRequest) returns (InfoReply) {}
+   rpc Watch(WatchRequest) returns (stream WatchReply) {}
+   rpc PushConfig(PushConfigRequest) returns (PushConfigReply) {}
+   rpc Get(GetRequest) returns (stream GetReply) {}
+   rpc Close(CloseRequest) returns (CloseReply) {}
+   rpc AddToHot(stream AddToHotRequest) returns (AddToHotReply) {}
```
asutula, line 231: `AddFile` is now `AddHot` and it only adds the file to IPFS and returns the CID that can be used in `PushConfig`. Figured its better to keep it simple. If any user of the client (eg the CLI) wants to created more automated UX, they can.

ffs/pb/fps.proto
```diff
@@ -6,92 +6,227 @@ option java_package = "io.textile.filecoin.ffs.pb";
...
-    rpc Show(ShowRequest) returns (ShowReply) {}
+   rpc Create(CreateRequest) returns (CreateReply) {}
+   rpc ID(IDRequest) returns (IDReply) {}
+   rpc WalletAddr(WalletAddrRequest) returns (WalletAddrReply) {}
+   rpc GetDefaultCidConfig(GetDefaultCidConfigRequest) returns (GetDefaultCidConfigReply) {}
+   rpc GetCidConfig(GetCidConfigRequest) returns (GetCidConfigReply) {}
+   rpc SetDefaultCidConfig(SetDefaultCidConfigRequest) returns (SetDefaultCidConfigReply) {}
+   rpc Show(ShowRequest) returns (ShowReply) {}
+   rpc Info(InfoRequest) returns (InfoReply) {}
+   rpc Watch(WatchRequest) returns (stream WatchReply) {}
```
asutula, line 227: The service implementation of `Watch` internally calls `Unwatch` if/when the client closes the connection, so there is no need to expose `Unwatch` over gRPC.

asutula:
I updated the CLI minimally just to get things compiling with the Client updates. A full CLI update will come in a next PR.

api/client/ffs_test.go
jsign: Sounds good to get the CLI covered!

ffs/grpc/service.go
```diff
@@ -82,115 +201,169 @@ func (s *Service) Show(ctx context.Context, req *pb.ShowRequest) (*pb.ShowReply,
...
-		}
-		_, writeErr := writer.Write(req.GetChunk())
-		if writeErr != nil {
-			writer.CloseWithError(writeErr)
+		select {
+		case <-srv.Context().Done():
+			return nil
+		case job, ok := <-updates:
+			if !ok {
+				return nil // do we want to return an error here? nil will result as EOF on the client i think
```
jsign, line 306: Under normal circumstances, this shouldn't happen since the client should control this with the `Unwatch` call.
I'd add maybe a warning/error here just to see if that is happening and maybe the consequence of a bug.
I'd keep the `return nil` and let the client decide what to do in that case.

ffs/pb/fps.proto
jsign: Sounds good to me... safer than letting the client forget to do it.

api/client/ffs.go
```diff
@@ -0,0 +1,244 @@
...
+	return f.client.Show(ctx, &pb.ShowRequest{
+		Cid: c.String(),
+	})
+}
+
+func (f *ffs) Info(ctx context.Context) (*pb.InfoReply, error) {
+	return f.client.Info(ctx, &pb.InfoRequest{})
+}
+
+func (f *ffs) Watch(ctx context.Context, jids ...ff.JobID) (<-chan JobEvent, error) {
```
jsign, line 95: I wonder if adding some comments to client APIs would be useful. In particular in this kind of APIs where proper canceling of the `ctx` is needed to free up resources.

APIs that aren't stream-based, canceling the `ctx` isn't necessary since the request will finish and things will be freed up on the server-side.

Not really a particular comment of this method, nor this repo... but in general on stream-based gRPC client methods.

asutula: Yea good call. Will make an issue for it.

PR #244, opened 2020-04-14
Watch -> WatchJobs & refactor

jsign:
Since the addition of `WatchLogs`, the method `Watch` sounds confusing. 
Renaming `Watch` to `WatchJobs` in multiple layers. 

Also, changing the style of the method to simplify background goroutines handling and allow to remove the `Unwatch` API. Also, delegates to the client decide the buffer size of the channel.


ffs/pb/ffs.proto
```diff
@@ -173,11 +173,11 @@ message InfoReply {
    InstanceInfo info = 1;
 }
 
-message WatchRequest {
+message WatchJobRequest {
```
asutula, line 176: Sorry, nit here, but maybe should be plural `WatchJobsRequest` and below, `WatchJobsReply`

jsign: Agree! will fix

PR #247, opened 2020-04-15
Net and Health modules

### LINKED ISSUE #191: Peers Module
There is no existing PG abstraction for data about node peers, so for now we'll create a new PG Peers module that exposes the data available from Lotus apis. This will drive some of the web UI in addition to being generally available for any client.

### LINKED ISSUE #242: RFC: Health, Peers, Net modules
There are potentially a few new modules we'll want to implement. They should be generally useful and will also help support the Lotus web UI project. I want to get my thoughts about the design and implementation written down to collect feedback and come up with a final plan.

To start, the three modules I imagine existing are Health, Peers and Net:

- **Health** - Provide information about the node health. This could be derived from many sources of information.
- **Peers** - Information about node peers. This will likely be augmented with derived data such as ip-based peer location. For this reason, it may be useful to think of/model this like an _index_ similar to the existing Miners index.
- **Net** - Provides general information about and methods for interacting with the network

The first pass at these modules can easily be supported by the "Net" methods of the Lotus API:

```
NetConnectedness func(context.Context, peer.ID) (network.Connectedness, error)
NetPeers         func(context.Context) ([]peer.AddrInfo, error)
NetConnect       func(context.Context, peer.AddrInfo) error
NetAddrsListen   func(context.Context) (peer.AddrInfo, error)
NetDisconnect    func(context.Context, peer.ID) error
NetFindPeer      func(context.Context, peer.ID) (peer.AddrInfo, error)
```

To begin, the single input to the Health module will be `NetConnectedness`. It's good to have Health as a module because we will add other sources of information to compute "health".

`NetPeers` can be queried on some interval (chain update maybe?) and augmented with derived data to build the peers index.

The remaining methods can be used to build the Net module with simple methods for taking the node on and offline and fining peers.

That is my first thought on how to model this. I think it is simple but gives us some flexibility going forward to incorporate other data sources and/or adapt to other Filecoin implementations.

asutula:
Ended up defining the Net module with a very similar API as the Lotus Net methods, but wraps up peer information with IP based location. This allowed me not to have to worry about a Peers module for now... We just get all the peer information we need from the Net module.

The Health module depends on the Net module and currently just uses some very very simple logic to determine local node health.

closes #242 
closes #191


health/module.go
```diff
@@ -0,0 +1,55 @@
...
+// New creates a new net module
+func New(net net.Module) *Module {
+	m := &Module{
+		net: net,
+	}
+	return m
+}
+
+// Check reutuns the current health status and any messages related to the status
+func (m *Module) Check(ctx context.Context) (status Status, messages []string, err error) {
```
asutula, line 36: Just makes sure we have an acceptable connection status to each of our peers. This can be improved in the future

health/rpc/Makefile
```diff
@@ -0,0 +1,16 @@
...
+	protoc -I=. -I=$(GOPATH)/src \
+	--go_out=\
+	plugins=grpc:\
+	. $<
+
+clean:
+	rm -f *.pb.go
+	rm -f *pb_test.go
+
+.PHONY: clean
```
asutula, line 16: Lots of these Makefiles copied around the repo now. The syntax of this thing is a mystery to me, need to learn how they work, but is there a way to parameterize it so we can just have one and pass in the path to the dir we want to run on?

health/rpc/api.go
```diff
@@ -0,0 +1,40 @@
+package rpc
```
asutula, line 1: Trying a slightly different directory structure in the new modules where all rpc things (.proto file, generated code, api impl) go in a `rpc` directory. I think I like it.

index/miner/miner_test.go
```diff
@@ -28,7 +24,7 @@ func TestFullRefresh(t *testing.T) {
 	client, _, miners := tests.CreateLocalDevnet(t, 1)
 	time.Sleep(time.Second * 15) // Allow the network to some tipsets
 
-	mi, err := New(tests.NewTxMapDatastore(), client, &p2pHostMock{}, &lrMock{})
+	mi, err := New(tests.NewTxMapDatastore(), client, &tests.P2pHostMock{}, &tests.LrMock{})
```
asutula, line 27: Moved these mocks to a shared location

net/interface.go
```diff
@@ -0,0 +1,45 @@
...
+	Unknown
+
+	// Error means there was an error determining connectedness
+	Error
+)
+
+type PeerInfo struct {
+	AddrInfo peer.AddrInfo
+	Location iplocation.Location
+}
```
asutula, line 36: `PeerInfo` augments address info with location information.

net/lotus/module.go
```diff
@@ -0,0 +1,94 @@
+package lotus
```
asutula, line 1: This file is our lotus implementation of the `net` interface

net/lotus/module_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/textileio/powergate/net"
+	"github.com/textileio/powergate/tests"
+)
+
+var (
+	ctx = context.Background()
+)
+
+func TestModule(t *testing.T) {
```
asutula, line 15: It's cool all these methods actually work in the dev net. It sees the miner as peer.

tests/utils.go
```diff
@@ -0,0 +1,10 @@
+package tests
+
+import "testing"
+
+func CheckErr(t *testing.T, err error) {
```
asutula, line 5: Going to start to put test utils and other things (like the mocks above) in this `tests` package for reuse.

tests/mocks/mocks.go
```diff
@@ -0,0 +1,36 @@
+package mocks
```
asutula, line 1: These mocks are used in both the health and net modules tests. They should be used in `miner_test.go`, but it creates a circular dependency that needs to be resolved. Can't wrap my head around that at the moment, so leaving these mocks duplicated there in `miner_test.go` for now. Will ticket it up.

health/module.go
```diff
@@ -0,0 +1,55 @@
+package health
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/textileio/powergate/net"
+)
+
+// Module exposes the filecoin wallet api.
```
jsign, line 10: Nit: wallet?

health/module.go
```diff
@@ -0,0 +1,55 @@
...
+func (m *Module) Check(ctx context.Context) (status Status, messages []string, err error) {
+	peers, err := m.net.Peers(ctx)
+	if err != nil {
+		return Error, nil, err
+	}
+	for _, peer := range peers {
+		con, err := m.net.Connectedness(ctx, peer.AddrInfo.ID)
+		if err != nil {
+			messages = append(messages, fmt.Sprintf("error checking connectedness for peer %v: %v", peer.AddrInfo.ID.String(), err))
+		}
```
jsign, line 45: I think we need a `continue` after that append.

health/module_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/textileio/powergate/tests"
+	"github.com/textileio/powergate/tests/mocks"
+)
+
+var (
+	ctx = context.Background()
+)
+
+func TestModule(t *testing.T) {
+	t.Parallel()
```
jsign, line 17: Nit: I'm usually avoiding the parallel flag on tests that require local devnets since puts a lot of stress to the CI runner.

health/rpc/Makefile
jsign: Sure, it sounds like a good idea.
I'd even prefer a more general one, like a `make protos` in the root Makefile that will generate all protos recursively in the repo. Even if you're interested in a single one that changed, it's usually quite fast to worry about targeting a single one.

health/rpc/api.go
jsign: Yeah, the `pb` folder not being inside sounds purist, but in reality is only used for grpc.
I like that too, less folders is good.

health/rpc/api.proto
```diff
@@ -0,0 +1,25 @@
...
+option java_multiple_files = true;
+option java_package = "io.textile.powergate.health.rpc";
+option java_outer_classname = "PowergateHealth";
+option objc_class_prefix = "TTE";
+
+enum Status {
+    Ok = 0;
+    Degraded = 1;
+	Error = 2;
+}
```
jsign, line 13: Nit: indentation.

net/lotus/module.go
```diff
@@ -0,0 +1,100 @@
...
+	case network.CanConnect:
+		return net.CanConnect, nil
+	case network.CannotConnect:
+		return net.CannotConnect, nil
+	case network.Connected:
+		return net.Connected, nil
+	case network.NotConnected:
+		return net.NotConnected, nil
+	default:
+		return net.Unknown, nil
```
jsign, line 98: Ok, I was wondering before the meaning of `Unkown`. SGTM.
Maybe may worth adding a `log.Warn` so we might spot this situation in the logs to seems it will likely require some action.

net/rpc/api.proto
```diff
@@ -0,0 +1,85 @@
...
+    Location location = 2;
+}
+
+enum Connectedness {
+    NotConnected = 0;
+    Connected = 1;
+	CanConnect = 2;
+	CannotConnect = 3;
+	Unknown = 4;
+	Error = 5;
```
jsign, line 31: Nit: indent.

tests/utils.go
jsign: You can consider the pkg used in `integration_test.go` for assertions.
It's very short, and when fails it automatically provides nice output messages.
For this case would be: `require.Nil(t, err)` (in the test itself, so no need for `CheckErr` func).
I use it for other things like: `require.Equal(t, 10, someNumber)` and have many helping things that shrkings `if` lines.

health/rpc/Makefile
asutula: I like that idea. I have a lot of learning to do about Makefiles, so for now I'll make an issue for it.

tests/utils.go
asutula: That's great, will switch to that.

net/lotus/module.go
asutula: Yea, it's just because you can't do an exhaustive switch on "enums" in go, there is always some chance that there is a value you're not aware of. Cool, I'll add the log statement.

PR #249, opened 2020-04-16
Configure golangci-lint via yml file

asutula:
This is a cleaner way of integrating `golangci-lint` because the yml file will be used by the GitHub action, your IDE, and `golangci-lint` run locally. I think I have the yml file configured to effectively the same as was being run in the CI action (running `golint` with `exclude-use-default: false`). I also tried enabling the `mispell` linter and it is catching things that `reviewdog` is not, so I removed the `reviewdog` step in favor of this linter.


asutula:
If you want to see the `misspell` output, here's a failure from before I fixed the problems

https://github.com/textileio/powergate/pull/249/checks?check_run_id=590725004#step:4:5

asutula:
Actually looks like the `misspell` linter is just for code comments and "commonly misspelled" words, so I added `reviewdog` back in so now we can use both. I'm not sure why `reviewdog` didn't find the typos that `misspell` did.

jsign:
> I'm not sure why `reviewdog` didn't find the typos that `misspell` did.
I wonder the same. I agree better with both!

PR #253, opened 2020-04-17
Client & CLI updates for Health & Net

asutula:
Pretty straight forward. Also added tests for Health & Net client code. I want to make sure the pattern I'm using for testing is good, then we can expand the client tests next.


api/client/client.go
```diff
@@ -37,6 +41,8 @@ func NewClient(target string, opts ...grpc.DialOption) (*Client, error) {
 		Wallet:     &Wallet{client: walletPb.NewAPIClient(conn)},
 		Reputation: &Reputation{client: reputationPb.NewAPIClient(conn)},
 		Ffs:        &ffs{client: ffsPb.NewAPIClient(conn)},
+		Health:     &health{client: healthRpc.NewHealthClient(conn)},
+		Net:        &net{client: netRpc.NewNetClient(conn)},
```
asutula, line 45: Planning on making the rest of the client modules non-exported like this so they can only be accessed through the top level `Client`

api/client/net_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	n "github.com/textileio/powergate/net"
+	"github.com/textileio/powergate/net/rpc"
+)
+
+func TestNet(t *testing.T) {
+	c, done := setupNet(t)
+	defer done()
```
asutula, line 13: Not sure if the is the best setup, but other setups brought difficulties. This setup has one top level test func that creates the server and client and then all nested `t.Run` calls use those.

The other options I explored were:

- Creating stand alone tests that each created their own server and client. In this case, I was getting errors about the Prometheus port already being in use, presumably because each stand alone test was running with some amount of parallelism. I'd also imagine this would be quite heavy on the CI machine.
- Using `TestMain` to setup a single server and client shared between all stand alone tests. I actually really like this idea, but all of our code for setting up the server depends on a `testing.T` being passed in (which doesn't exist in `TestMain`). So this is possible, but required refactoring of a bunch of test setup code for creating the server.

Open to input here for sure.

net/rpc/api.go
```diff
@@ -73,7 +73,10 @@ func (a *API) Peers(ctx context.Context, req *PeersRequest) (*PeersReply, error)
 
 // FindPeer calls module.FindPeer
 func (a *API) FindPeer(ctx context.Context, req *FindPeerRequest) (*FindPeerReply, error) {
-	peerID := peer.ID(req.PeerID)
+	peerID, err := peer.Decode(req.PeerID)
```
asutula, line 76: Important update here to get sensical peer ids.

asutula:
Appears that the Prometheus port is going to be a problem when we expand the client tests, I guess because the tests run in parallel and Prometheus port is fixed. I'm very unclear about the behavior of go tests when it comes to parallelism because these client tests don't use `t.Parallel()` but they do seem to be running in parallel.

jsign:
> Appears that the Prometheus port is going to be a problem when we expand the client tests, I guess because the tests run in parallel and Prometheus port is fixed. I'm very unclear about the behavior of go tests when it comes to parallelism because these client tests don't use `t.Parallel()` but they do seem to be running in parallel.

Tests in different packages run in parallel by default. The `t.Parallel()` flag is for parallelization within the same packages. For serial execution we have the `-p 1` flag in the `make test`; that makes _all_ test run in serial.

asutula:
>Tests in different packages run in parallel by default.
The parallel tests I was experiencing were in the same package `client`.

jsign:
> > Tests in different packages run in parallel by default.
> 
> The parallel tests I was experiencing were in the same package `client`.

If you're sure `t.Parallel()` wasn't used in any of the tests, then there should be other thing going on. Test don't run in parallel in the same package without that flag. Maybe the _close_ logic of tests is revealing something that isn't being closed correctly?

api/client/net_test.go
jsign: I think it's better that each subtest runs a clean environment. The thing with sharing it is that one test might run correctly because other subtests leave a _convenient_ state, and test order might include other hidden magics.

Closing the server should liberate all resources that were created... that includes all files, ports, etc. So that might be an indication that some closing thing isn't closing. I you want I can take a look.

asutula:
Ok, thanks for your thoughts and thanks for taking a look at the http server thing @jsign. If you want to get that PR out, I'll hold off on this one until I merge your work in and update the tests here.

jsign:
> Ok, thanks for your thoughts and thanks for taking a look at the http server thing @jsign. If you want to get that PR out, I'll hold off on this one until I merge your work in and update the tests here.

[PR ready](https://github.com/textileio/powergate/pull/254).

asutula:
Ok @jsign I merged master in to get the http server fix, and I can now run client tests with no problems. Only potential problem is the time it takes to run the tests... If you do a fresh server startup and shutdown for each test, each one takes ~18 seconds per test on my laptop -- This will make the client tests take quite a while to run.

jsign:
> Ok @jsign I merged master in to get the http server fix, and I can now run client tests with no problems. Only potential problem is the time it takes to run the tests... If you do a fresh server startup and shutdown for each test, each one takes ~18 seconds per test on my laptop -- This will make the client tests take quite a while to run.

We could use free ports to spinup servers to all API endpoints and run in parallel since ipfs and lotus devnet are dedicated per test in docker. The only bottleneck in that scenario would be CI horsepower to enable some paralelism without killing it.

We can try later if at least with 2 parallel tests things are ok. That would cut times in half. Anyway, integration tests are heavy in general... but we can push to hardware limits and see if don't become too flaky.

asutula:
Cool, merged this for now and we can look at those optimizations asap.

PR #255, opened 2020-04-17
rpc, client, and cli for FFS remove and replace methods

asutula:
straight forward, no surprises here

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -1,6 +1,8 @@
 package client
 
 import (
+	"context"
+
 	dealsPb "github.com/textileio/powergate/deals/pb"
 	ffsRpc "github.com/textileio/powergate/ffs/rpc"
 	healthRpc "github.com/textileio/powergate/health/rpc"
@@ -25,6 +27,25 @@ type Client struct {
 	Health     *health
 	Net        *net
 	conn       *grpc.ClientConn
+}
+
+type AuthKey string
+
+type TokenAuth struct {
+	secure bool
+}
+
+func (t TokenAuth) GetRequestMetadata(ctx context.Context, _ ...string) (map[string]string, error) {
+	md := map[string]string{}
+	token, ok := ctx.Value(AuthKey("ffstoken")).(string)
+	if ok && token != "" {
+		md["X-ffs-Token"] = token
+	}
+	return md, nil
+}
+
+func (t TokenAuth) RequireTransportSecurity() bool {
+	return t.secure
 }
 
 // NewClient starts the client
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -40,7 +40,7 @@ func NewClient(target string, opts ...grpc.DialOption) (*Client, error) {
 		Deals:      &Deals{client: dealsPb.NewAPIClient(conn)},
 		Wallet:     &Wallet{client: walletPb.NewAPIClient(conn)},
 		Reputation: &Reputation{client: reputationPb.NewAPIClient(conn)},
-		Ffs:        &ffs{client: ffsRpc.NewFFSAPIClient(conn)},
+		Ffs:        &ffs{client: ffsRpc.NewFFSClient(conn)},
 		Health:     &health{client: healthRpc.NewHealthClient(conn)},
 		Net:        &net{client: netRpc.NewNetClient(conn)},
 		conn:       conn,
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -31,12 +31,15 @@ type Client struct {
 
 type ctxKey string
 
+// AuthKey is the key that should be used to set the auth token in a Context
 const AuthKey = ctxKey("ffstoken")
 
+// TokenAuth provides token based auth
 type TokenAuth struct {
 	secure bool
 }
 
+// GetRequestMetadata returns request metadata that includes the auth token
 func (t TokenAuth) GetRequestMetadata(ctx context.Context, _ ...string) (map[string]string, error) {
 	md := map[string]string{}
 	token, ok := ctx.Value(ctxKey(AuthKey)).(string)
@@ -46,6 +49,7 @@ func (t TokenAuth) GetRequestMetadata(ctx context.Context, _ ...string) (map[str
 	return md, nil
 }
 
+// RequireTransportSecurity specifies if the connection should be secure
 func (t TokenAuth) RequireTransportSecurity() bool {
 	return t.secure
 }
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -29,7 +29,9 @@ type Client struct {
 	conn       *grpc.ClientConn
 }
 
-type AuthKey string
+type ctxKey string
+
+const AuthKey = ctxKey("ffstoken")
 
 type TokenAuth struct {
 	secure bool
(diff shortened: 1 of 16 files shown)


asutula:
Not sure what to make of this test failure. They all pass when I run locally:
```
Error Trace:	health_test.go:15
9
        	Error:      	Expected nil, but got: &status.statusError{Code:2, Message:"can't resolve multiaddr location information", Details:[]*any.Any(nil), XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}
10
        	Test:       	TestCheck
```

jsign:
> Not sure what to make of this test failure. They all pass when I run locally:
> 
> ```
> Error Trace:	health_test.go:15
> 9
>         	Error:      	Expected nil, but got: &status.statusError{Code:2, Message:"can't resolve multiaddr location information", Details:[]*any.Any(nil), XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}
> 10
>         	Test:       	TestCheck
> ```

I ran the CI again and passed, so seems was a particular flakiness

asutula:
My intention here was to get the gRPC and CLI up to date with the API... I didn't realize `WatchLogs` wasn't in there yet. I'll add it now.

During the recent update of gRPC and CLI, I had removed the job monitoring from `pow ffs push`, so that is equivalent behavior to `pow ffs replace` now. We can always add the auto job monitoring in if we want.

jsign:
Yeah, without waiting for the job updates the client can't know if things went well or failed. Adding job monitoring as a flag sounds good.

api/client/ffs.go
```diff
@@ -236,6 +273,54 @@ func (f *ffs) Get(ctx context.Context, c cid.Cid) (io.Reader, error) {
 	return reader, nil
 }
 
+func (f *ffs) WatchLogs(ctx context.Context, ch chan<- LogEvent, c cid.Cid, opts ...WatchLogsOption) (func(), error) {
```
asutula, line 276: Hey @jsign I just added this client method for `WatchLogs`. I'm trying the same style as the API implementation on the server side where the user passes in the `chan`. Not that the user is passing in the `chan` do you see any way we can simplify the cancelation part of this, or is it still necessary to return a cancel func (or tell the user to cancel the context they passed in)?

(I was thinking maybe something like they could just close the `chan` they passed in, but I don't know if that is a legit pattern in Go)

jsign: Yes, the `cancel()` function can be avoided, and I think it _should_ be avoided.
The proper way to signal that the client is done is by `ctx` cancellation. 
Closing the channel would make the other part panic, since sending to a closed channel causes a panic.

That context cancellation is what's usually more idiomatic in Go, and it's what I used in on the server side.

asutula: Ok, got rid of those cancel functions. Made a little change to deal with the error that gets propagated to the client when it cancels the context. Will comment in the code and would like you feedback there.

api/client/client.go
```diff
@@ -27,6 +29,27 @@ type Client struct {
...
+	if ok && token != "" {
+		md["X-ffs-Token"] = token
+	}
+	return md, nil
+}
+
+func (t TokenAuth) RequireTransportSecurity() bool {
+	return t.secure
+}
+
```
asutula, line 52: Moved the auth stuff to the client so it can be used anywhere the client is used

api/client/ffs.go
```diff
@@ -236,6 +268,48 @@ func (f *ffs) Get(ctx context.Context, c cid.Cid) (io.Reader, error) {
...
+	}
+
+	stream, err := f.client.WatchLogs(ctx, &rpc.WatchLogsRequest{Cid: c.String(), Jid: config.jid.String()})
+	if err != nil {
+		return err
+	}
+	go func() {
+		for {
+			reply, err := stream.Recv()
+			if err == io.EOF || status.Code(err) == codes.Canceled {
```
asutula, line 284: Ok @jsign I added another error condition where you just want to close the channel and return. In this case when the user cancels the context they passed into this `WatchLogs` method, that cancels the server side process as well and returns an error of `context canceled`. In that case, I don't want to send a `LogEvent` containing the error because it is expected behavior. Made the same change in `WatchJobs` above.

api/client/ffs.go
```diff
@@ -118,33 +144,26 @@ func (f *ffs) Info(ctx context.Context) (*rpc.InfoReply, error) {
 	return f.client.Info(ctx, &rpc.InfoRequest{})
 }
 
-func (f *ffs) WatchJobs(ctx context.Context, jids ...ff.JobID) (<-chan JobEvent, func(), error) {
-	updates := make(chan JobEvent)
+func (f *ffs) WatchJobs(ctx context.Context, ch chan<- JobEvent, jids ...ff.JobID) error {
```
asutula, line 147: Moved this to taking the channel as a param, no returning a cancel func.

exe/cli/cmd/ffs_log.go
```diff
@@ -0,0 +1,73 @@
...
+
+func init() {
+	ffsLogCmd.Flags().StringP("token", "t", "", "FFS auth token")
+	ffsLogCmd.Flags().StringP("jid", "j", "", "Display information for only this job id")
+
+	ffsCmd.AddCommand(ffsLogCmd)
+}
+
+var ffsLogCmd = &cobra.Command{
+	Use:   "log [cid]",
```
asutula, line 25: Command for `WatchLogs`

asutula:
Hmm seem to be getting that `can't resolve multiaddr location information` all the time.

asutula:
Ok, added `-w` flag to `pow ffs push` and `pow ffs replace` that will automatically watch the resulting job.

asutula:
This is G2G now I think.

asutula:
Ok, tests magically passed.

api/client/ffs.go
jsign: Sounds good to me.

exe/cli/cmd/ffs_log.go
```diff
@@ -0,0 +1,73 @@
...
+		err = fcClient.Ffs.WatchLogs(authCtx(ctx), ch, cid, opts...)
+		checkErr(err)
+
+		c := make(chan os.Signal)
+		signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+		go func() {
+			<-c
+			cancel()
+			os.Exit(0)
+		}()
```
jsign, line 59: Idea for future: We can consider to always register a signal listener in the main cmd, and somehow have a nice helper ctx to subcommand that will magically autocancel ctx for them. So, commands will be signal-aware by default.

ffs/rpc/service.go
```diff
@@ -319,14 +319,14 @@ func (s *Service) WatchJobs(req *WatchJobsRequest, srv FFSAPI_WatchJobsServer) e
...
-func (s *Service) WatchLogs(req *WatchLogsRequest, srv FFSAPI_WatchLogsServer) error {
+func (s *Service) WatchLogs(req *WatchLogsRequest, srv FFS_WatchLogsServer) error {
 	i, err := s.getInstanceByToken(srv.Context())
 	if err != nil {
 		return err
 	}
 
 	var opts []api.GetLogsOption
-	if req.Jid != "" {
+	if req.Jid != ffs.EmptyJobID.String() {
```
jsign, line 329: 👍

exe/cli/cmd/ffs_log.go
jsign: @asutula , what do you think about this?
If that sounds good I can create an issue to remember it.

asutula: Yea that's a cool idea. Let's make an issue

jsign: Done #259.

PR #261, opened 2020-04-21
ffs: multiple addresses per instance

### LINKED ISSUE #258: Exploration: FFS Instance Multiple Addresses
Just typing out some thoughts and questions about how this might work...

Supporting multiple addresses is actually not looking too hard.

We can maintain a list of wallet addresses associated with the instance. Any operation that uses an address (`PushConfig`, `Replace`, etc) will need to verify that the address to be used exists in the list of instance addresses.

Methods that previously returned the instance wallet address or balance information will be converted to return a map or slice of address related information.

Main question I'm seeing is where and when to specify the address to use for a transaction. We can:

1. Add the wallet address as part of the `CidConfig` and `DefaultCidConfig` `Cold.FilConfig`. This provides a consistent way to specify the address, and gives us a place to store the default. Problem here is that it give less control or more difficult control over what address to use during a `Replace` operation since it will use the address that was previously used for that cid. Maybe that address is out of funds or no longer valid for some reason. This would require the user to realize that and push a new config for that cid before running `Replace`.
2. Require that an address is supplied with every call that needs one (`PushConfig`, `Replace`). We could still have a default if no address is specified. 

Another question is: Will it every be useful to know which address was used to store which cids? I don't think this information is already available since it was implied before to be the instance address. Do we want to add some tracking of wallet address <-> stored cid?

This also seems like an opportunity to add an optional display name associated with every address that is created. That looks like an easy change by modeling an address as a struct with `name` and `addr` fields, both will be persisted.

asutula:
Here's what I think is the minimum for supporting multiple addresses per ffs instance. I didn't try to create any new concepts for storing data, such as a "renew/repair address", I'm simply just storing a single address in the cold config for both the default settings and in the cid config that is kept for stored data. That address will be used for renew/repair for now. We can change that later easily.

closes #258


ffs/types.go
```diff
@@ -252,6 +252,8 @@ type FilConfig struct {
 	CountryCodes []string
 	// FilRenew indicates deal-renewal configuration.
 	Renew FilRenew
+	// Addr is the wallet address used to store the data in filecoin
+	Addr string
```
asutula, line 256: This is really the root of most of the changes. Added `Addr` string to `FilConfig`

ffs/api/api.go
```diff
@@ -48,18 +48,28 @@ type API struct {
 
 // New returns a new Api instance.
 func New(ctx context.Context, iid ffs.APIID, is InstanceStore, sch ffs.Scheduler, wm ffs.WalletManager, dc ffs.DefaultCidConfig) (*API, error) {
+	addr, err := wm.NewAddress(ctx, defaultAddressType)
+	if err != nil {
+		return nil, fmt.Errorf("creating new wallet addr: %s", err)
+	}
+
+	dc.Cold.Filecoin.Addr = addr
```
asutula, line 56: When we create a new ffs instance, we create a new address and set it as the default in the default config

ffs/api/api.go
```diff
@@ -48,18 +48,28 @@ type API struct {
...
+
+	addrInfo := AddrInfo{
+		Name: "Initial Address",
+		Addr: addr,
 	}
+
 	config := Config{
 		ID:               iid,
-		WalletAddr:       addr,
+		Addrs:            map[string]AddrInfo{addr: addrInfo},
```
asutula, line 69: We persist the new address in our map of addresses associated with this instance

ffs/api/api.go
```diff
@@ -96,9 +106,43 @@ func (i *API) ID() ffs.APIID {
...
-func (i *API) WalletAddr() string {
-	return i.cfg.WalletAddr
+// WalletAddr returns the wallet addresses.
+func (i *API) WalletAddrs() map[string]AddrInfo {
+	return i.cfg.Addrs
+}
+
+func (i *API) GetDefaultCfg() ffs.DefaultCidConfig {
+	return i.cfg.DefaultCidConfig
+}
```
asutula, line 116: Added this because it will be convenient to just fetch the default config without worrying about prepping for a cid.

ffs/api/api.go
```diff
@@ -96,9 +106,43 @@ func (i *API) ID() ffs.APIID {
...
+
+// NewAddr creates a new address managed by the FFS instance
+func (i *API) NewAddr(name string, options ...NewAddressOption) (string, error) {
```
asutula, line 119: API for creating a new address associated with this instance.

ffs/api/api.go
```diff
@@ -96,9 +106,43 @@ func (i *API) ID() ffs.APIID {
...
+	}
+
+	addr, err := i.wm.NewAddress(context.Background(), conf.addressType)
+	if err != nil {
+		return "", fmt.Errorf("creating new wallet addr: %s", err)
+	}
+	i.cfg.Addrs[addr] = AddrInfo{
+		Name: name,
+		Addr: addr,
+	}
```
asutula, line 138: Add it to our persistent map of instance addresses.

ffs/api/api.go
```diff
@@ -96,9 +106,43 @@ func (i *API) ID() ffs.APIID {
...
+	if conf.makeDefault {
```
asutula, line 139: Possibly set it as the default

ffs/api/api.go
```diff
@@ -96,9 +106,43 @@ func (i *API) ID() ffs.APIID {
...
+		i.cfg.DefaultCidConfig.Cold.Filecoin.Addr = addr
+	}
+	if err := i.is.PutConfig(i.cfg); err != nil {
```
asutula, line 142: persist it

ffs/api/api.go
```diff
@@ -128,7 +172,7 @@ func (i *API) SetDefaultCidConfig(c ffs.DefaultCidConfig) error {
 		return fmt.Errorf("default cid config is invalid: %s", err)
 	}
 	i.cfg.DefaultCidConfig = c
-	return nil
+	return i.is.PutConfig(i.cfg)
```
asutula, line 175: I think this was a bug in the wild... the newly set `DefaultCidConfig` wasn't being persisted, so just added that.

ffs/api/api.go
```diff
@@ -224,7 +276,12 @@ func (i *API) Replace(c1 cid.Cid, c2 cid.Cid) (ffs.JobID, error) {
 		return ffs.EmptyJobID, fmt.Errorf("getting replaced cid config: %s", err)
 	}
 	cfg.Cid = c2
-	jid, err := i.sched.PushReplace(i.iid, i.cfg.WalletAddr, cfg, c1)
+
+	if err := i.ensureValidColdCfg(cfg.Cold); err != nil {
```
asutula, line 280: For `Replace` and `PushConfig` we want to be sure the wallet address specified in the config is one actually managed by this instance (when necessary)

ffs/api/api.go
```diff
@@ -360,3 +420,21 @@ func (i *API) Close() error {
 	i.closed = true
 	return nil
 }
+
+func (i *API) ensureValidColdCfg(cfg ffs.ColdConfig) error {
```
asutula, line 424: This is used to check that if cold config is enabled, that the specified address is actually one managed by this instance

ffs/interfaces.go
```diff
@@ -29,12 +29,12 @@ var (
 type Scheduler interface {
 	// PushConfig push a new or modified configuration for a Cid. It returns
 	// the JobID which tracks the current state of execution of that task.
-	PushConfig(APIID, string, CidConfig) (JobID, error)
+	PushConfig(APIID, CidConfig) (JobID, error)
```
asutula, line 32: No need to pass in the address since it's part of the `CidConfig`

ffs/interfaces.go
```diff
@@ -29,12 +29,12 @@ var (
...
 
 	// PushReplace push a new or modified configuration for a Cid, replacing
 	// an existing one. The replaced Cid will be unstored from the Hot Storage.
 	// Also it will be untracked (refer to Untrack() to understand implications)
-	PushReplace(APIID, string, CidConfig, cid.Cid) (JobID, error)
+	PushReplace(APIID, CidConfig, cid.Cid) (JobID, error)
```
asutula, line 37: Same here

ffs/interfaces.go
```diff
@@ -95,15 +95,15 @@ type HotStorage interface {
 type ColdStorage interface {
 	// Store stores a Cid using the provided configuration and
 	// account address.
-	Store(context.Context, cid.Cid, string, FilConfig) (FilInfo, error)
+	Store(context.Context, cid.Cid, FilConfig) (FilInfo, error)
```
asutula, line 98: And here

ffs/interfaces.go
```diff
@@ -95,15 +95,15 @@ type HotStorage interface {
...
 
 	// Retrieve retrieves the data using an account address,
 	// and store it in a CAR store.
 	Retrieve(context.Context, cid.Cid, car.Store, string) (cid.Cid, error)
 
 	// EnsureRenewals executes renewal logic for a Cid under a particular
 	// configuration.
-	EnsureRenewals(context.Context, cid.Cid, FilInfo, string, FilConfig) (FilInfo, error)
+	EnsureRenewals(context.Context, cid.Cid, FilInfo, FilConfig) (FilInfo, error)
```
asutula, line 106: And here

ffs/rpc/service.go
```diff
@@ -269,8 +269,8 @@ func (s *Service) Info(ctx context.Context, req *InfoRequest) (*InfoReply, error
 				},
 			},
 			Wallet: &WalletInfo{
-				Address: info.Wallet.Address,
-				Balance: info.Wallet.Balance,
+				// Address: info.Wallet.Address,
+				// Balance: info.Wallet.Balance,
```
asutula, line 273: Just disabling stuff in the grpc service for now, will update once I'm confident of the underlying changes

ffs/scheduler/types.go
```diff
@@ -28,7 +28,6 @@ type JobStore interface {
 // Action represents an action to be executed by the Scheduler.
 type Action struct {
 	APIID       ffs.APIID
-	Waddr       string
```
asutula, line 31: Not needed since it's in the `Cfg`

ffs/api/api.go
```diff
@@ -96,9 +106,43 @@ func (i *API) ID() ffs.APIID {
 	return i.cfg.ID
 }
 
-// WalletAddr returns the Lotus wallet address.
-func (i *API) WalletAddr() string {
-	return i.cfg.WalletAddr
+// WalletAddr returns the wallet addresses.
+func (i *API) WalletAddrs() map[string]AddrInfo {
+	return i.cfg.Addrs
```
jsign, line 111: I think it would be better to return a copy of the map (guarded by locks).
Just in case the caller won't be messing with the same internal map and do some nasty thing.
Compared to other things this happens because `map`s are reference types.

ffs/api/api.go
```diff
@@ -96,9 +106,43 @@ func (i *API) ID() ffs.APIID {
...
+
+	conf := &NewAddressConfig{
+		makeDefault: false,
+		addressType: defaultAddressType,
+	}
+	for _, option := range options {
+		option(conf)
+	}
+
+	addr, err := i.wm.NewAddress(context.Background(), conf.addressType)
```
jsign, line 131: I think we should ask for a context in the method, since this is making a network call.

ffs/api/api.go
```diff
@@ -96,9 +106,43 @@ func (i *API) ID() ffs.APIID {
...
+	if err != nil {
+		return "", fmt.Errorf("creating new wallet addr: %s", err)
+	}
+	i.cfg.Addrs[addr] = AddrInfo{
```
jsign, line 135: I think we should check if that `name` wasn't used before and fail if that's the case. To avoid an accidental override.

ffs/api/api.go
jsign: Yep.

ffs/api/api.go
```diff
@@ -150,18 +194,26 @@ func (i *API) Info(ctx context.Context) (InstanceInfo, error) {
 	if err != nil {
 		return InstanceInfo{}, fmt.Errorf("getting pins from instance: %s", err)
 	}
-	balance, err := i.wm.Balance(ctx, i.cfg.WalletAddr)
-	if err != nil {
-		return InstanceInfo{}, fmt.Errorf("getting balance of %s: %s", i.cfg.WalletAddr, err)
+
+	wallet := make([]BalanceInfo, len(i.cfg.Addrs))
+	j := 0
+	for _, addr := range i.cfg.Addrs {
```
jsign, line 200: You can use `for j, addr := ..`

ffs/api/types.go
```diff
@@ -26,25 +26,52 @@ type InstanceStore interface {
...
+	Name string
+	Addr string
+}
+
 // InstanceInfo has general information about a running Api instance.
 type InstanceInfo struct {
 	ID               ffs.APIID
 	DefaultCidConfig ffs.DefaultCidConfig
-	Wallet           WalletInfo
+	Wallet           []BalanceInfo
```
jsign, line 42: Nit: Naming seems weird. Should the attributes be called `Balances`? Or the type `WalletInfo`.

ffs/scheduler/scheduler.go
```diff
@@ -85,9 +85,6 @@ func (s *Scheduler) push(iid ffs.APIID, waddr string, cfg ffs.CidConfig, oldCid
 	if err := cfg.Validate(); err != nil {
 		return ffs.EmptyJobID, fmt.Errorf("validating cid config: %s", err)
 	}
-	if waddr == "" {
-		return ffs.EmptyJobID, fmt.Errorf("invalid wallet address")
-	}
```
jsign, line 90: Yeah, this was mostly a safety net.

Do you mind adding this check now in `ffs/types.go` on `func (cc ColdConfig) Validate()`?

ffs/api/api.go
asutula: We can do that. Just to be clear, a duplicate name won't override any other address since they are keyed by address, not name. But you would end up with duplicate names which may not be ideal.

ffs/api/api.go
asutula: Oh, I thought `j` would be the `string` key of the map in that case.

asutula: (not an `int` index since we're ranging over a `map`)

jsign: Oh, thought was a slice. Dismiss!

ffs/api/api.go
jsign: Agree!

ffs/api/api.go
asutula: Cool, done.

ffs/api/api.go
asutula: anyway, works better just using `append`

ffs/api/types.go
asutula: I like the idea of `Balances` for the attribute

ffs/scheduler/scheduler.go
asutula: 👍

ffs/api/api.go
```diff
@@ -216,6 +289,7 @@ func (i *API) WatchJobs(ctx context.Context, c chan<- ffs.Job, jids ...ffs.JobID
 // Replace pushes a CidConfig of c2 equal to c1, and removes c1. This operation
 // is more efficient than manually removing and adding in two separate operations.
 func (i *API) Replace(c1 cid.Cid, c2 cid.Cid) (ffs.JobID, error) {
+	// ToDo: Should we be locking here?
```
asutula, line 292: Should we?

jsign: Yes, sounds good.

api/client/ffs.go
```diff
@@ -22,48 +23,55 @@ type JobEvent struct {
 	Err error
 }
 
-// PushConfigOption mutates a push configuration.
-type PushConfigOption func(o *PushConfig)
+// NewAddressOption is a function that changes a NewAddressConfig
+type NewAddressOption func(r *rpc.NewAddrRequest)
```
asutula, line 27: Found these options functions can operate on a rpc request nicely, no need for a separate config struct.

ffs/api/api.go
```diff
@@ -120,15 +183,15 @@ func (i *API) GetCidConfig(c cid.Cid) (ffs.CidConfig, error) {
 	return conf, nil
 }
 
-// SetDefaultCidConfig sets a new default CidConfig.
-func (i *API) SetDefaultCidConfig(c ffs.DefaultCidConfig) error {
+// SetDefaultConfig sets a new default CidConfig.
+func (i *API) SetDefaultConfig(c ffs.DefaultConfig) error {
```
asutula, line 187: Renamed things like "DefaultCidConfig" just simply "DefaultConfig" because in the default case, there is nothing about a cid in there and it was confusing since there is also `GetDefaultCidConfig` which prepares a `CidConfig` from the default using a cid. This makes it easier to understand.

net/rpc/api.go
```diff
@@ -33,7 +33,7 @@ func (a *API) ListenAddr(ctx context.Context, req *ListenAddrRequest) (*ListenAd
 	}
 
 	return &ListenAddrReply{
-		AddrInfo: &AddrInfo{
+		AddrInfo: &PeerAddrInfo{
```
asutula, line 36: See https://github.com/textileio/powergate/issues/262 about why I renamed this

PR #357, opened 2020-04-24
avoid ChainNotify & update to newer interopnet

### LINKED ISSUE #323: lotus-client: use Lotus directly
Consider if a PR landed yesterday in `testnet/3` (not yet `interopnet`) makes the API completely CGO free and allows to abandon `lotus-client` maintenance, and can be used directly.
It would be good.

### LINKED ISSUE #359: deals: avoid ChainNotify
ChainNotify stream caused some problems in parallel tests in which the Lotus client closed inadvertently the stream, making the logic fall into fallback scenarios which aren't nice.

The block output rate of the chain can be modeled with a ticker interval that is cross-configured in Powergate. This strategy isn't new, the same decision was made on indexes... this was the remaining use-case of ChainNotify.

jsign:
- Fixes #359
- Set unique tmp dir per test package since now they run in parallel and that was causing some problems. 

Also, a new version of `interopnet` was released again:
- Fixes #323: Good news, I don't have to maintain [textileio/lotus-client](https://github.com/textileio/lotus-client) anymore; archived it.
- Update to new bootstrap peers from new interopnet network.
- Switched to new `lotus-devnet` from new `interopnet` version. Involved doing: https://github.com/textileio/lotus-devnet/pull/11


chainsync/chainsync.go
```diff
@@ -4,9 +4,15 @@ import (
...
-	"github.com/textileio/lotus-client/api/apistruct"
-	"github.com/textileio/lotus-client/chain/store"
+)
+
+const (
+	hcApply = "apply"
+	// For completeness:
+	// hcRevert = "revert"
+	// hcCurrent = "current"
 )
```
jsign, line 16: Now `lotus/api` is CGO free. But of course not all other subpackages. The `chain/store` subpackage isn't CGO free and here was used to access some constants. I moved those constants here to avoid a single constants use break all PG with CGO stuff.

deals/deals.go
```diff
@@ -184,22 +185,14 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
 		return nil, fmt.Errorf("proposals list can't be empty")
 	}
 	ch := make(chan DealInfo)
```
jsign, line 187: #359

jsign:
Will merge to unblock Marcus. Still open to reviews.

PR #365, opened 2020-04-28
Expand CidInfo queries

asutula:
Trying to support more ways of slicing and dicing `CidInfo` results.


ffs/scheduler/cistore/cistore.go
```diff
@@ -59,6 +62,32 @@ func (s *Store) Put(ci ffs.CidInfo) error {
 	return nil
 }
 
+// Query returns a list of CidInfo
+func (s *Store) Query(selector func(ffs.CidInfo) bool) ([]ffs.CidInfo, error) {
```
asutula, line 66: Hey @jsign. My goal is to be able to query for all `CidInfo` associated with a single `ffs` instance. I don't think this does it because the `cistore.Store` is owned buy the scheduler, and data for all `ffs` instances is written to it. There isn't anything about the keys scoping it to an `ffs` instance. 

The ffs instance store `istore.Store` does persist a list of cids associated with the ffs instance. I could query that list, then for each cid, make a call to `cistore.Get` in here. That feels really inefficient to me, and may come with performance problems when there are lots of cids. But I'm also not too familiar with the performance characteristics of kv stores like Badger. Am I overthinking this?

jsign: Yes, the last paragraph is the correct approach. If that becomes a real problem, we have plenty of options.

ffs/api/api.go
```diff
@@ -209,6 +209,29 @@ func (i *API) Show(c cid.Cid) (ffs.CidInfo, error) {
 	return inf, nil
 }
 
+// ShowMatching returns a list of information about stored cids matching a selector
+func (i *API) ShowMatching(selector func(ffs.CidInfo) bool) ([]ffs.CidInfo, error) {
```
asutula, line 213: Ok @jsign here is the implementation of that method to get all `CidInfo` matching a selector. 

There are a handful of ways the web ui work requires this data to be processed and returned to the client. Since those requirements are driven only by the web client at this point, I'm considering not implementing any further methods here on `ffs.API` and instead just using this method within the gRPC service and adding only new gRPC methods for the client. What do you think?

ffs/rpc/service.go
```diff
@@ -530,6 +530,33 @@ func (s *Service) AddToHot(srv FFS_AddToHotServer) error {
 	return srv.SendAndClose(&AddToHotReply{Cid: c.String()})
 }
 
+// ActiveMiners returns information about the miners the specified FFS instance has deals with
+func (s *Service) ActiveMiners(ctx context.Context, req *ActiveMinersRequest) (*ActiveMinersReply, error) {
```
asutula, line 534: Here's an example of the kind of method that is going to be implemented to the ui requirements. It used the general query method to get all the `CidInfo`s for the ffs instance, then groups stored cids by miner.

asutula: If we do go down this path of implementing some things for the web ui only in RPC layer, we could even create a separate rpc package/service just for this category of things. This might help keep straight what is directly exporting the ffs api vs more custom client driven requirements. Overkill?

jsign: > If we do go down this path of implementing some things for the web ui only in RPC layer, we could even create a separate rpc package/service just for this category of things. This might help keep straight what is directly exporting the ffs api vs more custom client driven requirements. Overkill?

Totally agree, good idea to avoid exploding ffs with custom stuff.

ffs/api/api.go
jsign: Sounds good!

asutula:
This has been left behind, not clear it's really required at this point. Closing.

PR #378, opened 2020-05-04
FFS Send Fil API

### LINKED ISSUE #275: ffs: send funds from managed address to any address

asutula:
Send fil from a managed address to any other address

closes #275


wallet/wallet.go
```diff
@@ -74,3 +74,24 @@ func (m *Module) Balance(ctx context.Context, addr string) (uint64, error) {
...
+	t, err := address.NewFromString(to)
+	if err != nil {
+		return err
+	}
+	msg := &types.Message{
+		From:     f,
+		To:       t,
+		Value:    types.BigInt{Int: amount},
+		GasLimit: 1000, // ToDo: how to handle gas?
+		GasPrice: types.NewInt(0),
```
asutula, line 93: Not sure if now is the time or place to deal with gas. I see there is a method to estimate gas, but I don't really understand the scope of this for now.

jsign: Last time I read about that, was still a ToDo in Lotus.
Better to leave this as it is until we know the correct way to do that... good default to be 0.

ffs/integrationtest/integration_test.go
```diff
@@ -1027,6 +1028,55 @@ func TestRemove(t *testing.T) {
...
+		require.Nil(t, err)
+		return bal == uint64(iWalletBal+amt)
+	}
+
+	require.Eventually(t, hasInitialBal, time.Second*5, time.Second)
+
+	err = fapi.SendFil(ctx, addr1, addr2, big.NewInt(amt))
+	require.Nil(t, err)
+
+	require.Eventually(t, hasNewBal, time.Second*5, time.Second)
```
asutula, line 1077: The `Eventually` api is nice

wallet/wallet.go
asutula: 👍

api/client/ffs.go
```diff
@@ -325,6 +325,16 @@ func (f *ffs) WatchLogs(ctx context.Context, ch chan<- LogEvent, c cid.Cid, opts
 	return nil
 }
 
+func (f *ffs) SendFil(ctx context.Context, from string, to string, amount int64) error {
```
jsign, line 328: Add some comment to describe that this send is async.
I'm starting to think we should comment all client APIs, because when the client tries to use a method it doesn't have much clue of what that _client_ API will do. 

I wonder why this public method doesn't bubble up in the golangci check.

ffs/api/api.go
```diff
@@ -435,6 +436,14 @@ func (i *API) WatchLogs(ctx context.Context, ch chan<- ffs.LogEntry, c cid.Cid,
 	return nil
 }
 
+// SendFil sends fil from a managed address to any another address
```
jsign, line 439: Nit: mention is async, aka: not confirmed.

ffs/integrationtest/integration_test.go
jsign: Nice! Didn't notice it exists. I can imagine changing some thing in other tests to use this.

api/client/ffs.go
asutula: It's because the `ffs` struct isn't public (although it is exported through a public field on `Client`). I started doing this to keep people from using the `ffs` struct directly, forcing them instead to use `Client` to access `ffs` apis. Maybe it's not worth the way it breaks linting. I can PR the change that makes it public and adds missing comments. Let me know what you think

jsign: Oh, I see... `ffs.Ffs` is a public attribute but the type is private.
Agree that might be better to make them public, since I think the comment methods for the client are a must for good DX.

PR #379, opened 2020-05-05
update bootstrap peers

jsign:
Missed update of bootstrap peers.


jsign:
Merging since it's only bootstrappers update.

PR #384, opened 2020-05-06
various usage fixes

asutula:
- we were missing `repairable` on the rpc `CidConfig` model, so added that
- return a proper `ffs.CidConfig` from `Client` so that it and its cid can be properly marshaled to json and displayed in the cli
- the `pow ffs config` command was attached to the wrong place in the command tree
- various wording and env var naming improvements
- added `repairable` to `DeafaultConfig` because that should have a default


api/client/ffs.go
```diff
@@ -160,8 +162,41 @@ func (f *FFS) NewAddr(ctx context.Context, name string, options ...NewAddressOpt
...
+	res, err := f.client.GetDefaultCidConfig(ctx, &rpc.GetDefaultCidConfigRequest{Cid: c.String()})
+	if err != nil {
+		return ff.CidConfig{}, err
+	}
+	resCid, err := cid.Decode(res.Config.Cid)
+	if err != nil {
+		return ff.CidConfig{}, err
+	}
+	return ff.CidConfig{
+		Cid:        resCid,
```
asutula, line 175: We want to return a real cid here so it can be properly marshaled to json by the cli or any other user

docker/docker-compose-embedded.yaml
```diff
@@ -19,9 +19,9 @@ services:
       - ipfs
       - lotus
     environment:
-      - TEXPOWERGATE_EMBEDDED=true
-      - TEXPOWERGATE_LOTUSHOST=/dns4/lotus/tcp/7777
-      - TEXPOWERGATE_IPFSAPIADDR=/dns4/ipfs/tcp/5001
+      - POWD_EMBEDDED=true
+      - POWD_LOTUSHOST=/dns4/lotus/tcp/7777
+      - POWD_IPFSAPIADDR=/dns4/ipfs/tcp/5001
```
asutula, line 24: Made some consistent naming pattern across all Cobra/Viper env vars

exe/cli/cmd/ffs_push.go
```diff
@@ -19,8 +19,8 @@ import (
 
 func init() {
 	ffsPushCmd.Flags().StringP("token", "t", "", "FFS access token")
-	ffsPushCmd.Flags().StringP("config", "c", "", "Optional path to a file containing cid storage config json, falls back to stdin, uses FFS default by default")
-	ffsPushCmd.Flags().BoolP("override", "o", false, "Path to a file containing cid storage config json")
+	ffsPushCmd.Flags().StringP("conf", "c", "", "Optional path to a file containing cid storage config json, falls back to stdin, uses FFS default by default")
```
asutula, line 22: This was conflicting with the global `--config` flag

exe/cli/cmd/root.go
```diff
@@ -24,9 +24,9 @@ var (
 	cmdTimeout = time.Second * 10
 
 	rootCmd = &cobra.Command{
-		Use:   "filecoin",
```
jsign, line 27: Ups :P

PR #389, opened 2020-05-11
Detailed Deal errors & others

### LINKED ISSUE #390: ffs/scheduler: include Deal Errors as outputs
If the Cold Storage is enabled, a Job execution will succeed if at least one proposed deals succeeded (it's up to the repair feature to reconsider if further work should be enforced). If all proposed deals failed, the Job is considered _failed_ (we can't consider it succeed if the data wasn't at least stored in 1 miner).

That means that proposed deals errors can happen in success or failed job executions. We should enable exposing these failures and _causes_ to the client to have better visibility of:
- If the Job _Failed_, see the underlying Lotus messages about why it failed... so it will hint the client what went wrong, e.g: insufficient funds for storing the data.
- If the Job _Succeeded_, it means that the data was stored in at least one proposed deals... but it may be useful to see of others failed and why. This could mean that maybe a miner isn't willing to accept the deal (and there's nothing fundamentally wrong with the proposal or funds).

The error are _per-proposed failed deal_.

Apart from exposing this information inside _Job_ data, it also gets included in the human-friendly logger for the Cid.

jsign:
- Closes #390 (see the issue for an explanation)
- Update to `go-ipfs:v0.5.1`
- Some docs improv.
- Renamed Job status _InProgress_ to _Executing_ to be single worded and a more usual feel about Job executions common knowledge.


ffs/integrationtest/integration_test.go
```diff
@@ -1141,6 +1141,28 @@ func TestRepair(t *testing.T) {
 	}
 }
 
+func TestFailedJobMessage(t *testing.T) {
```
jsign, line 1144: An test (and example) to check about Deal Erros happened during a Job executiong.
In this test, we propose storing data bigger than the sector size which will definitely fail.

Now we have `DealErrors` field in the _Job_, which allows to inspect why the Job failed with a reasonable message from the Lotus node. These errors aren't _enumed_ in Lotus, only available as strings.

PR #391, opened 2020-05-12
wallet: consider that NewFromString returns Undef address 

jsign:


wallet/wallet.go
```diff
@@ -44,9 +44,9 @@ func (m *Module) NewAddress(ctx context.Context, typ string) (string, error) {
 		return "", err
 	}
 
-	if m.masterAddr != nil {
+	if m.masterAddr != address.Undef {
```
jsign, line 47: Here was a little problem. `address.NewFromString` returns an `Undef` address, not an error nor nil.

api/client/ffs.go
```diff
@@ -542,9 +541,13 @@ func toRPCColdConfig(config ff.ColdConfig) *rpc.ColdConfig {
 func protoToDealError(des []*rpc.DealError) ([]ffs.DealError, error) {
 	res := make([]ffs.DealError, len(des))
 	for i, de := range des {
-		propCid, err := cid.Decode(de.ProposalCid)
-		if err != nil {
-			return nil, fmt.Errorf("parsing proposal cid: %s", err)
+		var propCid cid.Cid
```
jsign, line 544: Took advantage to cover from a border case.

PR #392, opened 2020-05-13
Update to new interopnet

jsign:
Required https://github.com/textileio/lotus-devnet/pull/16 (plus 5 direct commits to `master` since things got hairy...), and merging plus other deps updates and generated docker images.
Nothing much interesting on this repo... `lotus-devnet` needed some changes.


deals/deals_test.go
```diff
@@ -160,8 +160,13 @@ func checkErr(t *testing.T, err error) {
...
-	buf := make([]byte, size)
 	r := rand.New(rand.NewSource(22))
+	return randomBytesFromSource(size, r)
+}
+
+func randomBytesFromSource(size int, r *rand.Rand) []byte {
+	buf := make([]byte, size)
 	_, _ = r.Read(buf)
 	return buf
+
```
jsign, line 171: This was left from a custom code I had on my local machine for testing a weird scenario.
Not hurting.

tests/ldevnet.go
```diff
@@ -37,6 +39,28 @@ func LaunchDevnetDocker(t *testing.T, numMiners int) *dockertest.Resource {
 			panic(fmt.Sprintf("couldn't purge lotus-devnet from docker pool: %s", err))
 		}
 	})
+	debug := false
```
jsign, line 42: Leaving this also. It allows to interleave the docker container runs on stdout. By default I left it disabled, but it's handy for looking when things break.

PR #397, opened 2020-05-14
ffs: get Cid log history API

### LINKED ISSUE #396: FFS instance log history API
We currently have an API to subscribe to the _CidLogger_ output to have friendly logs about the Job execution. We should add a way to also retrieve history.
Like, client would retrieve history and then subscribe to keep watching for progress, so it can have a full picture.

jsign:
Closes #396 
Required for Buckets integration, so to have better UX.

In the future, we can include some extra config for limiting.


jsign:
> Awesome. I think this is good to go. One question to think about for a possible enhancement...
> 
> What if I want to get the existing log history and know that is all that exists for now and exit? With this setup, you have no way of knowing. But I'm also not sure this question needs to be answered for any existing use case. Will have to dig into the web ui features more.

Sure, for that case we can create another non-stream based API that simply returns a slice. Fortunately, this PR already creates the `GetLogs` that would be called internally, so we'd only need to create the gRPC endpoint.

PR #408, opened 2020-05-18
Gen js grpc lib + proto cleanup

asutula:
This does a few things:

- Adds a script and CI config for building and publishing a npm module containing the JS/TS gRPC bindings. It does this all from scratch, so no skeleton project files in the repo or anything like that.
- Had to fix a couple namespacing issues in the `.proto` files, so this led to to standardizing the `.proto` files and related project structure a bit. Things are now consistent all over the repo.
- Create commands in the top level `Makefile` that clean/generate the Go proto output for the whole repo. Removed all the module-specific `Makefile`s


.github/workflows/publish-grpc-libs.yml
```diff
@@ -0,0 +1,39 @@
+name: Publish JS gRPC bindings
```
asutula, line 1: No good way to test this, so will have to tag an alpha release once this is merged to see what happens.

reputation/rpc/api.proto
```diff
@@ -1,10 +1,7 @@
 syntax = "proto3";
-package filecoin.reputation.pb;
+package reputation.rpc;
 
-option java_multiple_files = true;
-option java_package = "io.textile.filecoin.reputation.pb";
-option java_outer_classname = "FilecoinReputation";
-option objc_class_prefix = "TTE";
+option go_package = "rpc";
```
asutula, line 4: Useful option I didn't know about. Useful because you really do need to make the `package` above unique and consistent with the directory the .proto file is in. This isn't a problem in Go, but is for JS and other languages.

scripts/gen-js-protos.sh
```diff
@@ -0,0 +1,36 @@
+LIB_VERSION=$1
```
asutula, line 1: Simple script for generating the node module containing the js grpc bindings

scripts/protoc_gen_plugin.bash
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env bash
+
+# From https://github.com/bufbuild/buf/blob/master/make/go/scripts/protoc_gen_plugin.bash
```
asutula, line 3: Awesome script for generating all grpc bindings in a very flexible and configurable way. Just using in the top level `Makefile` to generate the go bindings for now

scripts/gen-js-protos.sh
jsign: Cool 🧙

asutula:
Just to make this clear, every module now contains this:
```
.
├── rpc
│   ├── rpc.go
│   ├── rpc.pb.go
│   └── rpc.proto
```

scripts/protoc_gen_plugin.bash
jsign: Uhm, pretty cool. Agree that this recursive proto-gen will be valuable in other repos.

PR #411, opened 2020-05-20
use new devnet

jsign:
Use the new devnet which:
- Updated to master
- Using some bleeding edge sector-builder & filecoin-ffi deps to:
   - Allow the devnet be long running and not get problems with Windowed PoSts on sectorbuilder mock.
   - Use 512Mib sectors, which allow to store bigger files.

The `textile/lotus-devnet` supports a new flag `bigsectors` which if enabled will use 512Mib sectors. The `make embed` setup enables this flag, so if you run a common flow in the CLI you can store files up to ~512Mb. (I tested with 100mb, and works).

Refer https://github.com/textileio/lotus-devnet/pull/18 to know more details.


deals/deals.go
```diff
@@ -103,7 +105,7 @@ func (m *Module) Store(ctx context.Context, waddr string, data io.Reader, dcfgs
 				Root: dataCid,
 			},
 			MinBlocksDuration: dur,
-			EpochPrice:        types.NewInt(c.EpochPrice),
+			EpochPrice:        types.NewInt(uint64(math.Ceil(2 * gbSize * float64(c.EpochPrice)))),
```
jsign, line 108: So, turns out that is a better calculation of the `EpochPrice` to propose.
This will be improved over time.

TL;DR of the calculation:
What we receive in the `dcfgs` is the desired `EpochPrice` of the miners *per Gib, per epoch*.
So the calculation for the stored epoch price its quite simple:
*bytes -> transform to Gib -> multiply per Epoch Per Price desired by miner*

Now, there're are other things that should be solved. The real _Piece Size_ might have some Padding. Also, if the stored thing is in CAR format or not, also affect the calculation. The CAR Format is very close to the ending size; but without CAR the ending size is: original size -> create UnixFS DAG -> CAR the DAG -> store that... so a bit harder to estimate.

For the moment, I'm leaving here the double of the "correct" calculation. I'd raise some questions to Lotus guys about some of this challenges.

deals/deals_test.go
```diff
@@ -95,7 +95,7 @@ func storeMultiMiner(m *Module, client *apistruct.FullNodeStruct, numMiners int,
 	for i := 0; i < numMiners; i++ {
 		cfgs[i] = StorageDealConfig{
 			Miner:      miners[i].String(),
-			EpochPrice: 1000000,
+			EpochPrice: 500000000,
```
jsign, line 98: The miner in the devnet is using now the default ask price, which is 500m attofil. 
Unfortunately, I had to remove a call which lowered his price since it has a bug (that I'll fix in Lotus in some days).
To avoid doing that PR, needed to deal with the default price which is pretty high.

exe/server/main.go
```diff
@@ -43,7 +43,7 @@ func main() {
 	pflag.String("repopath", "~/.powergate", "repo-path")
 	pflag.Bool("embedded", false, "run in embedded ephemeral FIL network")
 	pflag.String("ipfsapiaddr", "/ip4/127.0.0.1/tcp/5001", "ipfs api multiaddr")
-	pflag.Int64("walletinitialfund", 4000000000, "created wallets initial fund in attoFIL")
+	pflag.Int64("walletinitialfund", 4000000000000000, "created wallets initial fund in attoFIL")
```
jsign, line 46: More money per FFS instance, considering that the miner now is more expensive and now we support bigger files.
Tested with a 100mb file (on a `make embed`) and all good. 👍

docker/docker-compose-embedded.yaml
```diff
@@ -32,10 +32,11 @@ services:
       - 5001:5001
 
   lotus:
-    image: textile/lotus-devnet:sha-7ed2544
+    image: textile/lotus-devnet:sha-1e166c9
     ports:
       - 7777:7777
     environment:
       - TEXLOTUSDEVNET_SPEED=1500
+      - TEXLOTUSDEVNET_BIGSECTORS=true
```
jsign, line 40: The devnet now allos to specify if you want to use "big sectors", so supporting 512Mib sectors.
By default uses 2Kb sectors, which is what is running for the CI tests... tried 512Mib but it's too heavy for parallelization.

exe/server/main.go
sanderpick: 🥳

jsign:
Thanks!

PR #414, opened 2020-05-25
Leverage Lotus -> IPFS integration & more improvements

### LINKED ISSUE #190: deals: retrieval/importing temp file management
Currently, the `DealModule` saves fetched files from Filecoin in the _repoPath_ with a `retrieve-*` temp file style. This file isn't deleted.

Some ideas to build on this:
- This file could be leveraged if a `Retrieve()` call is made again, so to avoid Fil fetching (acts like a cache)
- Seems reasonable to add some _prune_ feature to the module to delete these tempfiles/cache.

### LINKED ISSUE #388: ffs: use Lotus -> IPFS integration
It seems that now Lotus can directly import a Cid from an external IPFS node. 
This is a good thing to evaluate and leverage in FFS.
[See here](https://github.com/filecoin-project/lotus/pull/1641).

jsign:
This PR leverages initial support of IPFS by the Lotus node, which we extended later to a more powerful one. 

Now Lotus can use a remote IPFS node to get data for new deals, and to store retrieved deal data. This was originally not possible, so we proposed [this](https://github.com/filecoin-project/lotus/pull/1843) PR.

This had some deep implications for FFS implementation:
- Now things are wired to let Lotus connect to the _HotStorage_ ipfs node directly.
- Allows to start deals without _sending_ the data to Lotus via files, which:
   - Produce less _garbage_ since those files needed to be cleaned eventually.
   - Disk capacity on PG isn't needed for making deals.
   - Less work is done (so CPU/mem), so making deals for big files should be faster and more concurrent/parallel Jobs can be executed.
- Since we added a new feature to "fetch" deal data, we can indicate to Lotus that we want to get Deals data but only store it in IPFS, which:
   - Lotus doesn't need to save the retrieved data to a file, to be used by PG.
   - PG doesn't need to _unCAR_ the data retrieval or do any kind of extra processing. It can assume that the Deal data was automatically fed to the _HotStorage_ IPFS node. It only needs to indicate to _pin_ the Cid (if wanted).
- `make embed` uses the devnet connected with the IPFS node. The devnet is configured to use bigsectors.

All this change is making the following "big picture" becoming more true:
- PG: orchestrates Lotus and IPFS to have a smart way of storing and retrieving data. No mandatory data transfers are needed to be done by PG. (i.e: the addToHot step is really optional since you can feed the data directly to IPFS by other means if wanted).
- Lotus: only handle chain data. Lean on IPFS for raw deals data (stored and retrieved).
- IPFS: be responsible for storing raw data to be used in new Deals, or retrieved Deal data.

---

Regarding the Deals module, which is a lower-layer module (FFS agnostic):
- This module can be continued to be used without assumptions about Lotus being integrated with IPFS.
- The above keeps _the problem_ of having to deal with tmp-files managing about retrieved Deals. Now the `io.ReadCloser` that's returned by `dealsmodule.Retrieve(..)` will _autodelete_ the file when the client closes the reader. So to auto-clean up things.
- A new API `Import` was created. This allows `Store` to assume the data was imported previously, via `Import` that was done directly by PG to Lotus, or indirectly by feeding IPFS where Lotus is connected to.
- A new API `Fetch` allows to indicate to Lotus that we want to retrieve a Deals data but, compared with `Retrieve`, it indicates that should only feed that data to its underlying blockstorage. This is useful when using a Lotus node that it's connected to a IPFS node, since Lotus will directly store the data to the IPFS node. The `Retrieve` API goes a step further expects that data to be available in a file (which FFS isn't interested).

 ---

Regarding integration-tests:
- All FFS tests use the new Lotus -> IPFS integration since this will be the defacto configuration for FFS since it's the most optimal.
- Two skipped tests where unskipped since Lotus evolved in solving some problems that justified them being skipped sometime before. (So no skipped tests now).

---

Regarding deal renewal, I changed the implementation to let the renewal be made on the same miner where the deal will soon be expired. This wasn't possible before since the miners wouldn't allow to re-store data that are still active (most prob a bug). I checked now that this doesn't seem to be the case anymore, so now they're renewed in the same miner; which is a more natural action for renewals. If we think it might be a good idea to not necessarily renew on the same miners, we can add some new knob in the future.

---

Some extra changes:
- The `powd` flags description was rewritten.
- Readme was refreshed, might be more interesting to [see it rendered](https://github.com/textileio/powergate/blob/a1b9efbe7d0d34aca048e46a211714d500d2c864/README.md).
- The concept of "embed" network was changed to "devnet", so `--devnet`, `make devnet`, compared to `--embedded`, `make embed`.

---

- Closes #388 
- Closes #190 
- Required https://github.com/textileio/powergate/pull/414/
- Required https://github.com/filecoin-project/lotus/pull/1843


api/client/ffs.go
```diff
@@ -7,7 +7,6 @@ import (
 
 	cid "github.com/ipfs/go-cid"
 	"github.com/textileio/powergate/ffs"
-	ff "github.com/textileio/powergate/ffs"
```
jsign, line 10: Duplicated import, voted to leave unaliased one.

exe/server/main.go
```diff
@@ -32,19 +29,18 @@ var (
 )
 
 func main() {
```
jsign, line 31: Improve description messages.
Eventually might worth switching to a more powerful CLI framework for the server? 
I wouldn't expect much flags to be available, but might provide nicier output of `--help` commands? 
Something to consider.

deals/deals.go
```diff
@@ -75,20 +73,28 @@ func (m *Module) Store(ctx context.Context, waddr string, data io.Reader, dcfgs
...
 	ref := api.FileRef{
 		Path:  f.Name(),
 		IsCAR: isCAR,
 	}
 	dataCid, err := m.api.ClientImport(ctx, ref)
 	if err != nil {
-		return cid.Undef, nil, fmt.Errorf("error when importing data: %s", err)
+		return cid.Undef, 0, fmt.Errorf("error when importing data: %s", err)
 	}
+	return dataCid, size, nil
```
asutula, line 86: Ok, just want to make sure I understand... `Import` calls the Lotus client `ClientImport`, passing the file name and `isCar`. Lotus `ClientImport` adds the data to the Lotus node's configured IPFS node? Is the data simply added, or pinned as well?

deals/deals.go
```diff
@@ -124,44 +130,58 @@ func (m *Module) Store(ctx context.Context, waddr string, data io.Reader, dcfgs
...
 	}
-	return dataCid, res, nil
+	return res, nil
+}
+
+// Fetch fetches deal data to the underlying blockstore of the Filecoin client.
+// This API is meant for clients that use external implementations of blockstores with
+// their own API, e.g: IPFS.
+func (m *Module) Fetch(ctx context.Context, waddr string, cid cid.Cid) error {
+	return m.retrieve(ctx, waddr, cid, nil)
```
asutula, line 140: Just to be more explicit about what our current implementation is actually doing (for my own understanding)... This causes a retrieval deal to be executed and adds the resulting data to the Lotus node's IPFS node?

deals/deals.go
```diff
@@ -124,44 +133,58 @@ func (m *Module) Store(ctx context.Context, waddr string, data io.Reader, dcfgs
...
+
+	if err := m.retrieve(ctx, waddr, cid, &ref); err != nil {
+		return nil, fmt.Errorf("retrieving from lotus: %s", err)
+	}
+
+	f, err := os.Open(ref.Path)
+	if err != nil {
+		return nil, fmt.Errorf("opening retrieved file: %s", err)
+	}
+	return &autodeleteFile{File: f}, nil
```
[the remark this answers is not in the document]
asutula: Nice

deals/rpc/rpc.go
```diff
@@ -44,7 +44,12 @@ func store(ctx context.Context, dealsModule *deals.Module, storeParams *StorePar
 			EpochPrice: dealConfig.GetEpochPrice(),
 		}
 	}
-	dcid, sr, err := dealsModule.Store(ctx, storeParams.GetAddress(), r, dealConfigs, storeParams.GetDuration(), false)
+	dcid, size, err := dealsModule.Import(ctx, r, false)
+	if err != nil {
+		ch <- storeResult{Err: err}
+		return
+	}
+	sr, err := dealsModule.Store(ctx, storeParams.GetAddress(), dcid, uint64(size), dealConfigs, storeParams.GetDuration())
```
asutula, line 52: Do we want to expose the separate `Import` and `Store` methods?

exe/server/main.go
asutula: Yea, I get sad whenever I see a CLI not using Cobra now 😢

ffs/Design.md
```diff
@@ -132,9 +132,8 @@ type ColdStorage interface {
 	// during execution.
 	Store(context.Context, cid.Cid, FilConfig) (FilInfo, []DealError, error)
 
-	// Retrieve retrieves the data using an account address,
-	// and store it in a CAR store.
-	Retrieve(context.Context, cid.Cid, car.Store, string) (cid.Cid, error)
+	// Fetch fetches the cid data in the underlying storage.
+	Fetch(context.Context, cid.Cid, car.Store, string) error
```
asutula, line 136: I like the "unfreeze" naming we use in the storage configs... I wonder if that would be appropriate here and be a cleaner separate meaning from "retrieve"

ffs/filcold/filcold.go
```diff
@@ -158,16 +147,16 @@ func (fc *FilCold) EnsureRenewals(ctx context.Context, c cid.Cid, inf ffs.FilInf
 	return inf, retErrors, nil
 }
 
-func (fc *FilCold) renewDeal(ctx context.Context, c cid.Cid, p ffs.FilStorage, activeMiners []string, fcfg ffs.FilConfig) (ffs.FilStorage, []ffs.DealError, error) {
+func (fc *FilCold) renewDeal(ctx context.Context, c cid.Cid, size uint64, p ffs.FilStorage, fcfg ffs.FilConfig) (ffs.FilStorage, []ffs.DealError, error) {
 	f := ffs.MinerSelectorFilter{
-		ExcludedMiners: activeMiners,
+		TrustedMiners: []string{p.Miner},
```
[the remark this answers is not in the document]
asutula: This makes sense. So nice that trusted miners are prioritized and used in a best effort way and the renewal can still happen with other miners if necessary 👍

ffs/filcold/filcold.go
```diff
@@ -177,21 +166,16 @@ func (fc *FilCold) renewDeal(ctx context.Context, c cid.Cid, p ffs.FilStorage, a
...
-	r := ipldToFileTransform(ctx, fc.dag, c)
-
+func (fc *FilCold) makeDeals(ctx context.Context, c cid.Cid, size uint64, cfgs []deals.StorageDealConfig, fcfg ffs.FilConfig) ([]ffs.FilStorage, []ffs.DealError, error) {
 	for _, cfg := range cfgs {
 		fc.l.Log(ctx, c, "Proposing deal to miner %s with %d fil per epoch...", cfg.Miner, cfg.EpochPrice)
 	}
 
 	var sres []deals.StoreResult
-	dataCid, sres, err := fc.dm.Store(ctx, fcfg.Addr, r, cfgs, uint64(fcfg.DealDuration), true)
+	sres, err := fc.dm.Store(ctx, fcfg.Addr, c, size, cfgs, uint64(fcfg.DealDuration))
```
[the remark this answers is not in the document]
asutula: I wonder if this means we can more easily save the size in some metadata about cold storage. When trying to display data size for the UI work, it previously seems impossible to know if hot storage was disabled.

ffs/integrationtest/integration_test.go
```diff
@@ -1084,11 +1059,7 @@ func TestSendFil(t *testing.T) {
 // Better than no test is some test, so this tests that the repair logic gets triggered
 // and the related Job ran successfully.
 func TestRepair(t *testing.T) {
-	// ToDo: Flaky on CI, runs ok locally.
-	// See if can be tuned to run better on CI.
-	t.SkipNow()
```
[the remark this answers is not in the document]
asutula: 🎆

ffs/filcold/filcold.go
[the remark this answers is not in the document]
asutula: It appears the answer is yes. I see the new size property in the model struct

deals/deals.go
sanderpick: AFAICT, I think it would be pinned via `Store`, this just makes it available

ffs/filcold/filcold.go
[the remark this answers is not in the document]
jsign: Correct.

deals/deals.go
jsign: @asutula , @sanderpick , yes. I'll extend a bit the explanation just in case.

`deals.Import()` import data to the Lotus node by sending it from Powergate. This make Lotus have the available data in a further call to `deals.Store`.

Now, something important. The FFS world *never* calls `deals.Import`, since in FFS the Lotus client is directly connected to IPFS. So it doesn't need to call `deals.Import` since the Lotus node has the data _already imported_ in IPFS.

The `deals.Import()` API is mostly useful for other use cases that uses the Deal Module which doesn't rely on a Lotus node connected to IPFS. In that case, the client should provide de deal data before making the deal.

Makes sense?

deals/deals.go
jsign: Correct.

deals/rpc/rpc.go
jsign: Uhm, not sure but it's a possibility if useful for some use case.

ffs/Design.md
jsign: Ha!, yeah the _unfreeze_ name is pretty cool, mostly regarding is understood in a "Cold"Storage scope. I'll give it a thought and might change name in the future, I'll let that suggestion sink a bit more :P

deals/deals.go
asutula: Yep, thank you

PR #415, opened 2020-05-25
Integration Fixes

asutula:
Just some small fixes I found while writing integration tests for the JS client.


ffs/rpc/rpc.go
```diff
@@ -638,3 +598,44 @@ func toRPCDealErrors(des []ffs.DealError) []*DealError {
 	}
 	return ret
 }
+
+func fromRPCHotConfig(config *HotConfig) ffs.HotConfig {
```
asutula, line 602: These two functions provide safe conversion from the nested pointer types in gRPC requests to our native Go types. Checks for `nil` objects where appropriate

ffs/types.go
```diff
@@ -226,6 +226,9 @@ type HotConfig struct {
 
 // Validate validates a HotConfig.
 func (hc HotConfig) Validate() error {
+	if !hc.Enabled {
+		return nil
+	}
```
asutula, line 231: Cleaning up and making the validators a little more smart. If something isn't enabled, then you can easily say it is valid. Same below.

scripts/gen-js-protos.sh
```diff
@@ -19,7 +19,7 @@ mkdir -p $PTOTOC_GEN_OUT_DIR
 
 printf '{"name":"%s", "version":"%s", "files":["%s"]}' $LIB_NAME $LIB_VERSION $LIB_SRC_DIR > $OUT_PATH/package.json
 
-(cd $OUT_PATH && npm install --save-dev ts-protoc-gen)
+(cd $OUT_PATH && npm install --save-dev ts-protoc-gen @types/google-protobuf)
```
asutula, line 22: Needed this dependency to make the JS library more plug n play

PR #418, opened 2020-05-27
change binary names and building

jsign:
Changes:
- Rename `exe` folder to `cmd` since it's more idiomatic Go naming.
- `make` defaults to `make build`
- `make build` will do `go install ./...` instead of local binaries.
- Binaries folders names changed to intended names (pow, powd, powbench).
- Updated readme.
- Left individual _installs_ if faster concrete builds might be desired.


jsign:
> That was fast! LGTM. I think `cmd/pow/cmd` could be removed, but no big deal.

Yeah, that felt weird. To be changed eventually!

asutula:
We can change that `cmd` subdiretory to be called whatever, but it's nice to only have one file `main.go` in the `main` package and have it import from `cmd` sub packages.

PR #423, opened 2020-05-28
Refactors

### LINKED ISSUE #394: ffs: switch val 0 of JobStatus to Unknown
It would make job status interpretation much safer regarding default values.

### LINKED ISSUE #395: Encapsulate internal dependencies
There're some things in the dependency graph which are internal components and shouldn't be really exposed in the constructor.

jsign:
This is a PR that doesn't add features but does some necessary refactors in multiple places. Mostly, to keep code healthy, lower complexity (mostly regarding dep graph creation), and understandability.

Not really expecting to be a code-review since a lot of code movement, deletion, renaming, simplification, etc was done. I think maybe seeing the end result directly (without diffs) might give a better appreciation of what happened.

A subset of changes:
- `powd/main.go` is cleaner
- `api/server/server.go` now create fewer dependencies, since internal components create their own unshared ones. So the dependency graph is much more encapsulated. Still more things to possible improvements, but this was a big chunk.
- Internal components of `scheduler` are now internal, so even exported components aren't visible out of the `ffs` package scope. (as they should, nobody outside ffs should use them).
- Left interface definitions only for things that now we _could_ expect to have more than one implementation. e.g: the `Scheduler` interface is redundant, we don't expect different scheduler implementations. e.g: `JobStore` we wouldn't expect to have more than one implementation (at maxim it could receive a different datastore implementation). On the other side, _HotStorage_ and _ColdStorage_ are still interfaces since they like interesting fields to explore other impls.
- Closes #394 
- Closes #395

Since a lot of datastore namespaces (which become nested in internal components) were done, I did a little experiment of running a usual use case of FFS, and then dumping the keyspace on close to check that all things look reasonable:
```
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700916607438100
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700916610985100
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700916617184300
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700916620825300
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700916720712400
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700916723271900
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700916726414400
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700916731446400
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700917737310600
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700947709706000
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700977680164200
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700977684519200
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700977687906500
powergate_1  | /ffs/cidlogger/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq/1590700977697244000
powergate_1  | /ffs/manager/api/2c06f9f6-129c-43b0-8e7b-f7e8d8842c4c/istore/cidconfig/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq
powergate_1  | /ffs/manager/api/2c06f9f6-129c-43b0-8e7b-f7e8d8842c4c/istore/defaultconfig
powergate_1  | /ffs/manager/auth/83f558b3-d768-4c68-afa8-1502ca1ad0ea
powergate_1  | /ffs/scheduler/astore/7ffc5a49-36e7-49f1-8b5b-7f76933aada3
powergate_1  | /ffs/scheduler/cistore/cistore/QmQBPZBN3S8wn59MZBsZanJAMjJ8ViMZjNTi4bQaBQwcTq
powergate_1  | /ffs/scheduler/jstore/7ffc5a49-36e7-49f1-8b5b-7f76933aada3
powergate_1  | /index/miner/chainstore/data/q�� �M@�c�X�Ngߵ
                                                           �a�V��N�@VIC�
powergate_1  | /index/miner/chainstore/data/q�� �o���m%�#���6�.��9�]D��b�����
powergate_1  | /index/miner/chainstore/id/1
powergate_1  | /index/miner/chainstore/id/2
powergate_1  | /index/miner/index/meta
powergate_1  | /index/slashing/chainstore/data/q�� �M@�c�X�Ngߵ
                                                              �a�V��N�@VIC�
powergate_1  | /index/slashing/chainstore/data/q�� �o���m%�#���6�.��9�]D��b�����
powergate_1  | /index/slashing/chainstore/id/1
powergate_1  | /index/slashing/chainstore/id/2
```
(The strange symbols are some binary keys).


api/server/server.go
```diff
@@ -79,9 +76,6 @@ type Server struct {
 	hm   *health.Module
 
 	ffsManager *manager.Manager
-	js         *jstore.Store
-	cis        *cistore.Store
-	as         *astore.Store
```
jsign, line 84: These are internal components of the scheduler. They should be encapsulated.

api/server/server.go
```diff
@@ -184,16 +178,13 @@ func NewServer(conf Config) (*Server, error) {
 		return nil, fmt.Errorf("creating ipfs client: %s", err)
 	}
 
-	lchain := lotuschain.New(c)
+	chain := filchain.New(c)
```
jsign, line 181: Rename to be _lotus_-nameless.

api/server/server.go
```diff
@@ -184,16 +178,13 @@ func NewServer(conf Config) (*Server, error) {
...
 
 	l := cidlogger.New(txndstr.Wrap(ds, "ffs/scheduler/logger"))
-	cs := filcold.New(ms, dm, ipfs, lchain, l)
+	cs := filcold.New(ms, dm, ipfs, chain, l)
 	hs := coreipfs.New(ipfs, l)
-	js := jstore.New(txndstr.Wrap(ds, "ffs/scheduler/jstore"))
-	as := astore.New(txndstr.Wrap(ds, "ffs/scheduler/astore"))
-	cis := cistore.New(txndstr.Wrap(ds, "ffs/scheduler/cistore"))
-	sched := scheduler.New(js, as, cis, l, hs, cs)
+	sched := scheduler.New(txndstr.Wrap(ds, "ffs/scheduler"), l, hs, cs)
```
jsign, line 187: All the deleted stores are internal things of the `scheduler`, which now is responsible for creating them, and closing them. The `server.go` only knows about `scheduler`, don't have to worry about internal things.

cmd/powd/main.go
```diff
@@ -29,150 +29,177 @@ var (
 )
```
jsign, line 30: So a lot of movement here to have a cleaner main.

ffs/api/api_actions.go
```diff
@@ -0,0 +1,150 @@
+package api
```
jsign, line 1: Splitted `api.go` in some more logical files, so we can find things easily and don't get so long files.

ffs/api/types.go
```diff
@@ -12,17 +12,6 @@ var (
 	ErrNotFound = errors.New("stored item not found")
 )
 
-// InstanceStore is a repository for all state of a Api.
```
jsign, line 15: We won't expect to have more than one impl. Unnecessary.

ffs/api/types.go
```diff
@@ -57,45 +46,9 @@ type NewAddressConfig struct {
...
-	}
-}
-
-// WithHistory indicates that prior log history should be sent
-// to the channel before getting realtime logs.
-func WithHistory(enabled bool) GetLogsOption {
-	return func(c *GetLogsConfig) {
-		c.history = enabled
-	}
-}
```
jsign, line 101: Moved to the existing `options.go` file.

ffs/interfaces.go
```diff
@@ -20,53 +19,6 @@ type WalletManager interface {
...
 
-var (
-	// ErrHotStorageDisabled returned when trying to fetch a Cid when disabled on Hot Storage.
-	// To retrieve the data, is necessary to call unfreeze by enabling the Enabled flag in
-	// the Hot Storage for that Cid.
-	ErrHotStorageDisabled = errors.New("cid disabled in hot storage")
-)
-
-// Scheduler enforces a CidConfig orchestrating Hot and Cold storages.
-type Scheduler interface {
```
jsign, line 31: There will be only one `Scheduler` implementation.

ffs/manager/manager.go
```diff
@@ -76,8 +74,7 @@ func (m *Manager) Create(ctx context.Context) (ffs.APIID, string, error) {
 
 	log.Info("creating instance")
 	iid := ffs.NewAPIID()
-	is := istore.New(iid, namespace.Wrap(m.ds, istoreNamespace))
-	fapi, err := api.New(ctx, iid, is, m.sched, m.wm, defCidConfig)
+	fapi, err := api.New(ctx, namespace.Wrap(m.ds, datastore.NewKey("ffs/manager/api")), iid, m.sched, m.wm, defCidConfig)
```
jsign: Each `api` is responsible of creating their own `InstanceStore`. The manager just creates/loads `api`s, not also their dependencies.

ffs/scheduler/types.go
```diff
@@ -1,59 +0,0 @@
...
-
-	"github.com/ipfs/go-cid"
-	"github.com/textileio/powergate/ffs"
-)
-
-var (
-	// ErrNotFound is returned when an item isn't found on a Store.
-	ErrNotFound = errors.New("item not found")
-)
-
```
jsign, line 15: We don't expect more than one impl of this stores.
At most, we can give the stores different implementations of `datastore.Datastore` (if want to save in other databases, etc), that's no problem.

ffs/types.go
```diff
@@ -57,7 +57,7 @@ type JobStatus int
 
 const (
 	// Queued indicates the Job is queued in the Scheduler.
-	Queued JobStatus = iota
+	Queued JobStatus = iota + 1
```
jsign, line 60: Thought that starting enums at 1 would be safer. Just to avoid that the default value is `Queued`, and some client (or some internal code) that fogets checking errors interprets a "valid" status which shouldn't be considered.

This idea might apply to many places (and repos!)... so kind of throwing the idea to see feedback.

ffs/integrationtest/integration_test.go
```diff
@@ -768,6 +764,8 @@ Loop:
 }
 
 func TestRenewWithDecreasedRepFactor(t *testing.T) {
+	// Too flaky for CI.
+	t.SkipNow()
```
jsign, line 768: Too flaky to my taste. I prefer not bumping the CI machine price just to make this smooth.

ffs/scheduler/scheduler.go
```diff
@@ -2,13 +2,19 @@ package scheduler
...
 	"time"
 
 	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/textileio/powergate/ffs"
+	"github.com/textileio/powergate/ffs/scheduler/internal/astore"
+	"github.com/textileio/powergate/ffs/scheduler/internal/cistore"
+	"github.com/textileio/powergate/ffs/scheduler/internal/jstore"
+	txndstr "github.com/textileio/powergate/txndstransform"
```
asutula, line 17: Love the internal packages

ffs/types.go
asutula: Seems like a pretty good idea. Protobuf style recommends always to have the 0 value as the "unspecified" value:

> The zero value enum should have the suffix UNSPECIFIED

Maybe that could work well in Go as well.

asutula: That's from https://developers.google.com/protocol-buffers/docs/style#enums

asutula: Looks like we need to update the rpc layer to be safe mapping to the proto version of the enum... we're just assuming the int values are equal, we need to check explicitly. PR coming up

PR #425, opened 2020-05-29
Fix JobStatus handling

asutula:
- Add a default "unspecified" value for job status
- Applying some best practices for proto enums
- Don't assume raw values of Go and proto job state are equal, check with `switch`


ffs/rpc/rpc.proto
```diff
@@ -105,11 +105,12 @@ message InstanceInfo {
...
-   Executing = 1;
-   Failed = 2;
-   Canceled = 3;
-   Success = 4;
+   UNSPECIFIED = 0;
+   QUEUED = 1;
+   EXECUTING = 2;
+   FAILED = 3;
+   CANCELED = 4;
+   SUCCESS = 5;
```
asutula, line 113: These changes come from proto style guide. Will apply everywhere if we like it.

ffs/types.go
```diff
@@ -56,8 +56,10 @@ func (i APIID) String() string {
 type JobStatus int
 
 const (
+	// Unspecified indicates a default or empty value
+	Unspecified JobStatus = iota
```
asutula, line 60: This feels right to me, but let me know if it doesn't to you. The idea is this would be the default value. It's never used otherwise.

asutula:
No idea what's up with this CI error: https://github.com/textileio/powergate/pull/425/checks?check_run_id=718939763#step:3:5

Let me know if you have any ideas.

ffs/rpc/rpc.proto
jsign: Proto style guide sounds like a good argument to me. 👍

ffs/types.go
jsign: I thought about this when I moved to starting at 1. I preferred to avoid giving name to something that doesn't have a meaning about "Job Status". But no strong opinion since I had my doubts at that moment too. LGTM.

jsign:
> No idea what's up with this CI error: https://github.com/textileio/powergate/pull/425/checks?check_run_id=718939763#step:3:5
> 
> Let me know if you have any ideas.

Yeah, weird. chmoded `/tmp/powergate` in the CI and re-run ok.

PR #428, opened 2020-05-29
Cold config max price option

asutula:
Pretty self explanatory. Just like country codes and miner lists, this allows you to specify a max price you're willing to pay for a deal. You can do this in your ffs default storage config on or a per-config basis. Checks some boxes for UI support and seems like a good feature.


ffs/integrationtest/integration_test.go
```diff
@@ -488,6 +488,38 @@ func TestFilecoinCountryFilter(t *testing.T) {
 	require.Equal(t, p.Miner, "t01001")
 }
 
+func TestFilecoinMaxPriceFilter(t *testing.T) {
```
asutula, line 491: This just verifies that we fail to find miners if our max price is too low, then verifies we find the miner if our max price is high enough.

ffs/minerselector/reptop/reptop.go
```diff
@@ -29,24 +29,36 @@ func New(rm *reputation.Module, ai *askRunner.Runner) *RepTop {
 // GetMiners returns n miners using the configured Reputation Module and
 // Ask Index.
 func (rt *RepTop) GetMiners(n int, f ffs.MinerSelectorFilter) ([]ffs.MinerProposal, error) {
-	ms, err := rt.rm.QueryMiners(n, f.ExcludedMiners, f.CountryCodes, f.TrustedMiners)
+	if n < 1 {
+		return nil, fmt.Errorf("the number of miners should be greater than zero")
+	}
+	ms, err := rt.rm.QueryMiners(f.ExcludedMiners, f.CountryCodes, f.TrustedMiners)
```
asutula, line 35: `rt.rm.QueryMiners` now returns a list of _all_ miners that match excluded, countries, and trusted. We don't want to limit the length of that list because we can't know if we'll use any miner in the list until we inspect their asks and compare to max price.

ffs/minerselector/reptop/reptop.go
```diff
@@ -29,24 +29,36 @@ func New(rm *reputation.Module, ai *askRunner.Runner) *RepTop {
...
-	res := make([]ffs.MinerProposal, 0, len(ms))
+	res := make([]ffs.MinerProposal, 0, n)
 	for _, m := range ms {
 		sa, ok := aidx.Storage[m.Addr]
 		if !ok {
 			continue
 		}
+		if f.MaxPrice > 0 && sa.Price > f.MaxPrice {
+			continue
+		}
```
asutula, line 51: If max price isn't the default (0, which means no max price), compare the ask price to the max price and ignore the ask if the price is too high.

ffs/minerselector/reptop/reptop.go
```diff
@@ -29,24 +29,36 @@ func New(rm *reputation.Module, ai *askRunner.Runner) *RepTop {
...
 		res = append(res, ffs.MinerProposal{
 			Addr:       sa.Miner,
 			EpochPrice: sa.Price,
 		})
+		if len(res) == n {
+			break
+		}
+	}
+	if len(res) < n {
+		return nil, fmt.Errorf("not enough miner asks that satisfy the max price constraint")
```
asutula, line 61: Error if we can't find enough asks below our max price

ffs/types.go
```diff
@@ -182,6 +182,19 @@ func (c CidConfig) WithColdFilRenew(enabled bool, threshold int) CidConfig {
...
+func (c CidConfig) WithColdMaxPrice(maxPrice uint64) CidConfig {
+	c.Cold.Filecoin.MaxPrice = maxPrice
+	return c
+}
+
+// WithColdAddr specifies the wallet address that should be used for transactions
+func (c CidConfig) WithColdAddr(addr string) CidConfig {
+	c.Cold.Filecoin.Addr = addr
+	return c
+}
```
asutula, line 196: Just notices this was missing, so added it. There may be others missing.

reputation/reputation.go
```diff
@@ -83,17 +83,10 @@ func (rm *Module) AddSource(id string, maddr ma.Multiaddr) error {
 
 // QueryMiners makes a filtered query on the scored-sorted miner list.
 // Empty filter slices represent no-filters applied.
-func (rm *Module) QueryMiners(n int, excludedMiners []string, countryCodes []string, trustedMiners []string) ([]MinerScore, error) {
-	if n < 1 {
-		return nil, fmt.Errorf("the number of miners should be greater than zero")
-	}
-
+func (rm *Module) QueryMiners(excludedMiners []string, countryCodes []string, trustedMiners []string) ([]MinerScore, error) {
```
asutula, line 86: Like I said, this no longer takes a number of miners to find. It just returns all miners that match the provided filters.

ffs/minerselector/reptop/reptop.go
jsign: I see. I think eventually we should give more thought about miner-selectors... since returning the complete list sounds like might give some problems if there're a lot of miners. Definitely not something with a good return on time investment now, but to food for thought.

ffs/minerselector/reptop/reptop.go
jsign: Maybe the message could be more generic, since the Reputation module could already return less than `n` results.
Even if the error sentence is true, wasn't really because of _price contraint_ but maybe countries, or other filters.

ffs/minerselector/reptop/reptop.go
asutula: yea, agreed.

ffs/minerselector/reptop/reptop.go
jsign: Ah, lost that with the diff thing. LGTM.

Issue #386, opened 2020-05-08
ffs/manager: add auth token & unpublished api


jsign:
Would be good to enable an _admin token_ to call _admin_ related APIs such as _Create_ or other kind of APIs. 

We're living on the premise of not being exposed to the public, but that shouldn't be blocked by just adding a simple auth here. Might be wise to separate the admin endpoint to an unpublished one for extra security.

andrewxhill:
this one is starting to feel important in light of the up coming competition. @asutula @jsign not sure how hard it is though...

jsign:
BTW, I can jump on this soon. Apart from creating an FFS instance, another candidate?
We can tune that set of APIs later anyway, just getting a bit ahead.

jsign:
@andrewxhill , @asutula, maybe feels if an admin-token is specified, all APIs should be authenticated:
Maybe a rule like: if the API doesn't expect an FFS token, then it should receive the admin-token?

Even in read-only APIs, like reading indexing data, it might be a good idea to protect it with a token to avoid someone just simply flooding with requests?

andrewxhill:
well… perhaps there is another way to dice it (3rd key?). because i can see a developer using ffs apis in a scope that doesn’t have access to the admin but wanting a straight-forward way to read some of the indices

jsign:
@andrewxhill , so maybe:
- FFS token (for things that already is used)
- Admin token: for some APIs
- Developer token: for some other APIs

That would mean all APIs would be covered with some token.

asutula:
Sorry need to read this more closely when I have time, but quick thought: there should be admin apis and then all other apis should be covered by what we now call the FFS token. My thinking is that in a future version of the Powergate API, the default mode of operation will be token-authenticated, scoped addresses and data etc... the word "FFS" will disappear from the API.

asutula:
So basically, you're either an admin or you're a user with token access. That token gives you access to all non-admin apis, and those apis are scoped to your FFS instance where appropriate, but that is an internal implementation detail.

jsign:
Yeah, that's another possibility. That would force to always create an FFS instance to interact with the API, compared to using a _developer token_ which doesn't need that. The only way I could think that is a problem is if the manager has the auto-funding feature enabled, and you just want to give a developer access to the APIs (and not really FFS).

[...]

PR #429, opened 2020-05-29
ffs/manager: configurable persisted default CidConfig & better integrationtest setup 

### LINKED ISSUE #426: ffs/manager: make default config changable
Currently, the default CidConfig for the manager, which is used as default config for newly created FFS instances, is fixed and unchangable.

We should do two things:
- Create Get/Set API
- Persist the current config.

jsign:
In this PR three things are addressed:
- Closes #426 
- Fixes a bug regarding Replace on HotStorage not updating the cached pinset.
- Reestructure integrationtest helpers better.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ build-powbench:
 .PHONY: build-powbench
 
 test:
-	go test -short -p 4 -race -timeout 30m ./... 
+	go test -short -parallel 4 -race -timeout 30m ./... 
 .PHONY: test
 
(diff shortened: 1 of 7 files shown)


jsign:
> Looks good. Any thoughts or plan about how to expose those apis to grpc and the client/cli? Will be easy to add, but more a question of do we want to secure it with an admin auth token?

Yes, I created #386 some weeks ago to have that on the radar.

PR #430, opened 2020-05-29
Integrate Buf for protobuf lint

asutula:
Friday afternoon project integrating this great library called [Buf](https://buf.build/docs/introduction).

- Adds `Makefile.buf` for running Buf-related things
- Adds a GitHub action that runs Buf lint checks for PRs
- Corrects all the lint problems Buf found

You can run these checks locally with `make -f Makefile.buf local`.

Also supports breaking change detection. I've disabled that for now, but we way want to enable it once we're in a production release mode.

I'm leaning more and more towards doing Protobuf things the strict Protobuf way. I think it will prevent lots of unforeseen headaches in the future. I'm holding out on one lint rule that requires you version your Protobuf package names and directory structure because it really cuts against the way we have things organized currently. But I'm going to think about that one some more and maybe make the change soon.


asutula:
This is 99.9% changes for Buf lint fixes. The interesting stuff is just the new makefile and GitHub action config

index/slashing/rpc/rpc.proto
```diff
@@ -4,7 +4,7 @@ package index.slashing.rpc;
 option go_package = "index/slashing/rpc";
 
 message Index {
-    string tipSetKey = 1;
+    string tip_set_key = 1;
```
jsign, line 7: This might be an exception to the rule: should be `tipsetkey`.
Or at least ideally should be an exception, not sure if that's hard to make an exception in the linter.
If it's a headache, I would leave it as is... np.

asutula: That was my interpretation, but if you want it all one word, that is fine with the linter.

asutula: It will be `Tipsetkey` in Go. making the change now.

asutula:
@sanderpick I added you to review just because I want to keep you in the loop and you should see how this works. Gonna merge, but let me know what you think of the whole thing.

PR #434, opened 2020-06-02
ffs: resume scheduler abruptly closed state

### LINKED ISSUE #221: Support arbitrary crashed states
Despite this should be the case for any repo, there's a particularly important case in this one. 

The scheduler is executing long-running processes in which arbitrary crashes might leave the state in some non-final states. When the process is recovered, it should expect jobs in _In Process_ states that aren't really handled but might indicate a previous crash. Considering the jobs are long-running, the probability of finding these situations in Powergate is orders of magnitude bigger than in other systems where most _important executions_ are in the order of milliseconds.

The above case should consider what impact that state of the system has regarding other external systems, such as the IPFS node or the Lotus node. Between the crash, and crash recovery external processes might have continued successfully (or not). 

So this involves not only recognizing the current state of Powergate, but external systems. After that, reconciling a reasonable execution plan to continue on-going work or fail if that can't be possible.

The good part of this system is that it was thought from the ground up to be _auto-repairable_, so in the worst-case scenario this _arbitrary crash recovery_ sounds like it should be healed (in the worst case) by a forced repair logic.

jsign:
Closes #221, which the issue description gives a general notion of the problem that is trying to be solved.

At the implementation level, this was solved as follows. Whenever a new deals are being created, they're are now saved in a temporal storage as "in progress deals". If all works correctly, they will be deleted of that storage as soon as the deals become active or fail.

If Powergate crashes or is forcibly closed, whenever is restarted again it will inspect which jobs "are" in _Executing_ status. For those, we should have in our temporal storage the _ProposalCids_ which were being waited when the abrupt-close happened.

As soon as the scheduler restart, it will reattach to inspect and wait for those Jobs first, let them finish (with _Success_ or _Failed_ as the usual case), and then continue with the other _Queued_ jobs that existed. So in short, it will recover first ongoing work that was left behind, and after all is "good" to continue, continue as usual.

This required some changes in the _ColdStorage_ interface, which now has the `Store(...)` API which fires the new deals that should be made, and a new API `WaitForDeals(...)` that waits for provided deals to reach a final status. Before, both things were done in `Store(...)`.  This change allows the scheduler to know which _ProposalCids_ were fired to Filecoin, save them in temporal storage, and then continue waiting for them. It also helps the _Scheduler_ to resume work since it has only to `WaitForDeals` on resuming, and not firing new deals.

---

I also renamed in multiple parts the _Duration_ concept, with _MinDuration_ since the client provides a _minimal duration_ of the deal which, when the deal really happens, gets automatically adjusted to a real duration.


ffs/filcold/filcold.go
```diff
@@ -158,56 +157,67 @@ func (fc *FilCold) renewDeal(ctx context.Context, c cid.Cid, size uint64, p ffs.
...
+		okDeals = append(okDeals, r.ProposalCid)
 	}
-	return proposals, errors, nil
+	return okDeals, failedDeals, nil
 }
 
-func (fc *FilCold) waitForDeals(ctx context.Context, c cid.Cid, storeResults []deals.StoreResult, duration int64) ([]ffs.FilStorage, []ffs.DealError, error) {
+// WaitForDeals blocks waiting for all the provided proposals to reach a final status.
+// It returns the ones that finished successfully and the other deal errors that happened on failed ones.
+func (fc *FilCold) WaitForDeals(ctx context.Context, c cid.Cid, proposals []cid.Cid) ([]ffs.FilStorage, []ffs.DealError, error) {
```
jsign, line 211: Since now `Store` doesn't wait for deals, some of the work `waitForDeals` was doing is more clearly done by `Store`. Now `WaitForDeals` has a clearer logic.

ffs/integrationtest/integration_test.go
```diff
@@ -420,7 +420,7 @@ func TestDurationConfig(t *testing.T) {
 	cinfo, err := fapi.Show(cid)
 	require.Nil(t, err)
 	p := cinfo.Cold.Filecoin.Proposals[0]
-	require.Equal(t, duration, p.Duration)
+	require.Greater(t, p.Duration, duration)
```
jsign, line 423: So, this is considering all the discussion today with the _duration_ in Lotus.
The real/confirmed Deal duration isn't what we proposed in the deal (since, its minDuration). So the confirmed duration of the deal should be greater than the proposed one (not equal).

api/server/server.go
```diff
@@ -109,7 +110,7 @@ type Config struct {
 func NewServer(conf Config) (*Server, error) {
 	var err error
 	var masterAddr address.Address
-	c, cls, err := lotus.New(conf.LotusAddress, conf.LotusAuthToken)
+	c, cls, err := lotus.New(conf.LotusAddress, conf.LotusAuthToken, lotusConnectionRetries)
```
jsign, line 113: Now we retry some times before considering the Lotus node to be down and abort Powergate start.
Lotus is generally slower to start and enable its API endpoint.

deals/types.go
```diff
@@ -31,6 +31,7 @@ type DealInfo struct {
 	Size     uint64
 
 	PricePerEpoch uint64
+	StartEpoch    uint64
```
jsign, line 34: `StartEpoch` is at which epoch the Deal will start charging the client.
`ActivationEpoch` is at which epoch the deal could be considered active for retrieval.

Both are different things (Activation <= Start).

ffs/filcold/filcold.go
```diff
@@ -219,22 +229,13 @@ func (fc *FilCold) waitForDeals(ctx context.Context, c cid.Cid, storeResults []d
...
-				}
-			}
-			if d == nil {
-				return nil, errors, fmt.Errorf("deal watcher return unasked proposal, this must never happen")
-			}
 			activeProposals[di.ProposalCid] = &ffs.FilStorage{
 				ProposalCid:     di.ProposalCid,
-				Duration:        duration,
-				Miner:           d.Config.Miner,
+				Duration:        int64(di.Duration),
```
jsign, line 234: Notice how `Duration` change its meaning.
Before we said that Duration was what the client specified in its deal proposal, that was wrong.
`Duration` should be (and is now) what the confirmed-active deal say it is.
This was an important fix since the "Start+Duration" consideration for deal renewals.

ffs/integrationtest/integration_test.go
```diff
@@ -774,7 +774,7 @@ func TestRenew(t *testing.T) {
 	ra := rand.New(rand.NewSource(22))
 	cid, _ := addRandomFile(t, ra, ipfsAPI)
 
-	renewThreshold := 50
+	renewThreshold := 12600
```
jsign, line 777: Magic number!, considering that Lotus does multiple calculations and buffering to calculate when the deals will end. This number isn't that magic, I could explain it but will be quite long and a bit pointless for Powergate scope.

ffs/integrationtest/integration_test.go
```diff
@@ -1242,6 +1242,40 @@ func TestLogHistory(t *testing.T) {
 	require.Greater(t, len(lgs), 3) // Ask to have more than 3 log messages.
 }
 
+func TestResumeScheduler(t *testing.T) {
```
jsign, line 1245: A new test to test the new feature of Scheduler properly "reattaching" to "lost" deals that were executing while Powergate was abruptly terminated.

It was quite tricky to do, because firing deals and closing and reopening the thing isn't really "abruptly closing". Doing a proper closing wouldn't leave things "orphaned" because even in Ctrl+C things are closed properly.

Figured out a clean way to "freeze" state in an arbitrary point, see next comment.

ffs/integrationtest/integration_test.go
```diff
@@ -1242,6 +1242,40 @@ func TestLogHistory(t *testing.T) {
...
+	fapi, err := manager.GetByAuthToken(auth)
+	require.NoError(t, err)
+
+	r := rand.New(rand.NewSource(22))
+	c, _ := addRandomFile(t, r, ipfs)
+	jid, err := fapi.PushConfig(c)
+	require.Nil(t, err)
+
+	time.Sleep(time.Second * 3)
+	ds2, err := ds.Clone()
```
jsign, line 1264: I created a `Clone()` new API for the in memory datastore. 
So I completely clone the datastore in an arbitrary point of execution, and then use that cloned database in a newly spinned up FFS. 

That is the only way to have a ""corrupted/interrupted/etc"" state of the application. (Since doing any kind of closing would leave things in an expected way).

Hope makes sense to you!

ffs/scheduler/internal/jstore/jstore.go
```diff
@@ -36,11 +40,15 @@ type watcher struct {
...
-func New(ds datastore.Datastore) *Store {
-	return &Store{
-		ds:         ds,
-		inProgress: make(map[cid.Cid]struct{}),
+func New(ds datastore.Datastore) (*Store, error) {
+	s := &Store{
+		ds:            ds,
+		executingCids: make(map[cid.Cid]ffs.JobID),
 	}
+	if err := s.loadExecutingJobs(); err != nil {
```
jsign, line 48: When the `JobStore` of the `Scheduler` is created, it recognizes which Jobs appear to be in "Executing" status... which is necessary to recover them and continue them.

ffs/scheduler/scheduler.go
```diff
@@ -211,6 +215,10 @@ func (s *Scheduler) Close() error {
 
 func (s *Scheduler) run() {
 	defer close(s.finished)
+	if err := s.resumeStartedDeals(); err != nil {
+		log.Errorf("resuming started deals: %s", err)
+		return
+	}
 	for {
```
jsign, line 222: So before start listening to the new pushed `Jobs` to execute, we inspect if we have to resume deals that where orphaned from an interrupted close of Powergate. After we continued that and all clean, continue with the usual.

tests/txmapds.go
```diff
@@ -48,6 +49,29 @@ func (d *TxMapDatastore) Query(q query.Query) (query.Results, error) {
 	return d.MapDatastore.Query(q)
 }
 
+// Clone returns a cloned datastore.
+func (d *TxMapDatastore) Clone() (*TxMapDatastore, error) {
```
jsign, line 53: Clone datastore for what was mentioned before.

asutula:
👀

ffs/integrationtest/integration_test.go
sanderpick: Magic numbers = more value for powergate :)

ffs/integrationtest/integration_test.go
sanderpick: Nice! Could there have been a way to leverage `exec` to run the program (after compiling in the test), and then use `Kill` from the command process (`cmd.Process.Kill()`). That might simulate Ctr+C a bit more directly. Just a side thought.

deals/types.go
jsign: Actually, many others were missing. Updated.

ffs/integrationtest/integration_test.go
sanderpick: 👍

deals/rpc/rpc.go
```diff
@@ -147,7 +147,7 @@ func (s *RPC) Watch(req *WatchRequest, srv RPCService_WatchServer) error {
 			PieceCid:      update.PieceCID.Bytes(),
 			Size:          update.Size,
 			PricePerEpoch: update.PricePerEpoch,
-			MinDuration:   update.Duration,
+			Duration:      update.Duration,
```
asutula: Ah right because DealInfo is the actual state of the deal, not the desired config

PR #437, opened 2020-06-03
paych module + client + cli

asutula:
I wanted to get some version of the pay channel api in so we can see how it feels in relation to the UI work. I decided to keep it as minimal as possible because there are still some unknowns and a nasty lotus bug that prevents it from working completely. Main thoughts are:

- This will give the UI work _something_ to get started with, and actually meets the requirements for the existing UI mockups.
- Pay channels are automatically created and managed as part of the retrieval deal process
- To really interact with pay channels in a more manual way will require more api implementation for creating, validating, and submitting "vouchers", but this is beyond the scope of the UI work for now.
- It kinda feels like pay channels are not really meant to be used manually or beyond retrieval deals. Or if they are, there is some UX work to be done there. Need to dig in more to this issue in the future.
- This work is based on my best understanding of pay channels and how they work. I'm sure I made some incorrect decisions, so we can fix things as we learn more.


ffs/api/api_paych.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/ipfs/go-cid"
+	"github.com/textileio/powergate/ffs"
+)
+
+// ListPayChannels lists all payment channels associated with this FFS
+func (i *API) ListPayChannels(ctx context.Context) ([]ffs.PaychInfo, error) {
```
asutula, line 12: The ffs pay channel api is mostly verifying that the addresses involved in payment channel operations are managed by the ffs instance, then calling down into the `paych` module.

ffs/interfaces.go
```diff
@@ -16,7 +16,31 @@ type WalletManager interface {
...
+
+// WithCreateVoucherLane allows you to specify a non-0 lane
+func WithCreateVoucherLane(lane uint64) CreateVoucherOption {
+	return func(conf *createVoucherConfig) {
+		conf.lane = lane
+	}
+}
+
+// PaychManager provides access to payment channels
+type PaychManager interface {
```
asutula, line 37: We've used a couple different designs with our modules and ffs, but what I chose to do here is just have ffs define the interface and types for `PaychManager`. The implementation's main goal is to support ffs. In fact, I didn't even add rpc, client or cli layers to the importation since that is all accessed through ffs. If we decide sometime there is a need to expose the raw `paych` module as an api, we can, but it would just be extra work for now.

paych/lotus/module.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// List lists all payment channels involving the specified addresses
+func (m *Module) List(ctx context.Context, addrs ...string) ([]ffs.PaychInfo, error) {
+	filter := make(map[string]struct{}, len(addrs))
+	for _, addr := range addrs {
+		filter[addr] = struct{}{}
+	}
+
+	allAddrs, err := m.api.PaychList(ctx)
```
asutula, line 36: Get the payment channel addresses.

paych/lotus/module.go
```diff
@@ -0,0 +1,178 @@
...
+		go func(ch <-chan statusResult) {
+			res := <-ch
+			resultsCh <- res
+		}(c)
+	}
+
+	results := make([]statusResult, len(chans))
+	for i := 0; i < len(chans); i++ {
+		results[i] = <-resultsCh
+	}
```
asutula, line 57: Above, use the wild ways of go async code to concurrently get the pay channel status info for each pay channel address

paych/lotus/module.go
```diff
@@ -0,0 +1,178 @@
...
+			final = append(final, info)
+		}
+	}
+
+	return final, nil
+}
+
+// Create creates a new payment channel
+func (m *Module) Create(ctx context.Context, from string, to string, amount uint64) (ffs.PaychInfo, cid.Cid, error) {
+	return ffs.PaychInfo{}, cid.Undef, fmt.Errorf("unimplemeted for now, blocked by lotus issue #1611")
```
asutula, line 96: For now just returning this error because lotus will crash until that issue is resolved.

paych/lotus/module.go
```diff
@@ -0,0 +1,178 @@
...
+	// if err != nil {
+	// 	return ffs.PaychInfo{}, cid.Undef, err
+	// }
+	// // ToDo: verify these addresses and direction make sense
+	// res := ffs.PaychInfo{
+	// 	CtlAddr:   from,
+	// 	Addr:      info.Channel.String(),
+	// 	Direction: ffs.PaychDirOutbound,
+	// }
+	// return res, info.ChannelMessage, nil
```
asutula, line 116: Interestingly, it seems that the `info.Channel` address is the pay channel address. I _think_ this is NOT the address you set up the pay channel "to". Strangely, it seems that after you initially create the pay channel, the "to" address is obscured away in the pay channel from then on. Will lead to some weird UX in a cli or UI. Takeaway is that pay channels might be more of an internal tool than the UI work wants them to be.

ffs/interfaces.go
```diff
@@ -16,7 +16,31 @@ type WalletManager interface {
...
+
+// CreateVoucherOption modifies a createVoucherCOnfig
+type CreateVoucherOption func(*createVoucherConfig)
+
+// WithCreateVoucherLane allows you to specify a non-0 lane
+func WithCreateVoucherLane(lane uint64) CreateVoucherOption {
+	return func(conf *createVoucherConfig) {
+		conf.lane = lane
+	}
+}
```
jsign, line 34: Can this be moved to `ffs/api/options.go`?

paych/lotus/module.go
jsign: Yeah, I guess will be discovered with time.

paych/lotus/module.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	var best *paych.SignedVoucher
+	for _, v := range vouchers {
+		spendable, err := m.api.PaychVoucherCheckSpendable(ctx, chAddr, v, nil, nil)
+		if err != nil {
+			return err
+		}
+		if spendable {
+			if best == nil || v.Amount.GreaterThan(best.Amount) {
```
jsign, line 137: Nit: may join both ifs in a single one?

ffs/api/api_paych.go
```diff
@@ -0,0 +1,50 @@
...
+		if i.isManagedAddress(channel.CtlAddr) {
+			managed = true
+			break
+		}
+	}
+	if !managed {
+		return fmt.Errorf("paych %v is not managed by ffs instance", addr)
+	}
+	return i.pm.Redeem(ctx, addr)
+}
```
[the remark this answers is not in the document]
asutula: Good catch, big mistake there. Fixed. Now comparing the `addr` param to the managed channel `Addr` (the address of the pay channel)

ffs/interfaces.go
asutula: I actually removed `CreateVoucher()` from the interface, so these are no longer needed. May have to add it back in some day.

paych/lotus/module.go
asutula: Yep

asutula:
Thanks for the review @jsign. Yea, this stuff still feels a bit unclear. I plan on re-visiting this code after that lotus bug is fixed and I can create some pay channels and learn more about them. I also thing our requirements will drive a lot of the work (or not) in this area.

ffs/api/api_paych.go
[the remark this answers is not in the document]
asutula: Yea, `CtlAddr`... I can't give you a definition, but from observation, I know that that when I have a pay channel setup to pay a miner for a retrieval, the `CtlAddr` is my wallet address.

Issue #421, opened 2020-05-28
index/miner: add number of deals


jsign:
Add number of deals of miners to the index.

jsign:
Closed by #438

Issue #393, opened 2020-05-14
ffs: consider sector size in miner selection


jsign:
It will be important to consider _sector sizes_ when selecting a miner since that will impact storing/retrieving times. A priori, might be reasonable to select miners with the smallest sector size to fit the data.

PR #438, opened 2020-06-04
index: new fault index, include new info in miners index

### LINKED ISSUE #235: index/slashing: fix strategy for new version
The current way of building the slashing index doesn't work anymore due to changes in the internals of Lotus VM. There are confirmed ideas to accommodate, but still in the process of deciding the right time since changes are still happening in the _faults_ logic of the VM.

jsign:
This PR:
- Re-implements the slashing index, which is now called _Fault index_ to be more precise. It records at which epochs miners have failed to provide proof of sectors. (Close #235)
- Include _Sector Size_ as new data in the miner index, since it sounds interesting to consider when selecting miners. If your data fits in a smaller sector than 32Gib, you might store it faster if you choose the right miner. (See #393)
- Include the _Active Deals_ count of miners in the index, which a higher number indicates the miner is storing more data and can possibly infer "more people might" be relying on him.
- `pow ffs logs` also returns history by default, so to get a clearer view of things.
- Fix setup of production docker-compose to connect Lotus to IPFS. Tested in prod when making deals with Powergate with miner `t01608`.

Changes already published in cloud: [Dashboard](http://35.238.61.140:3000/d/p4TiQMYWk/textile-filecoin?orgId=1&refresh=5s&from=now-30m&to=now), [Miner index](http://35.238.61.140:8889/index/miners), [Ask index](http://35.238.61.140:8889/index/ask), [Fault index](http://35.238.61.140:8889/index/faults)


index/faults/faults.go
```diff
@@ -119,163 +118,120 @@ func (s *Index) Close() error {
...
 			j = len(path)
 		}
-		if err := updateFromPath(s.ctx, s.api, &index, path[i:j]); err != nil {
-			return fmt.Errorf("getting update from path section: %s", err)
+
+		// Get section head and length, and get all faults happened
+		// in this section and include it into the updating index.
+		sectionLength := abi.ChainEpoch(j - i)
+		sectionHeadTs := path[j-1].Key()
+		faults, err := s.api.StateAllMinerFaults(s.ctx, sectionLength, sectionHeadTs)
```
jsign, line 191: So this is the new API we're leveraging to get miners faults.
We abandoned a more complex strategy getting internal miner state and inspecting fields.

index/faults/faults.go
```diff
@@ -119,163 +118,120 @@ func (s *Index) Close() error {
...
-			if !ok {
-				log.Debugf("casting slashedAt %v from %s at %d failed", iSlashedAt, addr, pts.Height)
-				return
-			}
-			slashedAt := uint64(fSlashedAt)
-			if slashedAt != 0 {
-				lock.Lock()
-				ret[addr] = (slashedAt)
-				lock.Unlock()
-			}
```
jsign: Good that went away.

index/miner/onchain.go
```diff
@@ -159,16 +160,33 @@ func updateForAddrs(ctx context.Context, api *apistruct.FullNodeStruct, chainInd
...
+
+	// Active deals
+	sectors, err := c.StateMinerSectors(ctx, addr, nil, false, types.EmptyTSK)
+	if err != nil {
+		return OnChainData{}, fmt.Errorf("getting sectors: %s", err)
+	}
+	var activeDeals uint64
+	for _, s := range sectors {
+		activeDeals += uint64(len(s.Info.Info.DealIDs))
+	}
```
jsign, line 185: New data added to the miners index.

index/miner/types.go
```diff
@@ -6,20 +6,21 @@ import (
...
+	Meta    MetaIndex
+	OnChain ChainIndex
 }
 
 // ChainIndex contains on-chain information about miners
 type ChainIndex struct {
 	LastUpdated int64
-	Power       map[string]Power
+	Miners      map[string]OnChainData
 }
```
jsign, line 17: So here is a change in structure of the index data. 
Instead of having data _per-dimension_ ("Power", "SectorSize", "ActiveDeals"), sounds much more useful to group the data by miner. Seems more natural. Unfortunately, that change scattered to many files but sounds like worth it.

index/miner/types.go
```diff
@@ -6,20 +6,21 @@ import (
...
-// Power contains power information of a miner
-type Power struct {
-	Power    uint64
-	Relative float64
+type OnChainData struct {
+	Power         uint64
+	RelativePower float64
+	SectorSize    uint64
+	ActiveDeals   uint64
 }
```
jsign, line 25: If adding more info in the index, would mean adding more fields here.

api/stub/faults.go
```diff
@@ -0,0 +1,30 @@
+package stub
```
asutula, line 1: We can probably get rid of all this stub code now that we have a reliable Devnet setup. No big deal either way I guess.

index/faults/faults.go
```diff
@@ -0,0 +1,244 @@
...
+	defer close(s.finished)
+	if err := s.updateIndex(); err != nil {
+		log.Errorf("initial updating faults index: %s", err)
+	}
+	for {
+		select {
+		case <-s.ctx.Done():
+			log.Info("graceful shutdown of background faults updater")
+			return
+		case <-time.After(util.AvgBlockTime):
```
asutula, line 132: Ah cool, so you are basically polling every `AvgBlockTime`? No listening for chain updates etc

index/miner/types.go
asutula: Yea good call, much better for consuming this way. Give you the total miner view in a simple way then it's easy to map to lists of single types.

jsign:
Closes #421

PR #439, opened 2020-06-05
devnet: not use volumes && linters rise bar

jsign:
- Since we made the integration of Lotus and IPFS, Powergate FFS doesn't need to share filesystem volumes with Lotus. I removed them from docker-compose-devnet, and the integrationtests, just to enforce nothing is shared.
- Leave the mount configurable since tests using directly the Deals module should share volumes, as they need to import data to the Lotus node in scenarios where that conectedness isn't necessary.
- Enable a bunch of new linters which are quite useful, mostly one to detect unused parameters which revealed some interesting situations which were fixed.

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -10,7 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - uses: actions-contrib/golangci-lint@v1
+      - uses: golangci/golangci-lint-action@v1
+        with:
+          version: v1.27
   spell-check:
      name: spell-check
      runs-on: ubuntu-latest
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,30 +3,31 @@ linters-settings:
     locale: US
 
 linters:
-  # please, do not use `enable-all`: it's deprecated and will be removed soon.
-  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
-  # disable-all: true
   enable:
     - golint
     - misspell
-
-  # don't enable:
-  # - gochecknoglobals
-  # - gocognit
-  # - godox
-  # - maligned
-  # - prealloc
+    - gocyclo
+    - bodyclose
+    - unconvert
+    - goconst
+    - goimports
+    - unparam
+    - whitespace
+    - godot
 
 issues:
   exclude-use-default: false
+
+  exclude-rules:
+    - path: cmd/powbench/runner/runner_test.go
+      linters:
+        - unused
 
 run:
   timeout: 30m
   skip-files:
     - gateway/assets.go
 
-# golangci.com configuration
-# https://github.com/golangci/golangci/wiki/Configuration
 service:
-  golangci-lint-version: 1.23.x # use the fixed version to not introduce new linters unexpectedly
+  golangci-lint-version: 1.27.x # use the fixed version to not introduce new linters unexpectedly
     
--- a/api/client/ffs.go
+++ b/api/client/ffs.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"time"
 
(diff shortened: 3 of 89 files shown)


.golangci.yml
```diff
@@ -3,30 +3,31 @@ linters-settings:
...
-  # - gocognit
-  # - godox
-  # - maligned
-  # - prealloc
+    - gocyclo
+    - bodyclose
+    - unconvert
+    - goconst
+    - goimports
+    - unparam
```
jsign, line 14: Useful. Detected functions with received params that aren't used, which is a bit of a smell.
Also, detected a bunch of functions which returned `error` but the returned value was always `nil`.

.golangci.yml
```diff
@@ -3,30 +3,31 @@ linters-settings:
...
+    - whitespace
+    - godot
```
jsign, line 16: All exporting comments should end with a period.
Took a lot of work, but is better for godocs.

.golangci.yml
```diff
@@ -3,30 +3,31 @@ linters-settings:
...
-  # - godox
-  # - maligned
-  # - prealloc
+    - gocyclo
+    - bodyclose
+    - unconvert
+    - goconst
+    - goimports
+    - unparam
+    - whitespace
```
jsign, line 15: Detects unecessary whitespaces, like:
```
return

}
```
And similars.

.golangci.yml
```diff
@@ -3,30 +3,31 @@ linters-settings:
...
-
-  # don't enable:
-  # - gochecknoglobals
-  # - gocognit
-  # - godox
-  # - maligned
-  # - prealloc
+    - gocyclo
+    - bodyclose
+    - unconvert
```
jsign, line 11: Detected unecessary conversions, like:
```
hello := uint64(20)
...
foo := unit64(hello)
```

Cases fixed are due to changing types over time, and making conversions unnecessary.

.golangci.yml
```diff
@@ -3,30 +3,31 @@ linters-settings:
     locale: US
 
 linters:
-  # please, do not use `enable-all`: it's deprecated and will be removed soon.
-  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
-  # disable-all: true
   enable:
```
jsign, line 6: Have others that I'd like to add but require some work that isn't priority now... more in the future.

.github/workflows/review.yml
```diff
@@ -10,7 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - uses: actions-contrib/golangci-lint@v1
+      - uses: golangci/golangci-lint-action@v1
```
jsign, line 13: Switched to an official version, which doesn't force v1.23.

api/client/ffs.go
```diff
@@ -572,7 +573,7 @@ func protoToDealError(des []*rpc.DealError) ([]ffs.DealError, error) {
 			var err error
 			propCid, err = cid.Decode(de.ProposalCid)
 			if err != nil {
-				propCid = cid.Undef
+				return nil, fmt.Errorf("proposal cid is invalid")
```
jsign, line 614: This was interesting. It always returned nil error.
@asutula , I thought this would be a better decision, agree?

ffs/filcold/filcold.go
```diff
@@ -272,7 +271,7 @@ func (fc *FilCold) getCidSize(ctx context.Context, c cid.Cid) (uint64, error) {
 	return uint64(s.CumulativeSize), nil
 }
 
-func makeDealConfigs(ctx context.Context, ms ffs.MinerSelector, cntMiners int, f ffs.MinerSelectorFilter) ([]deals.StorageDealConfig, error) {
+func makeDealConfigs(ms ffs.MinerSelector, cntMiners int, f ffs.MinerSelectorFilter) ([]deals.StorageDealConfig, error) {
```
jsign, line 274: Unused `ctx`.

ffs/scheduler/scheduler.go
```diff
@@ -359,11 +354,11 @@ func (s *Scheduler) executeQueuedJobs(ctx context.Context) {
 	}
 }
 
-func (s *Scheduler) executeQueuedJob(j ffs.Job) error {
+func (s *Scheduler) executeQueuedJob(j ffs.Job) {
```
jsign, line 357: Always returned `nil` error.

ffs/scheduler/scheduler.go
```diff
@@ -500,7 +494,7 @@ func (s *Scheduler) getRefreshedHotInfo(ctx context.Context, c cid.Cid, curr ffs
 	return curr, nil
 }
 
-func (s *Scheduler) getRefreshedColdInfo(ctx context.Context, c cid.Cid, curr ffs.ColdInfo) (ffs.ColdInfo, error) {
+func (s *Scheduler) getRefreshedColdInfo(ctx context.Context, curr ffs.ColdInfo) (ffs.ColdInfo, error) {
```
jsign, line 497: Interesting, the `c` param was unused.

scripts/gen-js-protos.sh
```diff
@@ -19,8 +19,8 @@ mkdir -p $PTOTOC_GEN_OUT_DIR
 
 printf '{"name":"%s", "version":"%s", "files":["%s"]}' $LIB_NAME $LIB_VERSION $LIB_SRC_DIR > $OUT_PATH/package.json
 
-(cd $OUT_PATH && npm install --save-dev ts-protoc-gen @types/google-protobuf)
-(cd $OUT_PATH && npm install google-protobuf @improbable-eng/grpc-web)
+(cd $OUT_PATH && npm install --save-dev ts-protoc-gen)
+(cd $OUT_PATH && npm install google-protobuf @improbable-eng/grpc-web @types/google-protobuf)
```
jsign, line 23: @asutula , check?

asutula: perfect. thx

api/client/ffs.go
asutula: Yes, I like that. Think I copy/pasted some code from the lotus cli here. Your solution is what I would have don't if I typed it myself or noticed this before ;)

PR #440, opened 2020-06-05
ffs: ProposalCid undefined correct handling

jsign:
So the new linters from past PR already uncovered an edge behavior.
In short, if a `cid.Cid` is `cid.Undef`, then it's stringified value is `b`, not empty. 

This was discovered by upgrading `textile` PG version, in an integration test testing a failure scenario.

### DIFF
--- a/api/client/ffs.go
+++ b/api/client/ffs.go
@@ -607,7 +607,7 @@ func fromRPCDealErrors(des []*rpc.DealError) ([]ffs.DealError, error) {
 	res := make([]ffs.DealError, len(des))
 	for i, de := range des {
 		var propCid cid.Cid
-		if de.ProposalCid != "" {
+		if de.ProposalCid != "" && de.ProposalCid != "b" {
 			var err error
 			propCid, err = cid.Decode(de.ProposalCid)
 			if err != nil {
--- a/deals/rpc/rpc.go
+++ b/deals/rpc/rpc.go
@@ -139,8 +139,12 @@ func (s *RPC) Watch(req *WatchRequest, srv RPCService_WatchServer) error {
 	}
 
 	for update := range ch {
+		var strProposalCid string
+		if update.ProposalCid.Defined() {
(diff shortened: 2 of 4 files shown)


api/client/ffs.go
```diff
@@ -607,7 +607,7 @@ func fromRPCDealErrors(des []*rpc.DealError) ([]ffs.DealError, error) {
 	res := make([]ffs.DealError, len(des))
 	for i, de := range des {
 		var propCid cid.Cid
-		if de.ProposalCid != "" {
+		if de.ProposalCid != "" && de.ProposalCid != "b" {
```
jsign, line 610: So, if the original  value was `cid.Undef`, the stringified result is `"b"`.  Cover in the client this edge case just in case it gets slipped in the RPC layer.

deals/rpc/rpc.go
```diff
@@ -139,8 +139,12 @@ func (s *RPC) Watch(req *WatchRequest, srv RPCService_WatchServer) error {
 	}
 
 	for update := range ch {
+		var strProposalCid string
+		if update.ProposalCid.Defined() {
+			strProposalCid = update.ProposalCid.String()
+		}
```
jsign, line 145: Everyone returning a stringified version, should take care of returning empty for coherence.
The client will have the safety-net mentioned in the above comment, but try to be coherent on returning empty for undefined.

ffs/integrationtest/integration_test.go
```diff
@@ -1203,7 +1203,7 @@ func TestFailedJobMessage(t *testing.T) {
 	require.NotEmpty(t, job.ErrCause)
 	require.Len(t, job.DealErrors, 1)
 	de := job.DealErrors[0]
-	require.NotEmpty(t, de.ProposalCid.String())
+	require.False(t, de.ProposalCid.Defined())
```
jsign, line 1206: This assertion was wrong... nice to catch.

PR #442, opened 2020-06-08
Correctly calculate size for tight PricePerEpoch calculation & interop update

### LINKED ISSUE #441: deals: correct min price calculation
A _Storage Deal Proposal_ should have the _PricePerEpoch_ that the client wants to pay the Miner for storing the data. This price can be loosely described as: _the data size in GB * PricePerEpoch of Miner storage ask_.

If the proposed price per epoch is less than that value, the miner will reject the deal since the proposed pay per epoch would be less that what's asking. So, underestimating that value would result in rejected deals. Also, overestimating that value would incur in paying _more_ that would be possible.

The main _issue_ is that the _data size in GB_ part of the calculation, isn't the raw data size but a more complex number. The correct _size_ should be the size of the CAR-serialize of the DAG, plus padding necessary by Filecoin (to create commitments, and proofs).

An ideal solution is calculating the size without generating the CAR file, which would imply being tied with the file size which might be costly (CPU/mem/storage).

jsign:
See #141 to get a description of the problem and an ideal solution, which is what was implemented here.

Also updates to latest lotus and devnet (interop 6.8).

Closes #441


ffs/filcold/filcold.go
```diff
@@ -69,10 +74,14 @@ func (fc *FilCold) Store(ctx context.Context, c cid.Cid, cfg ffs.FilConfig) ([]c
...
 		return nil, nil, 0, fmt.Errorf("making deal configs: %s", err)
 	}
-	size, err := fc.getCidSize(ctx, c)
+
+	fc.l.Log(ctx, c, "Calculating piece size...")
+	size, err := fc.calculatePieceSize(ctx, c)
 	if err != nil {
 		return nil, nil, 0, fmt.Errorf("getting cid cummulative size: %s", err)
 	}
+	fc.l.Log(ctx, c, "Estimated piece size is %d bytes.", size)
```
jsign, line 83: The piece size calculation is added to the human-friendly logs because it isn't that much negligible work.

ffs/filcold/filcold.go
```diff
@@ -260,15 +269,55 @@ func (fc *FilCold) WaitForDeals(ctx context.Context, c cid.Cid, proposals []cid.
...
 	return res, errors, nil
 }
 
-func (fc *FilCold) getCidSize(ctx context.Context, c cid.Cid) (uint64, error) {
-	s, err := fc.ipfs.Object().Stat(ctx, path.IpfsPath(c))
+// estimatePieceSize estimates the size of the Piece that will be built by the filecoin client
+// when making the deal. This calculation should consider the unique node sizes of the DAG, and
+// padding. It's important to not underestimate the size since that would lead to deal rejection
+// since the miner won't accept the further calculated PricePerEpoch.
+func (fc *FilCold) calculatePieceSize(ctx context.Context, c cid.Cid) (uint64, error) {
```
jsign, line 280: So the thing here is calculating the CAR file serialization without making the CAR, plus considering the padding that is needed for pieces in Filecoin. 

Not making the CAR makes this scale for bigger files. So this thing is bound by the # of nodes in the DAG, which presumably should be ~_log(size of data)_ API calls (assuming the UnixFS DAG is balanced). Work isn't related with byte size since no data is fetched in the apis.

jsign: I played making some tricks about returning the calculated size minus 1 to let the deal fail, and it seems its calculating exactly the minimum price. It may need more data sizes tests, but looks quite promising. We'll discover if this isn't that correct, because the deal proposal will fail for some use case. So, quite easy to spot.

deals/deals_test.go
```diff
@@ -98,14 +98,14 @@ func storeMultiMiner(m *Module, client *apistruct.FullNodeStruct, numMiners int,
...
-	dataCid, _, err := m.Import(ctx, bytes.NewReader(data), false)
+	dataCid, size, err := m.Import(ctx, bytes.NewReader(data), false)
 	if err != nil {
 		return cid.Undef, err
 	}
 	if !dataCid.Defined() {
 		return cid.Undef, fmt.Errorf("data cid is undefined")
 	}
-	srs, err := m.Store(ctx, addr.String(), dataCid, uint64(len(data)), cfgs, 1000)
+	srs, err := m.Store(ctx, addr.String(), dataCid, 2*uint64(size), cfgs, 1000)
```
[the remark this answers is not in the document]
jsign: So a good part of using FFS, is that it will calculate the correct size for your data. If you use the Deals Module directly, you should do the calculation yourself.

PR #447, opened 2020-06-15
Generate and add pow cli docs

asutula:
- New `pow` command `docs [out location]` that generates markup docs
- New `make docs-pow` to generate cli docs more easily
- New CI `ensure docs` that makes sure generated cli docs are up to date with code
- Adds the generated docs to `cli-docs/pow/`

Maybe not _the_ long term solution, but just a start. Eventually, we should generate some html docs and/or add this info to the general Textile docs.

Another option I considered is generating and committing the docs during CI merge to master to a separate branch that just holds these docs. Would be less discoverable, but would keep the repo cleaner. Let me know if that would be better.


andrewxhill:
Once this is merged, I'll add a build script to the docs repo that pull them in for rendering. easy peasy

.github/workflows/ensure-docs.yml
```diff
@@ -0,0 +1,23 @@
...
+    name: pow
+    runs-on: ubuntu-latest
+    steps:
+      - name: check out code
+        uses: actions/checkout@v1
+      - name: generate docs
+        run: make docs-pow
+      - name: detect changes
+        id: changes
+        uses: UnicornGlobal/has-changes-action@v1.0.11
```
asutula, line 20: Just using this plugin because checking for git changes AND untracked files is actually not super simple. This is simple.

Makefile
```diff
@@ -16,6 +16,10 @@ build-powbench:
 	$(BUILD_FLAGS) go install ./cmd/powbench
 .PHONY: build-powbench
 
+docs-pow:
```
asutula, line 19: You'd run this after making changes to the `pow` during development.

cmd/pow/cmd/docs.go
```diff
@@ -0,0 +1,28 @@
...
+	Short: "Generate markdown docs for pow command",
+	Long:  `Generate markdown docs for pow command`,
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		dir := args[0]
+		if _, err := os.Stat(dir); os.IsNotExist(err) {
+			err = os.MkdirAll(dir, os.ModePerm)
+			checkErr(err)
+		}
+		err := doc.GenMarkdownTree(rootCmd, args[0])
```
asutula, line 25: To generate docs, you need access to the cobra root command. Decided to just add a new command to `pow` to achieve this. Kinda weird to have a command to generate docs for your command, but maybe kinda cool actually. Works.

PR #457, opened 2020-06-18
ffs/scheduler: fix parallel runner

jsign:
Changes the scheduler runner to don't paralellize in batches, but as soon as a rate-limited slot is available or a run Job finishes, since may unblock some waiting new Push for the same Cid.

### DIFF
--- a/ffs/filcold/filcold.go
+++ b/ffs/filcold/filcold.go
@@ -262,6 +262,9 @@ func (fc *FilCold) WaitForDeals(ctx context.Context, c cid.Cid, proposals []cid.
 	}
 
 	if len(activeProposals) == 0 {
+		if ctx.Err() != nil {
+			return nil, errors, fmt.Errorf("all accepted proposals were untracked due to cancellation")
+		}
 		return nil, errors, fmt.Errorf("all accepted proposals failed before becoming active")
 	}
 	res := make([]ffs.FilStorage, 0, len(activeProposals))
--- a/ffs/integrationtest/integration_test.go
+++ b/ffs/integrationtest/integration_test.go
@@ -1294,8 +1294,15 @@ func TestParallelExecution(t *testing.T) {
 		// being batched in the same scheduler run.
 		time.Sleep(time.Millisecond * 100)
 	}
+	// Check that all jobs should be immediately in the Executing status, since
+	// the default max parallel runs is 50. So all should get in.
 	for i := 0; i < len(jids); i++ {
 		requireJobState(t, fapi, jids[i], ffs.Executing)
+	}
+
+	// Now just check that all of them finish successfully.
+	for i := 0; i < len(jids); i++ {
+		requireJobState(t, fapi, jids[i], ffs.Success)
 		requireCidConfig(t, fapi, cids[i], nil)
 	}
 }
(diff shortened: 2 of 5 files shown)


ffs/filcold/filcold.go
```diff
@@ -262,6 +262,9 @@ func (fc *FilCold) WaitForDeals(ctx context.Context, c cid.Cid, proposals []cid.
 	}
 
 	if len(activeProposals) == 0 {
+		if ctx.Err() != nil {
+			return nil, errors, fmt.Errorf("all accepted proposals were untracked due to cancellation")
+		}
```
jsign, line 267: Better error message if making deas failed but because Powergate is shutting down and things got canceled in purpose.

ffs/integrationtest/integration_test.go
```diff
@@ -1275,6 +1275,38 @@ func TestResumeScheduler(t *testing.T) {
 	require.Equal(t, 1, len(sh.Cold.Filecoin.Proposals)) // Check only one deal still exits.
 }
 
+func TestParallelExecution(t *testing.T) {
```
jsign, line 1278: Add to tests a case where three Cids are pushed and all of them should be in _Executing_ immediately. An of course, finish correctly.

ffs/integrationtest/integration_test.go
```diff
@@ -1389,6 +1421,10 @@ func requireJobState(t *testing.T, fapi *api.API, jid ffs.JobID, status ffs.JobS
 			require.True(t, ok)
 			require.Equal(t, jid, job.ID)
 			if job.Status == ffs.Queued || job.Status == ffs.Executing {
+				if job.Status == status {
+					stop = true
+					res = job
+				}
```
jsign, line 1427: Considering that now `requireJobState` might be called for _Executing_ scenario (instead of always final states), just stopping earlier if that's the case.

ffs/scheduler/scheduler.go
```diff
@@ -42,7 +42,8 @@ type Scheduler struct {
 	cis *cistore.Store
```
jsign, line 42: The change in the scheduler runner was the following.

There're two events that should trigger queued Job evaluation for run:
- When a new Job is queued.
- When a Executing Job finished.

The last point is important for two cases:
- If a queued job that's runnable couldn't get into _Executing_, is because the runner rate limiting is full. So it needs to be reevaluated for a slot whenever someone finished.
- Even if the rate-limiting isn' t full, a queued Job could not be runnable. That's the case if there's another Job fo the same Cid running. If that's the case, it can only be run when that executing Job finishes... hence the finishing triggering event.

tests/ldevnet.go
```diff
@@ -34,7 +34,7 @@ func LaunchDevnetDocker(t *testing.T, numMiners int, ipfsMaddr string, mountVolu
 	}
 
 	repository := "textile/lotus-devnet"
-	tag := "sha-72026a2"
+	tag := "sha-ff3d28e"
```
jsign, line 37: Change to latest too for integrationtests.

tests/ldevnet.go
```diff
@@ -27,14 +27,15 @@ func LaunchDevnetDocker(t *testing.T, numMiners int, ipfsMaddr string, mountVolu
 		devnetEnv("NUMMINERS", strconv.Itoa(numMiners)),
 		devnetEnv("SPEED", "500"),
 		devnetEnv("IPFSADDR", ipfsMaddr),
+		devnetEnv("BIGSECTORS", false),
```
jsign, line 30: This change is because now default this was switched to true in `lotus-devnet`, so not specifying a value would be the opposite as before.

PR #459, opened 2020-06-18
fix(docs): remove docs command from cli docs

andrewxhill:
also removes it from --help output


asutula:
I had thought about this, but then imagined maybe it would be useful for someone somewhere. But if you prefer it gone, fine with me.

andrewxhill:
yeah... just seems tidier

PR #460, opened 2020-06-19
update to testnet v0.4.0

jsign:
This PR updates the bootstrap peers, `lotus-devnet`, and `lotus` to the corresponding tags that targets testnet `v0.4.0`.

I updated `lotus-devnet` as usual, but I made a mirroring tag policy with `lotus`. That's to say, if `lotus-devnet` is using a tagged version of `lotus`, then I'll tag `lotus-devnet` with the same tag. This starts the path of having a clearer relationship with `lotus-devnet` DockerHub images, and Lotus versions. 

For example, now `lotus-devnet` is using `v0.4.0` tag of `lotus`. So, its tag in DockerHub will be `v0.4.0`. Now, if we make a new release of `lotus-devnet` that includes an extra commit which is untagged, we could tag it `v0.4.0-1`, `v0.4.0-2` etc. I think using that notation is fine, since doing `v0.4.1` would be incorrect since that indicates a Lotus `v0.4.1` tag.

For `lotus-build` which builds the production image of Lotus in DockerHub, it already tagged the images with the same sha from the Lotus commit. So there was a clear correspondence. Now if the commit is also tagged in Lotus, it will also tag it in DockerHub.

In short, at this moment we have: 
- `textile/lotus-devnet:v0.4.0`: Which means is a testnet running Lotus code at v0.4.0. If further versions add commits from a tagged version, should be called `v0.4.0-x`.
- `textile/lotus:v0.4.0`: which means is the testnet with Lotus tag v0.4.0.

Starting to align some planets for easier release flow.


docker/docker-compose-devnet.yaml
```diff
@@ -26,7 +26,7 @@ services:
       - 5001:5001
 
   lotus:
-    image: textile/lotus-devnet:sha-ff3d28e
+    image: textile/lotus-devnet:v0.4.0
```
asutula, line 29: Huge improvement

PR #461, opened 2020-06-19
Compose improvements and release publishing

asutula:
The goal of this one is:

- Streamline docker image dependency version numbering a little, making it so you just update any given dependency version in one place
- Provide some composability in the Docker Compose setup so that it's easier to compose variations of the powergate stack while sharing most configuration
- At release time, create a version of the Docker Compose artifacts and Makefile that run the release-tagged version of the powergate image (vs building from source)
- Zip up these Docker Compose artifacts and Makefile and make them available on the GH Release page

When it came to how much to decompose the docker compose files, I went down the path of breaking them into as many unique, reusable pieces as possible, but that seemed confusing and hard to manage. Final approach was to decompose only things that specify an image version number or build context AND were repeated in more than one place.

I also considered using `docker-compose -f ... -f ... config` to create static compose files to publish to the GH release, but that resulted in loosing the ability to dynamically specify the `BIGSECTORS` input to the compose process. Plus, it's easier to just tell people to run `make up` or `make devnet` or `BIGSECTORS=false make devnet`. The this works the same in a working copy of the Powergate repo as it does in the downloaded docker release artifacts.


.github/workflows/release.yml
```diff
@@ -1,9 +1,36 @@
-name: Publish JS gRPC bindings
+name: Release
```
asutula, line 1: No reason to have multiple GH Action files to do things we want done on release. Consolidated here.

.github/workflows/release.yml
```diff
@@ -1,9 +1,34 @@
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v1
+      - name: Get latest tag
+        id: latesttag
+        uses: "WyriHaximus/github-action-get-previous-tag@master"
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Update powergate image version
+        run: sed -i.bak 's/latest/${{steps.latesttag.outputs.tag}}/g' docker/powergate-image.yaml
```
asutula, line 18: Updates the powergate image tag to use in the compose setup to be the release tag.

.github/workflows/release.yml
```diff
@@ -1,9 +1,34 @@
...
+      - name: Update Makefile
+        run: sed -i.bak 's/powergate-build-context/powergate-image/g' docker/Makefile
```
asutula, line 20: Updates the Makefile running docker-compose to use a tagged powergate image instead of a build context to build from source.

.github/workflows/release.yml
```diff
@@ -1,9 +1,34 @@
...
+        run: rm -rf docker/*.bak
+      - name: Create archive
+        run: |
+          mv docker powergate-docker
+          zip -r powergate-docker.zip powergate-docker
+      - name: Upload artifacts to release
+        uses: AButler/upload-release-assets@v2.0
+        with:
+          files: 'powergate-docker.zip'
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
asutula, line 31: Give the archive a better name and upload it to the release

docker/Makefile
```diff
@@ -1,16 +1,28 @@
...
+		up --build 
 .PHONY: up
 
 devnet: 
-	docker-compose -p devnet -f docker-compose-devnet.yaml up --build -V
+	docker-compose -p devnet \
+		-f docker-compose-devnet.yaml \
+		-f ipfs-image.yaml \
+		-f powergate-build-context.yaml \
+		up --build -V
```
asutula, line 27: Ipfs image version is just specified in one place `ipfs-image.yaml`, so bring it in. We choose to use the build context to build from source by default, so bring it in from `powergate-build-context.yaml`. This is changed at release time to use a tagged image.

docker/docker-compose-devnet.yaml
```diff
@@ -21,7 +19,6 @@ services:
     restart: unless-stopped
 
   ipfs:
-    image: ipfs/go-ipfs:v0.5.1
```
asutula, line 24: Leave all this out of the base config since it's now specified in other composed files that are merged in.

docker/docker-compose.yaml
```diff
@@ -44,7 +42,6 @@ services:
     restart: unless-stopped
 
   ipfs:
-    image: ipfs/go-ipfs:v0.5.1
```
asutula, line 47: Same thing here.

docker/ipfs-image.yaml
```diff
@@ -0,0 +1,7 @@
+version: '3.7'
+
+services:
+
+  ipfs:
+    image: ipfs/go-ipfs:v0.5.1
```
asutula, line 6: One place to specify ipfs image version

docker/powergate-build-context.yaml
```diff
@@ -0,0 +1,8 @@
+version: '3.7'
+
+services:
+  
+  powergate:
+    build:
+      context: ../
```
asutula, line 7: Pulled in by default so we build from source

docker/powergate-image.yaml
```diff
@@ -0,0 +1,5 @@
+version: "3.7"
+
+services:
+  powergate:
+    image: textile/powergate:latest
```
asutula, line 5: At release time, this is updated to the release tag version and included in the released artifacts. Makefile is update to pull this in instead of `powergate-build-context.yaml`

asutula:
The subdirectories `dashboards`, `grafana` and `prometheus` are being included in the release zip file... not sure if they should be or need to be.

.github/workflows/release.yml
jsign: Nice, so double magic. First replace `latest`, and then switch context to tagged release. 💯

docker/Makefile
jsign: Nice.

jsign:
> The subdirectories `dashboards`, `grafana` and `prometheus` are being included in the release zip file... not sure if they should be or need to be.

I think that's OK, since `make up` will spinup the metrics-related containers which binds volumes to these assets. 👍

.github/workflows/release.yml
asutula: Yep, exactly

PR #466, opened 2020-06-22
ffs: better deal proposal and flow handling

jsign:
Improve deal proposal and tracking flow handling.

### DIFF
--- a/ffs/filcold/filcold.go
+++ b/ffs/filcold/filcold.go
@@ -224,7 +224,7 @@ func (fc *FilCold) WaitForDeals(ctx context.Context, c cid.Cid, proposals []cid.
 	defer cancel()
 	chDi, err := fc.dm.Watch(ctx, proposals)
 	if err != nil {
-		return nil, errors, err
+		return nil, errors, fmt.Errorf("watching proposals in deals module: %s", err)
 	}
 
 	activeProposals := make(map[cid.Cid]*ffs.FilStorage)
@@ -257,12 +257,8 @@ func (fc *FilCold) WaitForDeals(ctx context.Context, c cid.Cid, proposals []cid.
 			break
 		}
 	}
-
-	if len(activeProposals) == 0 {
-		if ctx.Err() != nil {
-			return nil, errors, fmt.Errorf("all accepted proposals were untracked due to cancellation")
-		}
-		return nil, errors, fmt.Errorf("all accepted proposals failed before becoming active")
+	if ctx.Err() != nil {
+		return nil, errors, fmt.Errorf("all accepted proposals were untracked due to cancellation")
 	}
 	res := make([]ffs.FilStorage, 0, len(activeProposals))
 	for _, v := range activeProposals {
--- a/ffs/filcold/filcold.go
+++ b/ffs/filcold/filcold.go
@@ -86,9 +86,6 @@ func (fc *FilCold) Store(ctx context.Context, c cid.Cid, cfg ffs.FilConfig) ([]c
 	if err != nil {
 		return nil, nil, 0, fmt.Errorf("starting deals: %s", err)
 	}
-	if len(okDeals) == 0 {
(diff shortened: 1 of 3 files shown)


ffs/filcold/filcold.go
```diff
@@ -86,9 +86,6 @@ func (fc *FilCold) Store(ctx context.Context, c cid.Cid, cfg ffs.FilConfig) ([]c
 	if err != nil {
 		return nil, nil, 0, fmt.Errorf("starting deals: %s", err)
 	}
-	if len(okDeals) == 0 {
-		return nil, failedStartingDeals, 0, fmt.Errorf("all proposed deals where rejected")
-	}
```
jsign, line 91: `Store` should be biased by the result. It's the caller responsibility to consider if his use-case is a failure or not. Store just fire things and indicates an error in extraordinary conditions. Rejected deals aren't extraordinary.
The caller can just check that all the deals failed if is in its interests (i.e: the `len` of the failed slice is equal to all proposed).

ffs/filcold/filcold.go
```diff
@@ -260,12 +257,8 @@ func (fc *FilCold) WaitForDeals(ctx context.Context, c cid.Cid, proposals []cid.
 			break
 		}
 	}
-
-	if len(activeProposals) == 0 {
-		if ctx.Err() != nil {
-			return nil, errors, fmt.Errorf("all accepted proposals were untracked due to cancellation")
-		}
-		return nil, errors, fmt.Errorf("all accepted proposals failed before becoming active")
```
jsign, line 268: Same here. `WaitForDeals` just wait for deals until final status. 
If all didn't reach `Success` or whatever, isn't something related to `WaitForDeals`. `WaitForDeals` waits... then the interpretation of "failure" is up to the caller.
All this makes simpler to use these APIs... no "border"cases error cases or similar.

ffs/scheduler/scheduler.go
```diff
@@ -556,25 +556,36 @@ func (s *Scheduler) executeColdStorage(ctx context.Context, curr ffs.CidInfo, cf
...
-		return ffs.ColdInfo{}, allErrors, fmt.Errorf("watching deals unfold: %s", err)
+		return ffs.ColdInfo{}, nil, fmt.Errorf("watching deals unfold: %s", err)
 	}
+	allErrors = append(allErrors, failedDeals...)
 	if err := s.js.RemoveStartedDeals(curr.Cid); err != nil {
 		return ffs.ColdInfo{}, allErrors, fmt.Errorf("removing temporal started deals storage: %s", err)
 	}
+	if len(failedDeals) == len(startedProposals) {
+		return ffs.ColdInfo{}, allErrors, fmt.Errorf("all started deals failed")
+	}
```
jsign, line 586: Here's the caller having responsibility to consider that if all proposed deals in a Job failed, then the Job failed. 
In some other case of a `WaitForDeals` call, maybe the caller wants to interpret things differently.

ffs/scheduler/scheduler.go
```diff
@@ -556,25 +556,36 @@ func (s *Scheduler) executeColdStorage(ctx context.Context, curr ffs.CidInfo, cf
...
 	s.l.Log(ctx, curr.Cid, "Current replication factor is lower than desired, making %d new deals...", deltaFilConfig.RepFactor)
 	startedProposals, rejectedProposals, size, err := s.cs.Store(ctx, curr.Cid, deltaFilConfig)
 	if err != nil {
+		s.l.Log(ctx, curr.Cid, "Starting deals failed, with cause: %s", err)
 		return ffs.ColdInfo{}, rejectedProposals, err
 	}
 	allErrors = append(allErrors, rejectedProposals...)
+	if len(startedProposals) == 0 {
+		return ffs.ColdInfo{}, allErrors, fmt.Errorf("all proposals were rejected")
+	}
```
jsign, line 570: Similar to below comment.

Issue #468, opened 2020-06-22
Add Job execution timeout


jsign:
In some strange conditions, maybe the miner keeps their deals in a transitive status for too much time.
It sounds like a good idea to bound Jobs executing times, similar to a timeout, to have the guarantee that all Job executions will end.

Since sealing times can take >6hs, we're talking about a very long timeout. For example, a reasonable timeout can be 12hs. It doesn't make sense to continue waiting for deals status changes after 12hs have passed. If we're in that situation, the reasonable decision is to consider that deal as _Failed_.

PR #472, opened 2020-06-24
Multiple features and improvements

### LINKED ISSUE #469: ffs: cancel executing job
If a user pushes a CidConfig with a high replication factor, there's a high chance of being locked on long-waiting times due to miners sealing forever or long sealing queues.

A Job in _Executing_ status will continue in that status until all of the involved deals reach a final status (success or failure). The sealing duration for miners has a lot of volatility, and in particular if some of the miners are very slow, buggy, or have long queues that can take hours/days.

The problem with that scenario is having Jobs in the _Executing_ status for an unreasonable time. When a Cid Job is _Executing_, it will queue any further CidConfig that's being pushed for the same Cid. This because running different Configs for the same Cid doesn't make sense or it might be simply wrong. 

Allow the user to say: "I want to cancel the execution of Job with ID X". This will make waiting for deals stop, save the successful proposals in the CidInfo store, and finalize the Job with the _Canceled_ status.
This gives the freedom to the client to cancel a Job that might be taking too long because of miners being slow, or just because it wants to execute a changed PushConfig (which needs to have the free-path of not other Executing job for the same Cid).

### LINKED ISSUE #470: ffs: enable masterAddr auto-creation & faucet funding
Create flag to enable automatically creating a `masterAddress`, and auto-fund it with testnet Faucet.

This will allow less manual actions from users on testnet, since their powergate instance will always have a master addr with funds which will fund newly created FFS instances. If this new flag isn't used, Powergate can be used as usual (without a master addr, or with provided one).

### LINKED ISSUE #471: powd: improve logging on shutdown
Improve logging on shutdown to account for times of closing the different components.

jsign:
Features:
- Closes #470: See issue description.
- Closes #469: See issue description.

Improvements/Fixes:
- Fixes accounting for cumulative size in `Replace` API (as is it done in `Store`), that's the correct size.
- Closes #471: See issue description. This enabled to discover some cancellation improvement handling, and some _possible_ problems in Lotus canceling some API calls (started exploring [here](https://github.com/filecoin-project/go-fil-markets/issues/292) if that's the case).
- For some strange reason, the askIndex `.Store()` wasn't being used, so at every start, the known "ask storage offers" where empty, so it needed to wait to have data. All was ready to do it, the call was just missing.
- Fixed a deadlock in the closing logic of the Reputation module, which explains some weird behavior in some cases while closing Powergate. Now fixed.


ffs/api/api_logs.go
```diff
@@ -34,7 +34,7 @@ func (i *API) WatchLogs(ctx context.Context, ch chan<- ffs.LogEntry, c cid.Cid,
 		}
 	}
 
-	ichan := make(chan ffs.LogEntry)
+	ichan := make(chan ffs.LogEntry, cap(ch))
```
jsign, line 37: Use same internal channel capacity as defined by the caller.

ffs/filcold/filcold.go
```diff
@@ -3,6 +3,7 @@ package filcold
 import (
```
jsign, line 3: TL;DR: I changed the _ColdStorage_ API `WaitForDeals` for `WaitForDeal`. If the caller want's to track multiple ones, it can do multiple calls concurrently. 

This greatly simplifies both this implementation, and gives more control to the caller.

Why was this necessary? Since now we support the client saying "cancel this executing job", the caller nice finer grained control about saving progress in individual deals is waiting for. So it can cancel the ones that are still waiting, and save the result of the ones that finished already.

In the previous implementations, since all deals were "waited in a single block", we couldn't solve that easily. That's saying, saving the ones that already finished and stopping the ongoing ones. They all were in the same `WaitForDeals` call, which was inconvenient.

The nice thing is that doing this change not only gives more control, but simplifies the code.

ffs/interfaces.go
```diff
@@ -72,10 +76,11 @@ type ColdStorage interface {
...
 
-	// WaitForDeals block until all provided Deal Proposals reach a
-	// final state, and return the success result plus failed deals
-	// that didn't reach success final status.
-	WaitForDeals(context.Context, cid.Cid, []cid.Cid) ([]FilStorage, []DealError, error)
+	// WaitForDeal blocks the provided Deal Proposal reach a
+	// final state. If the deal finishes successfully it returns a FilStorage
+	// result. If the deal finished with error, it returns a ffs.DealError
+	// error result, so it should be considered in error handling.
+	WaitForDeal(context.Context, cid.Cid, cid.Cid) (FilStorage, error)
```
jsign, line 83: Now `WaitForDeal` is a simpler API. It returns the success data, or an error.
The error can be of type `ffs.DealError` which has further info if the deal errored.
If the error is of other sort of nature (network timeout, or some other weird one), it would be just that error.

ffs/scheduler/scheduler.go
```diff
@@ -204,8 +208,24 @@ func (s *Scheduler) GetLogs(ctx context.Context, c cid.Cid) ([]ffs.LogEntry, err
 	return lgs, nil
 }
 
+// Cancel cancels an executing Job.
+func (s *Scheduler) Cancel(jid ffs.JobID) error {
```
jsign, line 212: The canceling feature is simple.
Each executing job has a dummy channel that should be closed if some wants to cancel it.

ffs/scheduler/scheduler.go
```diff
@@ -374,15 +395,39 @@ forLoop:
...
 
-	ctx := context.WithValue(s.ctx, ffs.CtxKeyJid, j.ID)
+	ctx, cancel := context.WithCancel(context.WithValue(context.Background(), ffs.CtxKeyJid, j.ID))
+	defer cancel()
+	go func() {
+		// If the user called Cancel to cancel Job execution,
+		// we cancel the context to finish.
+		<-cancelChan
+		cancel()
+	}()
```
jsign, line 423: Someone canceled the Job?. Then cancel its context of execution, so it will stop.

ffs/scheduler/scheduler.go
```diff
@@ -391,15 +436,30 @@ func (s *Scheduler) executeQueuedJob(j ffs.Job) {
...
 	}
-	if err := s.js.Finalize(j.ID, ffs.Success, nil, dealErrors); err != nil {
+
+	finalStatus := ffs.Success
+	// Detect if user-cancelation was triggered
+	select {
+	case <-cancelChan:
+		finalStatus = ffs.Canceled
+	default:
+	}
```
jsign, line 451: So the Job is terminating. Is this because it finished as expected, or because someone canceled it?

ffs/scheduler/scheduler.go
```diff
@@ -593,6 +651,41 @@ func (s *Scheduler) executeColdStorage(ctx context.Context, curr ffs.CidInfo, cf
 	}}, allErrors, nil
 }
 
+func (s *Scheduler) waitForDeals(ctx context.Context, c cid.Cid, startedProposals []cid.Cid) ([]ffs.FilStorage, []ffs.DealError) {
```
jsign, line 654: This is a helper func in the scheduler, if some logic wants to wait for multiple deals... considering now that the API for _ColdStorage_ is `WaitForDeal`.

index/ask/runner/runner.go
```diff
@@ -176,12 +178,20 @@ func (ai *Runner) update() error {
...
 
 	ai.lock.Lock()
 	ai.index = newIndex
 	ai.orderedAsks = cache
 	ai.lock.Unlock()
 
+	if err := ai.store.Save(newIndex); err != nil {
+		return fmt.Errorf("persisting ask index: %s", err)
+	}
+
```
jsign, line 194: Missing saving the new ask-index data in the datastore.

wallet/wallet.go
```diff
@@ -19,12 +28,32 @@ type Module struct {
...
 
 // New creates a new wallet module.
-func New(api *apistruct.FullNodeStruct, maddr address.Address, iam big.Int) (*Module, error) {
+func New(api *apistruct.FullNodeStruct, maddr address.Address, iam big.Int, autocreate bool) (*Module, error) {
 	m := &Module{
 		api:        api,
 		iAmount:    &iam,
 		masterAddr: maddr,
 	}
+	if maddr == address.Undef && autocreate {
```
jsign, line 37: Autocreate feature.

wallet/wallet.go
```diff
@@ -95,3 +124,24 @@ func (m *Module) SendFil(ctx context.Context, from string, to string, amount *bi
 	_, err = m.api.MpoolPushMessage(ctx, msg)
 	return err
 }
+
+// FundFromFaucet make a faucet call to fund the provided wallet address.
+func (m *Module) FundFromFaucet(ctx context.Context, addr string) error {
```
jsign, line 129: If makes sense in the future, we can make a pluggable component that might fund wallets by an external service, so this can be extendable.

asutula:
👀

jsign:
> Looks good. One thing I thought I'd see, but didn't, is some long timeout for a job. If I read this all right, even the longest running broken job/deal would need to be canceled by the user. There's no timeout that cancels a job automatically?
Yep, that's not in this PR. Tracked in #468. 
The good part is that doing the manual one did most of the heavy lifting.

Canceling automatically would be mostly having a timer with that duration, and do the `cancel()` as if done manually. (Maybe with some extra care about adding in the cid-logs that was canceled automatically and not by the user, etc). Also, taking care of setting some safe-default such as 0 meaning will never get auto-canceled, etc.

So, mostly waiting for the need for that feature... if that's the case, it can be done quite fast now.
The manual cancellation is useful to me know that I'm doing some tests in the testnet and I prefer having more fine-grained control on when to cancel things, or exploring waiting times.

PR #476, opened 2020-06-26
improve reputation score, ColdInfo enabled flag, fixes

jsign:
Some general improvements will do some quick comments in review.


ffs/rpc/rpc.go
```diff
@@ -728,8 +728,10 @@ func toRPCCidInfo(info ffs.CidInfo) *CidInfo {
 			},
 		},
 		Cold: &ColdInfo{
+			Enabled: info.Cold.Enabled,
```
jsign, line 731: `ColdInfo` was missing a flag indicating if _ColdStorage_ is enabled (since you can store things in hot storage).
The `HotInfo` already had this flag, so now it's symmetric.

ffs/scheduler/scheduler.go
```diff
@@ -676,6 +679,7 @@ func (s *Scheduler) waitForDeals(ctx context.Context, c cid.Cid, startedProposal
 				lock.Lock()
 				failedDeals = append(failedDeals, dealError)
 				lock.Unlock()
+				return
```
jsign, line 682: Missing return created some empty entries in `okDeals`, fixed.

reputation/reputation.go
```diff
@@ -252,7 +252,7 @@ func calculateScore(addr string, mi miner.IndexSnapshot, si faults.IndexSnapshot
 		askScore = 1
 	}
 
-	score := 50*faultsScore + 20*powerScore + 20*externalScore + 10*askScore
+	score := 50*faultsScore + 20*powerScore + 20*externalScore + 10*askScore + 1000*miner.ActiveDeals
```
jsign: Bias the Reputation Module to also use new information in indexes.
The number of Active Deals for a miner is now a big weighting factor in testnet.

reputation/reputation.go
```diff
@@ -252,7 +252,7 @@ func calculateScore(addr string, mi miner.IndexSnapshot, si faults.IndexSnapshot
 		askScore = 1
 	}
 
-	score := 50*faultsScore + 20*powerScore + 20*externalScore + 10*askScore
+	score := 50*faultsScore + 20*powerScore + 20*externalScore + 10*askScore + 1000*float64(miner.ActiveDeals)
```
jsign, line 255: Now that we have the number of active deals per miner, use it to bias heavily on that in the reputation module. That ratio is somewhat arbitrary as the others, but for testnet looks like the best one we can rely on for miner selection.

ffs/rpc/rpc.go
asutula: Nice, I've noticed this and wished it existed.

asutula: Something in my distant memory says the `ColdInfo.Validate()` can be improved by using this.

ffs/rpc/rpc.proto
```diff
@@ -78,12 +78,13 @@ message FilStorage {
...
    string data_cid = 1;
-   int64 size = 2;
+   uint64 size = 2;
    repeated FilStorage proposals = 3;
 }
 
 message ColdInfo {
-   FilInfo filecoin = 1;
+   bool enabled = 1;
+   FilInfo filecoin = 2;
```
asutula, line 87: Enjoy your breaking rpc changes while you can, days are numbered on things like this ;)

ffs/rpc/rpc.go
asutula: Never mind, was thinking of `ColdConfig`

jsign: Well, _ColdInfo_ doesn't have a _Validate_ method since it's only describing the current state of the storage. I think you're referring to _ColdConfig_, which already had the _Enabled_ flag and use it in `Validate()`. Very similar names. :P

PR #477, opened 2020-06-26
Deals Index

asutula:
This PR completes the first phase (out of 2 or maybe 3) of getting historical data about deals persisted. Most of the harder work is in here, and remaining work is mostly about filtering or mapping the data to some other representation. More on that at the end.

This PR:

- Records started storage deals as "pending"
- Moves storage records from pending to "final" when they are active on chain
- Includes logic to process pending deals on powergate startup
- Records retrievals when they're started (no concept of pending or final here... that is more complex)
- Makes this data available in the Deals module, RPC, Client, and CLI APIs

This data provides the ability to extract different types of information:

- Historical log of storage deals
- Historical log of retrieval deals
- Deals per data CID
- Storage price over time
- Retrieval price over time
- Data amount stored over time
- Data amount retrieved over time
- A higher level "transaction" log abstraction more focused on `from`, `to`, `amount`
- Probably more

Next steps:

- Integrate into FFS to retrieve all this data but filtered to only address managed by the FFS instance
- Possibly provide the "transaction" log abstraction I mentioned above, but that can also just be done by the client. Let's see.


deals/deals.go
```diff
@@ -19,12 +20,14 @@ import (
...
 	"github.com/filecoin-project/specs-actors/actors/abi/big"
 	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/textileio/powergate/util"
 )
 
 const (
 	chanWriteTimeout = time.Second
+	dealTimeout      = time.Hour * 24
```
asutula, line 30: At some point we have to give up monitoring an in process deal. I just chose 24 hours.

deals/deals.go
```diff
@@ -41,22 +44,26 @@ var (
...
-	return &Module{
-		api: api,
-		cfg: &cfg,
-	}, nil
+	m := &Module{
+		api:   api,
+		cfg:   &cfg,
+		store: newStore(ds),
+	}
+	m.initPendingDeals()
```
asutula, line 65: First thing, fetch any previously pending deals. More explanation in that func.

deals/deals.go
```diff
@@ -129,6 +136,7 @@ func (m *Module) Store(ctx context.Context, waddr string, dataCid cid.Cid, piece
 			ProposalCid: *p,
 			Success:     true,
 		}
+		m.recordDeal(params, *p)
```
asutula, line 139: Start the process of recording a deal record, more explanation in that func

deals/deals.go
```diff
@@ -176,9 +184,10 @@ func (m *Module) retrieve(ctx context.Context, waddr string, cid cid.Cid, ref *a
 	}
 	for _, o := range offers {
 		if err = m.api.ClientRetrieve(ctx, o.Order(addr), ref); err != nil {
-			log.Infof("fetching/retrieving cid %s from %s: %s", cid, o.Miner, err)
+			log.Infof("error fetching/retrieving cid %s from %s: %s", cid, o.Miner, err)
 			continue
 		}
+		m.recordRetrieval(waddr, o)
```
asutula, line 190: Record a retrieval

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
 }
 
+// FinalDealRecords returns a list of all finalized storage deals.
+func (m *Module) FinalDealRecords() ([]DealRecord, error) {
+	return m.store.getFinalDeals()
+}
+
+// PendingDealRecords returns a list of all pending storage deals.
+func (m *Module) PendingDealRecords() ([]DealRecord, error) {
+	return m.store.getPendingDeals()
```
asutula, line 244: These come back sorted by time

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+}
+
+// AllDealRecords returns a list of all finalized and pending deals.
+func (m *Module) AllDealRecords() ([]DealRecord, error) {
```
asutula, line 248: Combine pending and final into a single list.

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+		if ret[i].Pending || ret[j].Pending {
+			return ret[i].Time < ret[j].Time
+		}
+		if ret[i].DealInfo.ActivationEpoch < ret[j].DealInfo.ActivationEpoch {
+			return true
+		}
+		if ret[i].DealInfo.ActivationEpoch > ret[j].DealInfo.ActivationEpoch {
+			return false
+		}
+		return ret[i].Time < ret[j].Time
```
asutula, line 268: Sorting logic is, if either is pending, use timestamp. If both are not pending, first sort by activation epoch then timestamp

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+func (m *Module) RetrievalRecords() ([]RetrievalRecord, error) {
+	return m.store.getRetrievals()
+}
+
+// Close closes the deals Module.
+func (m *Module) Close() error {
+	return m.store.close()
+}
+
+func (m *Module) initPendingDeals() {
```
asutula, line 283: This gets calls on deal module creation. Goal here is to return as soon as possible by spawning off all sub routines in go routines. We query for any previously pending deals, calculate the time remaining to continue monitoring each one, then begin monitoring each one. Or, if the pending deal's monitoring period has timed out, we finalize it immediately.

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+	pendingDeals, err := m.store.getPendingDeals()
+	if err != nil {
+		log.Errorf("getting pending deals: %v", err)
+		return
```
asutula, line 287: Oh, a general thing with all this deal record logging feature... I decided not to return any `error`s and instead, just log errors. It seemed like there isn't a good reason to fail data storage if something goes wrong with logging deals. Open to input here for sure.

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+	}
+	for _, dr := range pendingDeals {
+		remaining := time.Until(time.Unix(dr.Time, 0).Add(dealTimeout))
```
asutula, line 290: Calculate monitoring time remaining based on `dealTimeout` and the original timestamp of the pending deal

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+		Time:     time.Now().Unix(),
+		DealInfo: di,
+		Pending:  true,
+	}
+	log.Infof("storing pending deal record for proposal cid: %s", proposalCid.String())
+	if err := m.store.putPendingDeal(record); err != nil {
+		log.Errorf("storing pending deal: %v", err)
+		return
+	}
+	go m.eventuallyFinalizeDeal(record, dealTimeout)
```
asutula, line 318: This is called from `Store`. Persist the pending deal, the spawn off the monitoring process to eventually finalize it.

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+}
+
+func (m *Module) finalizePendingDeal(dr DealRecord) {
```
asutula, line 321: This is called on start up for pending deals that have timed out while powergate was down.

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+	deletePending := func() {
+		if err := m.store.deletePendingDeal(dr.DealInfo.ProposalCid); err != nil {
+			log.Errorf("deleting pending deal for proposal cid %s: %v", dr.DealInfo.ProposalCid.String(), err)
+		}
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+	info, err := m.api.ClientGetDealInfo(ctx, dr.DealInfo.ProposalCid)
```
asutula, line 329: Directly fetch deal info

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+	if err != nil {
+		log.Errorf("getting deal info: %v", err)
+		deletePending()
+		return
```
asutula, line 333: If we can't get the deal info, log and error, delete the pending record and return

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+	}
+	if info.State != storagemarket.StorageDealActive {
+		log.Infof("pending deal for proposal cid %s isn't active yet, deleting pending deal", dr.DealInfo.ProposalCid.String())
+		deletePending()
```
asutula, line 337: If the deal isn't active, the deal has timed out, delete the pending record

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+	} else {
+		di, err := fromLotusDealInfo(ctx, m.api, info)
+		if err != nil {
+			log.Errorf("converting proposal cid %s from lotus deal info: %v", dr.DealInfo.ProposalCid.String(), err)
+			deletePending()
+			return
+		}
+		record := DealRecord{
+			Addr:     dr.Addr,
+			Time:     time.Now().Unix(), // Note: This can be much later in time than the deal actually became active on chain
```
asutula, line 347: Setting the time for completeness, but the active epoch in `di` is more important.

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+			DealInfo: di,
+			Pending:  false,
+		}
+		if err := m.store.putFinalDeal(record); err != nil {
```
asutula, line 351: This also takes care of deleting any pending record for the same proposal cid

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+			log.Errorf("storing propoosal cid %s deal record: %v", dr.DealInfo.ProposalCid.String(), err)
+		}
+	}
+}
+
+func (m *Module) eventuallyFinalizeDeal(dr DealRecord, timeout time.Duration) {
```
asutula, line 357: This is called from indirectly from `Store`, after the pending deal record is persisted. We use the `Watch` api to wait for the deal to become active, then finalize the deal record.

asutula: Can also be called from `initPendingDeals` if there is remaining time to monitor the pending deal

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+	for {
+		select {
+		case <-ctx.Done():
+			log.Infof("watching proposal cid %s timed out, deleting pending deal", dr.DealInfo.ProposalCid.String())
+			if err := m.store.deletePendingDeal(dr.DealInfo.ProposalCid); err != nil {
+				log.Errorf("deleting pending deal: %v", err)
+			}
+			return
+		case info, ok := <-updates:
+			if !ok {
```
asutula, line 374: Not sure if checking `ok` is overkill here since the channel only exists within the local scope, but did it anyway

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+			}
+			if info.StateID == storagemarket.StorageDealActive {
+				record := DealRecord{
+					Addr:     dr.Addr,
+					Time:     time.Now().Unix(),
+					DealInfo: info,
+					Pending:  false,
+				}
+				log.Infof("proposal cid %s is active, storing deal record", info.ProposalCid.String())
+				if err := m.store.putFinalDeal(record); err != nil {
```
asutula, line 389: If we get and active deal, finalize it.

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+					log.Errorf("storing propoosal cid %s deal record: %v", info.ProposalCid.String(), err)
+				}
+				return
+			} else if info.StateID == storagemarket.StorageDealProposalNotFound ||
```
asutula, line 393: If we detect any of these failed final states, delete the pending deal record.

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+				if err := m.store.deletePendingDeal(info.ProposalCid); err != nil {
+					log.Errorf("deleting pending deal: %v", err)
+				}
+				return
+			}
+		}
+	}
+}
+
+func (m *Module) recordRetrieval(addr string, offer api.QueryOffer) {
```
asutula, line 407: Recording retrieval deals is simple for now because we don't have and pending or final state. They are just logged.

asutula: I could imagine some future version where there is a pending or intermediate state, but that would probably rely on underlying payment channels, and all of that isn't clear to me right now. So keeping it simple for now.

deals/deals_test.go
```diff
@@ -42,10 +43,25 @@ func TestStore(t *testing.T) {
...
 			client, _, _ := tests.CreateLocalDevnet(t, nm)
-			m, err := New(client, WithImportPath(filepath.Join(tmpDir, "imports")))
+			m, err := New(tests.NewTxMapDatastore(), client, WithImportPath(filepath.Join(tmpDir, "imports")))
 			checkErr(t, err)
-			_, err = storeMultiMiner(m, client, nm, randomBytes(600))
+			_, pcids, err := storeMultiMiner(m, client, nm, randomBytes(600))
 			checkErr(t, err)
+			pending, err := m.PendingDealRecords()
+			require.Nil(t, err)
+			require.Len(t, pending, nm)
```
asutula, line 52: Inserted checks for the deal records into the storage test. Required refactoring `storeMultiMiner` and `waitForDealComplete` to be separate calls so we can check things in between

deals/deals_test.go
```diff
@@ -42,10 +43,25 @@ func TestStore(t *testing.T) {
...
+			final, err := m.FinalDealRecords()
+			require.Nil(t, err)
+			require.Empty(t, final)
+			err = waitForDealComplete(client, pcids)
+			checkErr(t, err)
+			time.Sleep(util.AvgBlockTime)
```
asutula, line 58: Not sure how to avoid this sleep. It exists because our deal monitoring uses `Watch` which polls for the deal state every `util.AvgBlockTime`. So we need to wait at least that long here to know that we've detected the deal state in `Watch`. Unfortunate for devnet because `util.AvgBlockTime` is much longer than needed since it is the average for test net or main net. Ideas?

deals/rpc/rpc.proto
```diff
@@ -62,8 +85,40 @@ message RetrieveResponse {
...
+}
+
 service RPCService {
     rpc Store(stream StoreRequest) returns (StoreResponse) {}
     rpc Watch(WatchRequest) returns (stream WatchResponse) {}
     rpc Retrieve(RetrieveRequest) returns (stream RetrieveResponse) {}
+    rpc FinalDealRecords(FinalDealRecordsRequest) returns (FinalDealRecordsResponse) {}
+    rpc PendingDealRecords(PendingDealRecordsRequest) returns (PendingDealRecordsResponse) {}
+    rpc AllDealRecords(AllDealRecordsRequest) returns (AllDealRecordsResponse) {}
+    rpc RetrievalRecords(RetrievalRecordsRequest) returns (RetrievalRecordsResponse) {}
```
asutula, line 123: Here's a good summary of the API changes.

deals/store.go
```diff
@@ -0,0 +1,195 @@
...
+	dsBaseRetrieval      = datastore.NewKey("retrieval")
+
+	// ErrNotFound indicates the instance doesn't exist.
+	ErrNotFound = errors.New("cid info not found")
+)
+
+type store struct {
+	ds   datastore.Datastore
+	lock sync.Mutex
+}
```
asutula, line 28: Persistence layer for storing deal records.

deals/store.go
```diff
@@ -0,0 +1,195 @@
...
+		ret = append(ret, dr)
+	}
+	sort.Slice(ret, func(i, j int) bool { return ret[i].Time < ret[j].Time })
+	return ret, nil
+}
+
+func (s *store) putFinalDeal(dr DealRecord) error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	err := s.ds.Delete(makePendingDealKey(dr.DealInfo.ProposalCid))
```
asutula, line 87: Importantly, also deletes any pending deal record for the same proposal cid

deals/store.go
```diff
@@ -0,0 +1,195 @@
...
+	return dsBaseStoragePending.ChildString(c.String())
+}
+
+func makeFinalDealKey(c cid.Cid) datastore.Key {
+	return dsBaseStorageFinal.ChildString(c.String())
+}
+
+func makeRetrievalKey(rr RetrievalRecord) datastore.Key {
+	str := fmt.Sprintf("%v%v%v%v", rr.Time, rr.Addr, rr.RetrievalInfo.Miner, rr.RetrievalInfo.PieceCID.String())
+	sum := md5.Sum([]byte(str))
```
asutula, line 193: needed a unique key so just built one like this, no idea if there is a better way

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+		log.Errorf("getting pending deals: %v", err)
+		return
+	}
+	for _, dr := range pendingDeals {
+		remaining := time.Until(time.Unix(dr.Time, 0).Add(dealTimeout))
+		if remaining <= 0 {
+			go m.finalizePendingDeal(dr)
+		} else {
+			go m.eventuallyFinalizeDeal(dr, remaining)
+		}
```
jsign, line 295: On a `Close()` call, these goroutines won't we waited. 
I think that's OK since things will spin up again re-evaluating all the situation again. 
Mentioning this just in case I may be missing something important to consider.

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+			return
+		}
+		record := DealRecord{
+			Addr:     dr.Addr,
+			Time:     time.Now().Unix(), // Note: This can be much later in time than the deal actually became active on chain
+			DealInfo: di,
+			Pending:  false,
+		}
+		if err := m.store.putFinalDeal(record); err != nil {
+			log.Errorf("storing propoosal cid %s deal record: %v", dr.DealInfo.ProposalCid.String(), err)
```
jsign, line 352: Nit: typo

deals/rpc/rpc.go
```diff
@@ -192,3 +192,81 @@ func (s *RPC) Retrieve(req *RetrieveRequest, srv RPCService_RetrieveServer) erro
...
+
+func toRPCDealRecords(records []deals.DealRecord) []*DealRecord {
+	ret := make([]*DealRecord, len(records))
+	for i, r := range records {
+		ret[i] = &DealRecord{
+			Addr:    r.Addr,
+			Time:    r.Time,
+			Pending: r.Pending,
+			DealInfo: &DealInfo{
+				ProposalCid:     r.DealInfo.ProposalCid.String(),
```
jsign, line 256: Q: Should `r.DealInfo` be nil checked?

deals/rpc/rpc.go
```diff
@@ -192,3 +192,81 @@ func (s *RPC) Retrieve(req *RetrieveRequest, srv RPCService_RetrieveServer) erro
...
+	if err != nil {
+		return nil, err
+	}
+	ret := make([]*RetrievalRecord, len(records))
+	for i, r := range records {
+		ret[i] = &RetrievalRecord{
+			Addr: r.Addr,
+			Time: r.Time,
+			RetrievalInfo: &RetrievalInfo{
+				PieceCid:                r.RetrievalInfo.PieceCID.String(),
```
jsign, line 235: Q: Should `r.RetrievalInfo` be nil checked?

deals/deals.go
jsign: Yeah, I think is is a safe reasonable value.

deals/deals_test.go
jsign: For tests, in some `TestMain`, you could do `util.AvgBlockTime=smaller_value` so things happen more often.
I do this in `integration_test.go` in `TestMain()`. Would that help?

deals/store.go
```diff
@@ -0,0 +1,195 @@
...
+	})
+	return ret, nil
+}
+
+func (s *store) close() error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	err := s.ds.Close()
+	return err
+}
```
jsign, line 181: Important: This `close` method should be deleted. The underlying datastore will be closed by main, since what this store is seeing is just a slice of the namespace. If this was a datastore created and dedicated to this datastore, then  this would be OK.

deals/store.go
jsign: Nice, was thinking about that in the `deals.go` logic.

deals/store.go
jsign: I guess should be fine. But maybe all this can be avoided by generating a `uuid` and using that as the key?
Since this key isn't that useful anyway to do a Query or similar.

deals/deals_test.go
asutula: Ah yea, that would work.

deals/store.go
asutula: Actually, I think my next PR will rethink the keys used to make it easier to query by address etc. So I'll leave this for now knowing that it is going to change anyway.

deals/deals.go
asutula: Done.

api/client/deals.go
```diff
@@ -178,3 +178,111 @@ func (d *Deals) Retrieve(ctx context.Context, waddr string, cid cid.Cid) (io.Rea
...
+				PieceCID:                pieceCid,
+				Size:                    rpcRecord.RetrievalInfo.Size,
+				MinPrice:                rpcRecord.RetrievalInfo.MinPrice,
+				PaymentInterval:         rpcRecord.RetrievalInfo.PaymentInterval,
+				PaymentIntervalIncrease: rpcRecord.RetrievalInfo.PaymentIntervalIncrease,
+				Miner:                   rpcRecord.RetrievalInfo.Miner,
+				MinerPeerID:             rpcRecord.RetrievalInfo.MinerPeerId,
+			}
+		}
+		ret[i] = record
```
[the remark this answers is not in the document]
asutula: Good catch. It should never be nil, but if it is for some reason, it makes no sense to include that record. Fixed.

deals/deals.go
asutula: Yea, that is fine because `Close()` could have to wait a loooong time. Correct, the state of things will get picked up where it left off.

deals/rpc/rpc.go
asutula: I think it's fine since there are no pointers involved, just structs which will have default empty values.

deals/rpc/rpc.go
asutula: Same as above

deals/store.go
asutula: Ah good call.

deals/deals.go
```diff
@@ -225,6 +234,195 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan DealInf
...
+// PendingDealRecords returns a list of all pending storage deals.
+func (m *Module) PendingDealRecords() ([]DealRecord, error) {
+	return m.store.getPendingDeals()
+}
+
+// AllDealRecords returns a list of all finalized and pending deals.
+func (m *Module) AllDealRecords() ([]DealRecord, error) {
+	ret, err := m.store.getFinalDeals()
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
asutula: Ok makes sense, updated.

api/client/deals.go
```diff
@@ -178,3 +178,111 @@ func (d *Deals) Retrieve(ctx context.Context, waddr string, cid cid.Cid) (io.Rea
...
+	res, err := d.client.AllDealRecords(ctx, &rpc.AllDealRecordsRequest{})
+	if err != nil {
+		return nil, err
+	}
+	ret, err := fromRPCDealRecords(res.Records)
+	if err != nil {
+		return nil, err
+	}
+	return ret, nil
+}
```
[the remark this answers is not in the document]
asutula: I like this idea. I think that integrating this into the FFS will get me thinking more closely about the right api, so I'll leave this for now and improve it in that next PR

deals/rpc/rpc.go
asutula: I mean below.

deals/deals.go
jsign: Forgot to mention here. Agree!

deals/store.go
jsign: Cool 👍

PR #493, opened 2020-07-01
ffs/scheduler: resume executing re-attached Jobs concurrently & update deps

### LINKED ISSUE #492: ffs/scheduler: resume Jobs concurrently
If Powergate is shutdown and multiple Jobs are still executing they'll be resumed on the next start.

Currently, the resumed Jobs will be resumed and waited to finish before executing new queued Jobs. That's a bit annoying since resumed deals _could_ take a long time to finish if, for example, still are in the early phase of sealing in miners.

We should resume them as if they were _just queued_ and allow taking more Jobs within the rate limiter. Saying it differently resumed deals shouldn't have a priority over execution than new stuff coming. In some sense, is the most natural behavior too.

jsign:
- Closes #492
- Updates deps mentioned by dependabot. Skipped ones are related to using the same version as what Lotus is using.
- Addition in `.gitignore`
- Fix `lotus/client` logger name.


ffs/scheduler/scheduler.go
```diff
@@ -272,15 +267,13 @@ func (s *Scheduler) resumeStartedDeals() error {
...
-		wg.Add(1)
 		go func(j ffs.Job) {
-			defer wg.Done()
+			log.Infof("resuming job %s with cid %s", j.ID, j.Cid)
 			// We re-execute the pipeline as if was dequeued.
 			// Both hot and cold storage can detect resumed job execution.
 			s.executeQueuedJob(j)
 		}(j)
 	}
-	wg.Wait()
```
jsign, line 283: Resume deals and don't wait for them, keep evaluating new Queued Jobs. (Just as if they were pushed in the usual case)

ffs/scheduler/scheduler.go
```diff
@@ -422,6 +408,17 @@ func (s *Scheduler) executeQueuedJob(j ffs.Job) {
...
+	// Get
+	a, err := s.as.Get(j.ID)
+	if err != nil {
+		log.Errorf("getting push config action data from store: %s", err)
+		if err := s.js.Finalize(j.ID, ffs.Failed, err, nil); err != nil {
+			log.Errorf("changing job to failed: %s", err)
+		}
+		s.l.Log(ctx, a.Cfg.Cid, "Job %s couldn't start: %s.", j.ID, err)
+		return
+	}
```
jsign, line 420: Moved from above, plus finalizing the job as a failure, since we can't do anything (isn't a retry issue).

ffs/scheduler/scheduler.go
```diff
@@ -466,6 +463,12 @@ func (s *Scheduler) execute(ctx context.Context, a astore.Action, job ffs.Job) (
 		return ffs.CidInfo{}, nil, fmt.Errorf("getting current cid info from store: %s", err)
 	}
 
+	if a.ReplacedCid.Defined() {
+		if err := s.Untrack(a.ReplacedCid); err != nil && err != astore.ErrNotFound {
+			return ffs.CidInfo{}, nil, fmt.Errorf("untracking replaced cid: %s", err)
+		}
+	}
```
jsign, line 470: This was moved from `push()` to here (when the Job really executes). This is safer since untracking the replaced Cid should only happen whenever the Job starts executing (if it cancels for some reason, it would be wrong to untrack the replaced cid in advance).

jsign:
Thanks!

PR #494, opened 2020-07-01
Better proto config

asutula:
Setting up proto code generation with the correct options and `go_package` values. The incorrect values before would result in incorrect import paths in generated code if you imported one .proto file into another .proto file to use messages defined in the first; The generated go import paths to the other generated file would be wrong. This fixes that problem and will allow us to import/reuse message definitions between .proto files. This is useful in situations where, say `ffs`, exports some api that uses the same types as `deals`.


Makefile
```diff
@@ -34,5 +34,5 @@ install-protoc:
 	
 PROTOCGENGO=$(shell pwd)/buildtools/protoc-gen-go
 protos: install-protoc clean-protos
-	PATH=$(PROTOCGENGO):$(PATH) ./scripts/protoc_gen_plugin.bash --proto_path=. --plugin_name=go --plugin_out=. --plugin_opt=plugins=grpc
```
asutula, line 37: This `paths=source_relative` option along with correctly specifying the full go package in the .proto files is the key

deals/rpc/rpc.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
 package deals.rpc;
 
-option go_package = "deals/rpc";
+option go_package = "github.com/textileio/powergate/deals/rpc";
```
asutula, line 4: This is the correct way to specify `go_package`

Makefile
```diff
@@ -34,5 +34,5 @@ install-protoc:
...
+	PATH=$(PROTOCGENGO):$(PATH) ./scripts/protoc_gen_plugin.bash --proto_path=. --plugin_name=go --plugin_out=. --plugin_opt=plugins=grpc,paths=source_relative
```
asutula, line 37: This `paths=source_relative` option along with correctly specifying the full go package in the .proto files is the secret

PR #495, opened 2020-07-02
feat(builds): adds pow build for linux/darwin to releases

andrewxhill:
Now that we have release builds anyway, might as well drop the CLI. Realized this in developing the tutorial, that saving the CLI build step can be a nice improvement for teaching. 

@asutula i know you had been thinking about some things here, so hopefully first pass version is okay. ideally we could just have one release out tomorrow for the workshop with builds.


asutula:
This is great!

asutula:
What I was working on was more about embedding the release version into the binaries and log output, so this is all necessary work even with that.

PR #497, opened 2020-07-02
Untangle gRPC client from Lotus API deps & rename devnet to localnet

jsign:
This refactor untangles some dependency paths in imports which leads to the Lotus API. The Lotus API interface ends up using types of other deeper repos which have some constraints about compilation in different OSs.
This means that any binary using the Powergate gRPC client can run on windows.

This PR also changes `devnet` to `localnet` in docker-compose files and `make` commands.


.github/workflows/test.yml
```diff
@@ -13,5 +13,7 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v1
+      - name: Compile CLI on Windows
+        run: GOOS=windows go build ./cmd/pow
```
jsign, line 17: Check in CI that compiles on Windows.

PR #499, opened 2020-07-02
Remove config file usage, add wallet list cmd

### LINKED ISSUE #496: Config File ".powergate" Not Found
```
▶ pow wallet new help  
> Success! Wallet address: t3vv5crhofq3wc2hx2spkwkou57nylfbqwdcegwr3amwhvkjyvufetdh5cyhprkispeibagt5fglh73dbg4wga
> Error! Config File ".powergate" Not Found in "[/Users/andrewhill]"
```

I'm getting this when building the binary and then manually moving to bin path. I built it when testing the new CI, so using the gox tool. I'm running devnet with the currently released docker image in Releases... which i see is 9 days old so perhaps is part of this.

asutula:
The config file just complicates things and adds confusion. It was keeping track of wallet addresses, so without that, needed a way for the user to list all wallet addresses... added a `pow wallet list` command with the supporting module func, rpc, and client code.

closes #496


asutula:
Ugh, most file changes are the md cli docs. Every single one is updated sine I removed the global `config` flag :/

PR #503, opened 2020-07-03
Remove unused HotStorage API & use ipfs 0.6.0 & improvements

jsign:
- Remove `Put` API from _HotStorage_ since this was used when Powergate did CAR encodings, which doesn't anymore since we leverage Lotus connecting to IPFS. 
- Changed implementation logic to fill pinset cache once.
- Use IPFS 0.6.0

### DIFF
--- a/ffs/coreipfs/coreipfs.go
+++ b/ffs/coreipfs/coreipfs.go
@@ -1,13 +1,11 @@
 package coreipfs
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"io"
 	"sync"
 
-	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	ipfsfiles "github.com/ipfs/go-ipfs-files"
 	logging "github.com/ipfs/go-log/v2"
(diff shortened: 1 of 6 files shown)


gateway/gateway.go
```diff
@@ -23,7 +23,7 @@ import (
 	"github.com/textileio/powergate/reputation"
 )
 
-const numTopMiners = 30
+const numTopMiners = 100
```
jsign, line 26: Bump it a bit more, so arbitrary. Eventually, we can support (bounded) query-parameters, etc.

ffs/interfaces.go
```diff
@@ -49,9 +48,6 @@ type HotStorage interface {
 	// thought for mutating data doing this efficiently.
 	Replace(context.Context, cid.Cid, cid.Cid) (int, error)
 
-	// Put adds a raw block.
-	Put(context.Context, blocks.Block) error
-
```
jsign, line 54: Unnecessary since we connect Lotus with IPFS.

ffs/coreipfs/coreipfs.go
```diff
@@ -1,13 +1,12 @@
 package coreipfs
```
jsign, line 1: Changed opinion here. Decided to pre-fill pinset cache on `New`. This saves work while calling hot APIs.

PR #508, opened 2020-07-07
Better list deals API

### LINKED ISSUE #507: Cli pow deals store duration error
Comment from support:

> Got another one for you guys. pow deals store takes a duration unit through the -d flag. However no matter what I put in there I get Error! Deal command minDuration should be > 0

asutula:
Idea here is to consolidate deal record querying as much as possible and get set up to provide API that FFS needs.

- Remove extra functions, just one func for querying storage deals and one for retrieval deals
- Provides options for including pending deals, sorting, address and data cid filters
- Updates to this new API from the deals module to the rpc layer to the client to the cli

Snuck a quick fix in that closes #507


api/client/deals.go
```diff
@@ -20,16 +21,58 @@ type Deals struct {
 
 // WatchEvent is used to send data or error values for Watch.
 type WatchEvent struct {
-	Deal deals.DealInfo
+	Deal deals.StorageDealInfo
 	Err  error
 }
 
+// ListDealRecordsOption updates a ListDealRecordsConfig.
+type ListDealRecordsOption func(*rpc.ListDealRecordsConfig)
```
asutula, line 29: Client-only options definitions that modify a request object

api/client/deals.go
```diff
@@ -20,16 +21,58 @@ type Deals struct {
...
+	}
+}
+
 // Store creates a proposal deal for data using wallet addr to all miners indicated
 // by dealConfigs for duration epochs.
 func (d *Deals) Store(ctx context.Context, addr string, data io.Reader, dealConfigs []deals.StorageDealConfig, minDuration uint64) ([]cid.Cid, []deals.StorageDealConfig, error) {
 	stream, err := d.client.Store(ctx)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("calling Store: %v", err)
```
asutula, line 75: Trying to start prepending public api errors when I come across them. Nice practice.

cmd/pow/cmd/deals_retrievals.go
```diff
@@ -0,0 +1,70 @@
...
+		err := viper.BindPFlags(cmd.Flags())
+		checkErr(err)
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx, cancel := context.WithTimeout(context.Background(), cmdTimeout)
+		defer cancel()
+
+		var opts []client.ListDealRecordsOption
+
+		if viper.IsSet("ascending") {
```
asutula, line 38: Nice viper method to check if a flag was actually set. Otherwise, we don't specify the option and go with whatever default exists on the back end

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
 }
 
-// FinalDealRecords returns a list of all finalized storage deals.
-// Records are sorted ascending by activation epoch then timestamp.
-func (m *Module) FinalDealRecords() ([]deals.DealRecord, error) {
-	ret, err := m.store.getFinalDeals()
-	if err != nil {
-		return nil, fmt.Errorf("getting final deals: %v", err)
+// ListStorageDealRecords lists storage deals according to the provided options.
+func (m *Module) ListStorageDealRecords(opts ...deals.ListDealRecordsOption) ([]deals.StorageDealRecord, error) {
```
asutula, line 239: Here are the new apis to query for deal records. Behavior is specified with the options params.

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
+	c := deals.ListDealRecordsConfig{}
+	for _, opt := range opts {
+		opt(&c)
+	}
```
asutula, line 243: Build config from options.

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
+	var final []deals.StorageDealRecord
+	if !c.OnlyPending {
```
asutula, line 245: In this case, we do want non-pending records.

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
-}
 
-// PendingDealRecords returns a list of all pending storage deals.
-// Records are sorted ascending by timestamp.
-func (m *Module) PendingDealRecords() ([]deals.DealRecord, error) {
-	ret, err := m.store.getPendingDeals()
-	if err != nil {
-		return nil, fmt.Errorf("getting pending deals: %v", err)
+	var pending []deals.StorageDealRecord
+	if c.IncludePending || c.OnlyPending {
```
asutula, line 254: In these cases, we need to add in pending records.

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
-// Records are sorted ascending by activation epoch, if available, then timestamp.
-func (m *Module) AllDealRecords() ([]deals.DealRecord, error) {
-	ret, err := m.store.getFinalDeals()
-	if err != nil {
-		return nil, fmt.Errorf("getting final deals: %v", err)
-	}
-	pending, err := m.store.getPendingDeals()
-	if err != nil {
-		return nil, fmt.Errorf("getting pending deals: %v", err)
+	combined := append(final, pending...)
```
asutula, line 262: Combine them

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
+
+	var filtered []deals.StorageDealRecord
+
+	if len(c.FromAddrs) > 0 || len(c.DataCids) > 0 {
```
asutula, line 266: I check that at least one of the filters has length. We're going to iterate through the returned records, so best to only do that once and handle both filters in the same iteration.

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
+		var fromAddrsFilter map[string]struct{}
+		var dataCidsFilter map[string]struct{}
+		for _, addr := range c.FromAddrs {
+			fromAddrsFilter[addr] = struct{}{}
+		}
+		for _, cid := range c.DataCids {
+			dataCidsFilter[cid] = struct{}{}
+		}
```
asutula, line 274: Above, build up filter maps we can use to quickly check values (vs iterating every time)

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
+		for _, record := range combined {
```
asutula, line 275: Iterate over the records and compare with the filter maps

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
+			_, inFromAddrsFilter := fromAddrsFilter[record.Addr]
+			_, inDataCidsFilter := dataCidsFilter[record.DealInfo.PieceCID.String()]
```
asutula, line 277: Above, just check if the address and data cid exist in the filters or not

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
+			includeViaFromAddrs := len(c.FromAddrs) == 0 || inFromAddrsFilter
+			includeViaDataCids := len(dataCidsFilter) == 0 || inDataCidsFilter
```
asutula, line 279: Then decide if each filter contributes to the record being included or not. The answer is yes if the user didn't provide any filter values OR if they did and there's a match.

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
+			if includeViaFromAddrs && includeViaDataCids {
```
asutula, line 280: We only include the record in the result set if both filters contribute `true`

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
-		if ret[i].Pending || ret[j].Pending {
-			return ret[i].Time < ret[j].Time
+
+	sort.Slice(filtered, func(i, j int) bool {
+		l := filtered[j]
+		r := filtered[i]
+		if c.Ascending {
+			l = filtered[i]
+			r = filtered[j]
+		}
```
asutula, line 294: Switch left and right around depending on configuration sort order.

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
-		if ret[i].DealInfo.ActivationEpoch < ret[j].DealInfo.ActivationEpoch {
+		if l.DealInfo.ActivationEpoch < r.DealInfo.ActivationEpoch {
 			return true
 		}
-		if ret[i].DealInfo.ActivationEpoch > ret[j].DealInfo.ActivationEpoch {
+		if l.DealInfo.ActivationEpoch > r.DealInfo.ActivationEpoch {
 			return false
 		}
-		return ret[i].Time < ret[j].Time
+		return l.Time < r.Time
```
asutula, line 304: Sort by timestamp if either record is pending, activation epoch otherwise, use timestamp as secondary sort when activation epochs are the same.

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
+
+	return filtered, nil
 }
 
-// RetrievalRecords returns a list of all retrievals.
-// Records are sorted ascending by timestamp.
-func (m *Module) RetrievalRecords() ([]deals.RetrievalRecord, error) {
+// ListRetrievalDealRecords returns a list of retrieval deals
+// according to the provided options.
+func (m *Module) ListRetrievalDealRecords(opts ...deals.ListDealRecordsOption) ([]deals.RetrievalDealRecord, error) {
```
asutula, line 312: Similar logic in here are above

deals/module/store.go
```diff
@@ -113,27 +111,18 @@ func (s *store) getFinalDeals() ([]deals.DealRecord, error) {
...
 	}
-	sort.Slice(ret, func(i, j int) bool {
-		if ret[i].DealInfo.ActivationEpoch < ret[j].DealInfo.ActivationEpoch {
-			return true
-		}
-		if ret[i].DealInfo.ActivationEpoch > ret[j].DealInfo.ActivationEpoch {
-			return false
-		}
-		return ret[i].Time < ret[j].Time
-	})
```
asutula, line 132: Sorting is done in the module implementation now depending on user config

deals/module/store_test.go
```diff
@@ -36,33 +36,30 @@ func TestGetPendingDeals(t *testing.T) {
...
-	err = s.putPendingDeal(deals.DealRecord{Addr: "c", Time: now + 3, Pending: true, DealInfo: deals.DealInfo{ProposalCid: c3}})
+	err = s.putPendingDeal(deals.StorageDealRecord{Addr: "c", Time: now + 3, Pending: true, DealInfo: deals.StorageDealInfo{ProposalCid: c3}})
 	require.Nil(t, err)
 
 	res, err := s.getPendingDeals()
 	require.Nil(t, err)
 	require.Len(t, res, 3)
-	require.Equal(t, res[0].Addr, "a")
-	require.Equal(t, res[1].Addr, "b")
-	require.Equal(t, res[2].Addr, "c")
```
asutula, line 57: The store no longer sorts data, can' test for that

deals/options.go
```diff
@@ -0,0 +1,75 @@
+package deals
+
+import "os"
+
+// Config contains configuration for storing deals.
+type Config struct {
```
asutula, line 6: Moved this from `types.go` to here

deals/options.go
```diff
@@ -0,0 +1,75 @@
...
+		if err := os.MkdirAll(path, 0700); err != nil {
+			return err
+		}
+		c.ImportPath = path
+		return nil
+	}
+}
+
+// ListDealRecordsConfig specifies the options for DealsManager.List.
+type ListDealRecordsConfig struct {
```
asutula, line 26: Config and options for listing storage deals

asutula: Both storage and retrieval record funcs use the same options. A couple options don't apply to retrieval deals, and that is noted in the code docs

deals/types.go
```diff
@@ -20,8 +18,8 @@ type StoreResult struct {
 	Message     string
 }
 
-// DealInfo contains information about a proposed storage deal.
-type DealInfo struct {
+// StorageDealInfo contains information about a proposed storage deal.
+type StorageDealInfo struct {
```
asutula, line 22: Making some names in here more succinct now that retrieval deal info and records are a thing

api/client/deals.go
```diff
@@ -20,16 +21,58 @@ type Deals struct {
...
+
+// WithAscending specifies to sort the results in ascending order.
+// Default is descending order.
+// If pending, records are sorted by timestamp, otherwise records
+// are sorted by activation epoch then timestamp.
+func WithAscending(ascending bool) ListDealRecordsOption {
+	return func(c *rpc.ListDealRecordsConfig) {
+		c.Ascending = ascending
+	}
+}
```
jsign, line 68: Q: May worth clarifying what happens if in the list you have both _pending_ and another status? (e.g: pending first, and then the rest)

cmd/pow/cmd/deals_store.go
```diff
@@ -36,7 +36,7 @@ var storeCmd = &cobra.Command{
 		defer cancel()
 
 		addr := viper.GetString("address")
-		minDuration := viper.GetUint64("minDuration")
+		minDuration := viper.GetUint64("duration")
```
jsign, line 39: 👍

deals/module/deals.go
jsign: Q: Maybe clearer to change `OnlyPending` to `IncludeFinal`? So both flags have "Include".
Like, this valid combination `c.IncludePending=false` and `c.OnlyPending=true` is worth considering? (that condition wouldn't match above if, and this if)

deals/module/deals.go
```diff
@@ -235,61 +235,125 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.D
...
+	sort.Slice(filtered, func(i, j int) bool {
+		l := filtered[j]
+		r := filtered[i]
+		if c.Ascending {
+			l = filtered[i]
+			r = filtered[j]
+		}
+		if l.Pending || r.Pending {
+			return l.Time < r.Time
 		}
```
jsign, line 297: I think there's a possible problem here with the sorting condition.

A pending, B final, C final. B.time<A.time<C.time, and B.Activation>C.Activation.
I think you can get different final sortition depending on the order in which the underlying algorithm sorts things.

deals/module/deals.go
asutula: Tough call. My hope was that if you didn't provide any options at all, the default would be some sensible results, in this case, only final deals sorted descending. If we make it `IncludeFinal`, there could still be sensible defaults, but only if the default value for `IncludeFinal` is `true`... That could work but that sort of name + default value combination seems unintuitive to me. Default values of `false` usually are easier to understand.

asutula: I think the current way it works actually does match the second `if`, but I agree it's confusing that there are two options that seemingly contradict each other. The current rule is that the `OnlyPending` wins out, but that would have to be explained.

asutula: So I'm thinking the less confusing of the options is what you're suggesting: `IncludeFinal` with a default of `true`. I'll give it a try

api/client/deals.go
asutula: Ha, it's probably worth writing a test for this because my intention was actually that the pending and final records are interleaved. I'll break the sorting logic into it's own method and write a test for it.

deals/module/deals.go
asutula: Hmm, trying to think through this. Can you think of how this could be fixed? I guess I was (probably incorrectly) assuming the underlying sorting algorithm would continue comparing pairs of items until things are settled completely. But that was just an assumption.

asutula: Will be sure to cover your case in a test.

jsign: The algorithm used by sort depends on the number of elements of the slice.
The important thing with the sorted condition is that it should be transitive, so A<B and B<C should always imply A<C. If that isn't the case, depending on the original slice circumstances, you can get different results.

Maybe just placing Pending first (with Time to subsort), and Finalized afterwards (with activation epoch) should be reasonable. Both deals are quite different and are trying to use unrelated temporal dimensions, so its quite hard to reconcile both at the same time.

jsign: Or maybe always by `Time`.  I'd let the application decide a more complex sortition criteria depending on what is trying to do?

asutula: Good call, keep it simple.

cli-docs/pow/pow_deals_storage.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+### Options
+
+```
+      --addrs strings     limit the records to deals initiated from  the specified wallet addresses, treated as and AND operation if --cids is also provided
+  -a, --ascending         sort records ascending, default is sort descending
+      --cids strings      limit the records to deals for the specified data cids, treated as and AND operation if --addrs is also provided
+  -h, --help              help for storage
+  -f, --include-final     include final deals (default true)
```
asutula, line 20: Cool that cobra detects the strange true default and notes it here. (that is automated, the only command doc in our code is `include final deals`)

asutula:
Ok in the end I decided to just make `IncludeFinal` default to `false`. So now there are no strange defaults, but you do have to supply one or the other or both of `WithIncludeFinal` or `WithIncludePending` to get any results. Slightly harder to use API to avoid all confusion. Seems like an ok trade off.

PR #509, opened 2020-07-07
FFS deal records API

asutula:
This PR creates a deal records API in FFS very similar to the API in the deals module, but assures that returned deal records are for only addresses managed by the ffs instance. Code is reused as much as possible from the deals module.


deals/module/deals.go
```diff
@@ -264,8 +264,8 @@ func (m *Module) ListStorageDealRecords(opts ...deals.ListDealRecordsOption) ([]
 	var filtered []deals.StorageDealRecord
 
 	if len(c.FromAddrs) > 0 || len(c.DataCids) > 0 {
-		var fromAddrsFilter map[string]struct{}
-		var dataCidsFilter map[string]struct{}
+		fromAddrsFilter := make(map[string]struct{})
+		dataCidsFilter := make(map[string]struct{})
```
asutula, line 268: nil map bug fix

deals/rpc/rpc.go
```diff
@@ -238,7 +239,8 @@ func buildListDealRecordsOptions(conf *ListDealRecordsConfig) []deals.ListDealRe
 	return opts
 }
 
-func toRPCStorageDealRecords(records []deals.StorageDealRecord) []*StorageDealRecord {
+// ToRPCStorageDealRecords converts from Go to gRPC representations.
+func ToRPCStorageDealRecords(records []deals.StorageDealRecord) []*StorageDealRecord {
```
asutula, line 243: Export this so it can be used by ffs version

deals/types.go
```diff
@@ -47,7 +47,7 @@ type StorageDealRecord struct {
 
 // RetrievalDealInfo contains information about a retrieval deal.
 type RetrievalDealInfo struct {
-	PieceCID                cid.Cid
+	RootCid                 cid.Cid
```
asutula, line 50: No that I understand `PieceCID` is not the same as a data cid, this is the correct name

ffs/api/api_deals.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/textileio/powergate/deals"
+)
+
+// ListStorageDealRecords lists storage deals for this FFS instance according to the provided options.
+func (i *API) ListStorageDealRecords(opts ...deals.ListDealRecordsOption) ([]deals.StorageDealRecord, error) {
+	c := deals.ListDealRecordsConfig{}
+	for _, opt := range opts {
+		opt(&c)
+	}
+	finalAddrs, err := i.finalAddresses(c.FromAddrs)
```
asutula, line 15: You can still provide a list of addresses that you want to query for, but the logic in `finalAddressees` will check that the addresses you specify are managed by the ffs instance. It returns an error if it finds an address not managed by the ffs instance

ffs/integrationtest/integration_test.go
```diff
@@ -1468,6 +1472,22 @@ func requireCidConfig(t *testing.T, fapi *api.API, c cid.Cid, config *ffs.CidCon
 	require.Equal(t, *config, currentConfig)
 }
 
+func requireStorageDealRecord(t *testing.T, fapi *api.API, _ cid.Cid) {
+	time.Sleep(time.Second)
+	recs, err := fapi.ListStorageDealRecords(deals.WithIncludeFinal(true))
+	require.Nil(t, err)
+	require.Len(t, recs, 1)
+	// ToDo: this should be PayloadCid once we can get it
+	// require.Equal(t, c, recs[0].DealInfo.PieceCID)
```
asutula, line 1481: Need to sort out the data cid thing. For now, just check that a record is returned.

ffs/interfaces.go
```diff
@@ -28,6 +29,12 @@ type PaychManager interface {
 	Redeem(ctx context.Context, ch string) error
 }
 
+// DealRecordsManager provides access to deal records.
+type DealRecordsManager interface {
+	ListStorageDealRecords(opts ...deals.ListDealRecordsOption) ([]deals.StorageDealRecord, error)
+	ListRetrievalDealRecords(opts ...deals.ListDealRecordsOption) ([]deals.RetrievalDealRecord, error)
+}
```
asutula, line 36: Needed to use an interface or else the windows build breaks

ffs/rpc/rpc.proto
```diff
@@ -3,6 +3,8 @@ package ffs.rpc;
 
 option go_package = "github.com/textileio/powergate/ffs/rpc";
 
+import "deals/rpc/rpc.proto";
```
asutula, line 6: Bringing some work from a previous PR full circle. Can now import our other proto files and share message types. The generated go code imports now work properly

ffs/integrationtest/integration_test.go
jsign: Can't this be enabled checking RootCid?

asutula: `StorageDealInfo` has no `RootCid`.

asutula: Only `PieceCID` and `PproposalCid`

jsign: Sorry, I meant that `Store` could use the `dataCid` value in the `recordDeal` to keep track of that.
Is that what you're planning?

asutula: Yea, something like that should work. Actually, I'll just try to get that in now.

asutula: Ok done. I'm storing `RootCid` on `StorageDealRecord` so that information is available when someone queries for storage deal records. Other apis that give deal status like `Watch` still just use `StoreDealInfo` which is derived directly from a Lotus `DealInfo` and doesn't not include root cid. That would be a good improvement in Lotus to include that information.

PR #519, opened 2020-07-12
Various updates

### LINKED ISSUE #516: Make GeoLite2-City.mmdb path configurable
It's currently hard coded. Should be a flag/env var.

asutula:
Various things I found while integrating latest APIs into the js client:

- Updated naming and APIs for wallet RPC, client and CLI
- Updated deals RPC, client and CLI
- Expose `--maxminddbfolder` on `powd` so you can put the database file wherever
- Adds CI configuration to publish Python gRPC binding lib on release

closes #516


api/client/deals.go
```diff
@@ -5,7 +5,7 @@ import (
 	"fmt"
```
asutula, line 5: Getting the deals client up to date with the latest deals module

api/client/wallet.go
```diff
@@ -12,8 +12,8 @@ type Wallet struct {
 	client rpc.RPCServiceClient
```
asutula, line 12: Getting the wallet client up to date with the latest wallet module.

api/server/server.go
```diff
@@ -154,7 +154,7 @@ func NewServer(conf Config) (*Server, error) {
 	}
 
 	log.Info("Wiring internal components...")
-	mm, err := maxmind.New(filepath.Join(conf.MaxMindDBFolder, "./GeoLite2-City.mmdb"))
+	mm, err := maxmind.New(filepath.Join(conf.MaxMindDBFolder, "GeoLite2-City.mmdb"))
```
asutula, line 157: This `conf.MaxMindDBFolder` is set either directly on a config struct or possibly from the new `powd` flag.

cmd/powd/main.go
```diff
@@ -272,6 +274,7 @@ func setupFlags() error {
 	pflag.String("ipfsapiaddr", "/ip4/127.0.0.1/tcp/5001", "IPFS API endpoint multiaddress. (Optional, only needed if FFS is used)")
 	pflag.Int64("walletinitialfund", 4000000000000000, "FFS initial funding transaction amount in attoFIL received by --lotusmasteraddr. (if set)")
 	pflag.String("gatewayhostaddr", "0.0.0.0:7000", "Gateway host listening address")
+	pflag.String("maxminddbfolder", ".", "Path of the folder containing GeoLite2-City.mmdb")
```
asutula, line 277: New flag to set db path

cmd/pow/cmd/deals_store.go
```diff
@@ -8,13 +8,15 @@ import (
 	"strconv"
```
jsign, line 8: I'm not sure the low-level Store API would be useful for people. Expecting someone to provide the size of the stored thing to calculate the right price sounds very demanding to do "manually".

deals/rpc/rpc.proto
```diff
@@ -53,47 +53,93 @@ message RetrievalDealRecord {
     RetrievalDealInfo deal_info = 3;
 }
 
-message DealConfig {
+message StorageDealConfig {
```
jsign, line 56: I think we should notify the team building the python client about proto changes, since they're relying on them in their work.

asutula:
Ok @jsign, deals rpc, client and cli code removed.

Only thing I'm not feeling great about is that the ffs api uses the deals module internally and exposes some deals types in the public api. Seems odd to me that way. Maybe something we can change with where the types and options are defined, but also will open a can of worms with circular imports etc. Will take some thought.

deals/rpc/rpc.proto
asutula: Yes, I'll let them know. They'll be able to pull in the generated python stubs from pypi at least, and I can give them a more detailed explanation as well.

jsign:
> Ok @jsign, deals rpc, client and cli code removed.
> 
> Only thing I'm not feeling great about is that the ffs api uses the deals module internally and exposes some deals types in the public api. Seems odd to me that way. Maybe something we can change with where the types and options are defined, but also will open a can of worms with circular imports etc. Will take some thought.

Using the Deals Module internally sounds OK. But I highly agree on not exposing its types in the public API so to no lock-in in that internal component of FFS. We can create an issue to scan better that all types of the public API's are in `ffs` package.

PR #523, opened 2020-07-17
Client-internal default gRPC options

asutula:
Makes sense that you could use the client without importing gRPC modules at all and without knowing anything about gRPC. This moves the default options for transport and auth inside the client `NewClient`. Still allows options to be passed in from outside if the user needs to override or customize something. Sensible defaults are good, would have saved 30 min of support today too ;)


api/client/client.go
```diff
@@ -55,11 +58,28 @@ func (t TokenAuth) RequireTransportSecurity() bool {
 }
 
 // NewClient creates a client.
-func NewClient(ma multiaddr.Multiaddr, opts ...grpc.DialOption) (*Client, error) {
+func NewClient(ma multiaddr.Multiaddr, optsOverrides ...grpc.DialOption) (*Client, error) {
```
asutula, line 61: Making this name more clear that it can override default options

api/client/client.go
```diff
@@ -55,11 +58,28 @@ func (t TokenAuth) RequireTransportSecurity() bool {
...
+	var opts []grpc.DialOption
+	if creds != nil {
+		opts = append(opts, grpc.WithTransportCredentials(creds))
+	} else {
+		opts = append(opts, grpc.WithInsecure())
+	}
+
+	auth := TokenAuth{}
+	opts = append(opts, grpc.WithPerRPCCredentials(auth))
+	opts = append(opts, optsOverrides...)
```
asutula, line 81: Quick look at the gRPC options functions looks as though any option run later will override whatever the previous one did, so appending the overrides after the defaults here.

Side note, I love the functional options stuff, but the pattern does leave a lot of unanswered questions about how things behave in cases like providing the same option twice. Fun stuff.

cmd/pow/cmd/root.go
```diff
@@ -1,18 +1,14 @@
...
 	"os"
-	"strings"
 	"time"
 
 	"github.com/multiformats/go-multiaddr"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	client "github.com/textileio/powergate/api/client"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials"
```
asutula, line 15: Nice to not have to import anything but `client` to use the client.

PR #526, opened 2020-07-17
Simplification with StorageConfig

asutula:
The concepts of `DefaultConfig` and `CidConfig` with their associated APIs `DefaultConfig()`, `SetDefaultConfig()`, `GetDefaultCidConfig()`, `GetCidConfig()` and `PushConfig()` were a bit complicated and causing confusion. This PR replaces `DefaultConfig` and `CidConfig` with one thing: `StorageConfig`: 
```
type StorageConfig struct {
	Hot        HotConfig
	Cold       ColdConfig
	Repairable bool
}
```
All the previously mentioned API functions, some of which have been renamed for clarity, now accept, usually along with the `Cid` they apply to, and return only `StorageConfig`. `GetDefaultCidConfig()` was removed because it is no longer needed. Much easier to understand and explain.

Updated the `pow` cli to reflect this relationship better, biggest thing there is the command to push a storage config is not `pow ffs config push [cid]`.

Biggest internal impact is this touches a couple of the data stores. I haven't written any migration yet. If we decide we need to support past version's data, I can write migrations or potentially continue storing data in the old format and convert from/to that old format on the way in/out of the data store. Let's discuss that.

While we're at the breaking changes, I changed `AddToHot` to `Stage`.


api/client/ffs.go
```diff
@@ -43,25 +43,25 @@ func WithAddressType(addressType string) NewAddressOption {
...
+// WithStorageConfig overrides the Api default Cid configuration.
+func WithStorageConfig(c ffs.StorageConfig) PushStorageConfigOption {
+	return func(r *rpc.PushStorageConfigRequest) {
 		r.HasConfig = true
-		r.Config = &rpc.CidConfig{
-			Cid:  c.Cid.String(),
-			Hot:  toRPCHotConfig(c.Hot),
-			Cold: toRPCColdConfig(c.Cold),
+		r.Config = &rpc.StorageConfig{
+			Repairable: c.Repairable,
```
asutula, line 54: Bug, wasn't setting `Repairable` before, oops.

ffs/api/api.go
```diff
@@ -115,32 +115,25 @@ func (i *API) ID() ffs.APIID {
 	return i.cfg.ID
 }
 
-// DefaultConfig returns the DefaultConfig.
-func (i *API) DefaultConfig() ffs.DefaultConfig {
+// DefaultStorageConfig returns the default StorageConfig.
+func (i *API) DefaultStorageConfig() ffs.StorageConfig {
```
asutula, line 119: Being more explicit about what this is.

ffs/api/api.go
```diff
@@ -115,32 +115,25 @@ func (i *API) ID() ffs.APIID {
...
 	return i.cfg.DefaultConfig
 }
 
-// GetDefaultCidConfig returns the default instance Cid config, prepared for a particular Cid.
-func (i *API) GetDefaultCidConfig(c cid.Cid) ffs.CidConfig {
```
asutula, line 124: No longer needed. The new workflow would be calling `DefaultStorageConfig()`, then calling `PushStorageConfig` with a cid and the returned or modified `StorageConfig` returned from `DefaultStorageConfig()`.

ffs/api/api.go
asutula: Same throughout this API files.

ffs/api/istore.go
```diff
@@ -13,7 +13,7 @@ import (
 
 var (
 	dsDefaultConfig = datastore.NewKey("defaultconfig")
-	dsCidConfig     = datastore.NewKey("cidconfig")
+	dsStorageConfig = datastore.NewKey("cidconfig")
```
asutula, line 16: If we decide not to migrate data, I'll update the key name, but leaving it for now.

ffs/api/istore.go
```diff
@@ -53,41 +53,41 @@ func (s *instanceStore) getConfig() (Config, error) {
 	return c, nil
 }
 
-func (s *instanceStore) putCidConfig(c ffs.CidConfig) error {
-	if !c.Cid.Defined() {
+func (s *instanceStore) putStorageConfig(c cid.Cid, sc ffs.StorageConfig) error {
```
asutula, line 56: Need to accept a cid here too because `StorageConfig` doesn't contain any cid.

ffs/api/istore.go
```diff
@@ -121,6 +121,6 @@ func (s *instanceStore) getCids() ([]cid.Cid, error) {
 	return cids, nil
 }
 
-func makeCidConfigKey(c cid.Cid) datastore.Key {
-	return dsCidConfig.ChildString(c.String())
+func makeStorageConfigKey(c cid.Cid) datastore.Key {
+	return dsStorageConfig.ChildString(c.String())
```
asutula, line 125: So, if we want, we can migrate data that is stored in here as old `CidConfig` to `StorageConfig`.

Another option would be to keep `CidConfig` as a file-private type and continue to use it as the serialization format and convert to/from it on the way in/out.

Another option is no migration at all and move forward with the new data format.

ffs/api/option.go
```diff
@@ -3,52 +3,37 @@ package api
...
-// PushConfigOption mutates a push configuration.
-type PushConfigOption func(o *PushConfig) error
+// PushStorageConfigOption mutates a push configuration.
+type PushStorageConfigOption func(o *PushStorageConfigConfig) error
 
-// PushConfig contains options for pushing a Cid configuration.
-type PushConfig struct {
-	Config         ffs.CidConfig
+// PushStorageConfigConfig contains options for pushing a Cid configuration.
+type PushStorageConfigConfig struct {
```
asutula, line 13: Blasphemy, I know. Just being consistent in naming ;)

asutula: Not used in any public api, so that's good.

ffs/manager/manager.go
```diff
@@ -178,30 +178,30 @@ func (m *Manager) Close() error {
...
+	if err := m.ds.Put(dsDefaultStorageConfigKey, buf); err != nil {
 		return fmt.Errorf("saving default config to datastore: %s", err)
 	}
 	m.defaultConfig = dc
 	return nil
 }
 
-func loadDefaultCidConfig(ds datastore.Datastore) (ffs.DefaultConfig, error) {
-	d, err := ds.Get(dsDefaultCidConfigKey)
+func loadDefaultStorageConfig(ds datastore.Datastore) (ffs.StorageConfig, error) {
```
asutula, line 193: This ds is another potential place to migrate data or not

ffs/scheduler/internal/astore/astore.go
```diff
@@ -22,12 +22,13 @@ var (
...
 	APIID       ffs.APIID
-	Cfg         ffs.CidConfig
+	Cid         cid.Cid
+	Cfg         ffs.StorageConfig
 	ReplacedCid cid.Cid
 }
 
 // Store is a Datastore backed implementation of ActionStore, which saves latests
-// PushConfig actions for a Cid.
+// PushStorageConfig actions for a Cid.
```
asutula, line 31: This store as well is another place to potential migrate data. Same possible solutions as mentioned elsewhere.

ffs/scheduler/internal/astore/astore.go
```diff
@@ -22,12 +22,13 @@ var (
 // Action represents an action to be executed by the Scheduler.
 type Action struct {
 	APIID       ffs.APIID
-	Cfg         ffs.CidConfig
+	Cid         cid.Cid
```
asutula, line 25: `Action` needs a `Cid` field now because `StorageConfig` doesn't have one.

ffs/scheduler/scheduler.go
```diff
@@ -83,47 +83,48 @@ func New(ds datastore.TxnDatastore, l ffs.CidLogger, hs ffs.HotStorage, cs ffs.C
 	return sch, nil
 }
 
-// PushConfig queues the specified CidConfig to be executed as a new Job. It returns
+// PushConfig queues the specified StorageConfig to be executed as a new Job. It returns
 // the created JobID for further tracking of its state.
-func (s *Scheduler) PushConfig(iid ffs.APIID, cfg ffs.CidConfig) (ffs.JobID, error) {
-	return s.push(iid, cfg, cid.Undef)
+func (s *Scheduler) PushConfig(iid ffs.APIID, c cid.Cid, cfg ffs.StorageConfig) (ffs.JobID, error) {
```
asutula, line 88: Same in here where we need to pass along the cid since it's not in `StorageConfig`

ffs/types.go
```diff
@@ -93,136 +93,108 @@ type Job struct {
 	DealErrors []DealError
 }
 
-// DefaultConfig contains a default storage configuration for an Api instance.
-type DefaultConfig struct {
+// StorageConfig contains a default storage configuration for an Api instance.
+type StorageConfig struct {
```
asutula, line 97: Now there's just `StorageConfig`. Moved extensions that were previously on `CidConfig` to `StorageConfig`

ffs/api/istore.go
jsign: Mentioned some of this in the slack convo.

ffs/manager/manager.go
```diff
@@ -42,7 +42,7 @@ var (
 			},
 		},
 	}
-	dsDefaultCidConfigKey = datastore.NewKey("defaultcidconfig")
+	dsDefaultStorageConfigKey = datastore.NewKey("defaultcidconfig")
```
jsign, line 45: I think it may worth adding a `// ToDo: StorageConfig renaming` (and also in other places that we're delaying this work). So to remember there's some weird/holding change, and easier to spot when doing that?

ffs/rpc/rpc.go
```diff
@@ -507,8 +480,8 @@ func (s *RPC) Close(ctx context.Context, req *CloseRequest) (*CloseResponse, err
 	return &CloseResponse{}, nil
 }
 
-// AddToHot stores data in the Hot Storage so the resulting cid can be used in PushConfig.
-func (s *RPC) AddToHot(srv RPCService_AddToHotServer) error {
+// Stage stores data in the Hot Storage so the resulting cid can be used in PushStorageConfig.
```
jsign, line 483: Maybe `s/stores/caches` as you described in the pow cli?

ffs/Design.md
```diff
@@ -12,13 +12,13 @@ The following picture presents principal packages and interfaces that are part o
 
 The picture has an advanced scenario where different _API_ instances are wired to different _Scheduler_ instances. Component names prefixed with * don't exist but are mentioned as possible implementations of existing interfaces.
 
-The central idea about the design is that an _API_ defines the desired storing state for a _Cid_ using a _CidConfig_ struct. This struct has information about desired storing state configuration in the Hot and Cold storages.
+The central idea about the design is that an _API_ defines the desired storing state for a _Cid_ using a _StorageConfig_ struct. This struct has information about desired storing state configuration in the Hot and Cold storages.
```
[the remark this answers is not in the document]
asutula: Yes, I have a task to update the docs for this asap. Agree the videos are rough. Maybe we can add some caption near under videos like:
> Update: The blah command has changed blah blah blah.

cli-docs/pow/pow_ffs_config.md
```diff
@@ -22,6 +22,7 @@ Provides commands to manage storage configuration
 
 * [pow ffs](pow_ffs.md)	 - Provides commands to manage ffs
 * [pow ffs config default](pow_ffs_config_default.md)	 - Returns the default storage config
-* [pow ffs config prep](pow_ffs_config_prep.md)	 - Returns the default config prepped for the provided cid
-* [pow ffs config set](pow_ffs_config_set.md)	 - Sets the default cid storage config from stdin or a file
+* [pow ffs config get](pow_ffs_config_get.md)	 - Fetches the storage config for the provided cid
+* [pow ffs config push](pow_ffs_config_push.md)	 - Add data to FFS via cid
+* [pow ffs config set-default](pow_ffs_config_set-default.md)	 - Sets the default cid storage config from stdin or a file
```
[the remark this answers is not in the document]
asutula: Yea I feel the same way, but I've realized the truth which is that `ffs` is the part of the command that needs to be removed ;) That will be the best longer term solution.

ffs/rpc/rpc.go
asutula: 👍

PR #529, opened 2020-07-20
Support new Lotus API + calibration network

jsign:
This PR targets the `ntwk-calibration` Lotus branch, which is a newer codebase with *breaking APIs*, this means it can't coexist supporting `testnet` nor current `nerpa`. `ntwk-calibration` code will be the next `nerpa` network.

The command `make up` now prints a message saying that current Testnet isn't supported since is still running a Lotus version with old broken APIs. Whenever is updated we can re-enable that. The message suggests running `make nerpa-up` so to help the user connect to some network.

--- 

Apart from changes in this repo, there were many changes in others. In particular, upgrading `textileio/lotus-devnet` and `textileio/lotus-build`. In the process of upgrading `lotus-devnet` which is a pillar of the CI, we've discovered multiple bugs which ended up being resolved: e.g: [here](https://github.com/filecoin-project/go-fil-markets/pull/326), [here](https://github.com/filecoin-project/lotus/pull/2462), and [here](https://github.com/filecoin-project/lotus/pull/2746). 

Considering that `lotus@master` is still targeting `testnet`, I made `textileio/lotus-devnet` have multiple branches per network. So, the newest one which Powergate is using now is `lotus-devnet@ntwk-calibration` which is the `localnet` running the latest code of `lotus`. Whenever `master` has the latest code, we might abandon that branch.

---

While updating Powergate to the latest Lotus APIs, I took advantage and did other changes:
- Avoid panicking scenarios in test favoring test failing.
- Gas limits were unset in sending txn since now new logic about gas-limits exists in Lotus. This means that max-gas and gas-price will be estimated by the Lotus node using stat data.
- New Lotus code for localnet seems heavier. I split the integration tests in multiple packages to we can cache passing results better to rerun if some flaky-test fails. The CI is a bit painful sometimes. I also created a helper artifact that retries a test N times before failing, this reduced a lot flakyness and need to rerun.
- As a side note, the localnet with bigsectors is very unstable because some extra changes are needed in `sector-storage`. For more info see [this](https://filecoinproject.slack.com/archives/CSG8QRE0K/p1595265430374600) and the mentioned PR there. I guess eventually that would be totally fixed. Not much to do now.
- The network has defined now a minimum deal duration which is 180 days. This is now a constant for Powergate that is used to check deal duration validations.
- I changed the chain speed for localnet to be faster since now deals takes more epoch to finish. Now with one block per 500ms, they take less than 2 min (with big sectors).


docker/Makefile
```diff
@@ -1,19 +1,10 @@
+COMPAT_MSG="The current version of Powergate support newer versions than Testnet network. You might be interested in using the Calibration network (make calibration-up)"
```
jsign, line 1: Some message that gets printed if `make up` is executed. Will eventually go away when Testnet resets to latest code.

jsign:
> Lots of changes, but it all makes sense. Probably the crazier changes were in the other repos we depend on here. Nice work.

You can be totally sure about that 😅

PR #530, opened 2020-07-20
Safe string cids

asutula:
Adds two util funcs, `CidToString` and `CidFromString` and uses them globally whenever converting a cid to or from a string. Magic string value for undefined cids is `CID_UNDEF`. `CidFromString` also handles any `b` string (the string value created by the `cid` module for `cid.Undef`) that snuck in somehow.


util/util.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
```
asutula, line 5: Only real additions are here and below. Everything else is just using these funcs

trashhalo:
FYI master build was red for this pr https://github.com/textileio/powergate/runs/891687223

PR #536, opened 2020-07-22
support different network faucets

jsign:
This PR addes support for multiple faucets depending the network Powergate is being connected to.
This is needed to correct support of master address auto-creation and funding.


go.mod
```diff
@@ -55,3 +55,5 @@ require (
 	google.golang.org/grpc v1.30.0
 	google.golang.org/protobuf v1.25.0
 )
+
```
asutula, line 58: 😮

PR #538, opened 2020-07-25
Make client take a string target

asutula:
I think at one point multi address was in style, but appears that was last season. All the current clients used in `textile` repo just use simple `host:port` string. Doing the same here will make it easier to integrate the Powergate `Client` into the same system.

### DIFF
--- a/cmd/powbench/runner/runner.go
+++ b/cmd/powbench/runner/runner.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	logging "github.com/ipfs/go-log/v2"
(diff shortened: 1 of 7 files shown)


cmd/powbench/runner/runner.go
```diff
@@ -20,17 +19,17 @@ var (
 
 // TestSetup describes a test configuration.
 type TestSetup struct {
-	LotusAddr    multiaddr.Multiaddr
```
asutula, line 23: Noticed this "lotus" naming, not sure what that meant. Seems like it should be "powergate".

PR #540, opened 2020-07-29
Allow sharing of master address among ffs instances

asutula:
Achieved through a configuration flag. Pretty straight forward and supports Piñata's use case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ Usage of ./powd:
       --autocreatemasteraddr      Automatically creates & funds a master address if none is provided
       --debug                     Enable debug log level in all loggers.
       --devnet                    Indicate that will be running on an ephemeral devnet. --repopath will be autocleaned on exit.
+      --ffsusemasteraddr          Use the master address as the initial address for all new FFS instances instead of creating a new unique addess for each new FFS instance.
       --gatewayhostaddr string    Gateway host listening address (default "0.0.0.0:7000")
       --grpchostaddr string       gRPC host listening address. (default "/ip4/0.0.0.0/tcp/5002")
       --grpcwebproxyaddr string   gRPC webproxy listening address. (default "0.0.0.0:6002")
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ To build from source, you need to have Go 1.14 or newer installed.
 
 ## Design
 
-Powergate is composed of different modules which can be used independently, and compose toegheter to provide other higher-level modules.
+Powergate is composed of different modules which can be used independently, and compose together to provide other higher-level modules.
 
 Here's a high-level overview of the main components, and how Powergate interacts with IPFS and a Filecoin client:
 ![Powergate Design](https://user-images.githubusercontent.com/6136245/86490337-6dbd5200-bd3d-11ea-9895-3689dd1c8a8f.png)
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -99,6 +99,7 @@ type Config struct {
 	LotusAuthToken       string
 	LotusMasterAddr      string
 	AutocreateMasterAddr bool
+	FFSUseMasterAddr     bool
 	Devnet               bool
 	GrpcHostNetwork      string
 	GrpcHostAddress      ma.Multiaddr
@@ -210,7 +211,7 @@ func NewServer(conf Config) (*Server, error) {
 		return nil, fmt.Errorf("creating scheduler: %s", err)
 	}
 
-	ffsManager, err := manager.New(txndstr.Wrap(ds, "ffs/manager"), wm, pm, dm, sched)
+	ffsManager, err := manager.New(txndstr.Wrap(ds, "ffs/manager"), wm, pm, dm, sched, conf.FFSUseMasterAddr)
 	if err != nil {
 		return nil, fmt.Errorf("creating ffs instance: %s", err)
 	}
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -112,6 +112,10 @@ type Config struct {
 
 // NewServer starts and returns a new server with the given configuration.
 func NewServer(conf Config) (*Server, error) {
+	if conf.FFSUseMasterAddr && !(len(conf.LotusMasterAddr) > 0 || conf.AutocreateMasterAddr) {
+		return nil, fmt.Errorf("FFSUseMasterAddr requires LotusMasterAddr or AutocreateMasterAddr to be provided")
+	}
+
 	var err error
 	var masterAddr address.Address
 	c, cls, err := lotus.New(conf.LotusAddress, conf.LotusAuthToken, lotusConnectionRetries)
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -136,9 +136,10 @@ func NewServer(conf Config) (*Server, error) {
 	if err != nil {
 		return nil, fmt.Errorf("getting Lotus network name: %s", err)
 	}
-	log.Infof("Detected Lotus node connected to network: %s", network)
+	networkName := string(network)
+	log.Infof("Detected Lotus node connected to network: %s", networkName)
 
-	fchost, err := fchost.New(string(network), !conf.Devnet)
+	fchost, err := fchost.New(networkName, !conf.Devnet)
 	if err != nil {
 		return nil, fmt.Errorf("creating filecoin host: %s", err)
 	}
@@ -181,7 +182,7 @@ func NewServer(conf Config) (*Server, error) {
 	if err != nil {
 		return nil, fmt.Errorf("creating deal module: %s", err)
 	}
-	wm, err := walletModule.New(c, masterAddr, conf.WalletInitialFunds, conf.AutocreateMasterAddr)
+	wm, err := walletModule.New(c, masterAddr, conf.WalletInitialFunds, conf.AutocreateMasterAddr, networkName)
 	if err != nil {
 		return nil, fmt.Errorf("creating wallet module: %s", err)
 	}
--- a/ffs/api/api.go
+++ b/ffs/api/api.go
@@ -51,29 +51,18 @@ type API struct {
 }
 
 // New returns a new Api instance.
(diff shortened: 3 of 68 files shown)


ffs/api/api.go
```diff
@@ -51,29 +51,18 @@ type API struct {
 }
 
 // New returns a new Api instance.
-func New(ctx context.Context, ds datastore.Datastore, iid ffs.APIID, sch *scheduler.Scheduler, wm ffs.WalletManager, pm ffs.PaychManager, drm ffs.DealRecordsManager, dc ffs.StorageConfig) (*API, error) {
+func New(ds datastore.Datastore, iid ffs.APIID, sch *scheduler.Scheduler, wm ffs.WalletManager, pm ffs.PaychManager, drm ffs.DealRecordsManager, dc ffs.StorageConfig, addrInfo AddrInfo) (*API, error) {
```
asutula, line 54: This now gets the initial address injected in instead of creating it internally

ffs/interfaces.go
```diff
@@ -5,12 +5,15 @@ import (
...
 
+	"github.com/filecoin-project/go-address"
 	"github.com/ipfs/go-cid"
 	"github.com/textileio/powergate/deals"
 )
 
 // WalletManager provides access to a Lotus wallet for a Lotus node.
 type WalletManager interface {
+	// MasterAddr returns the master address.
+	MasterAddr() address.Address
```
asutula, line 16: This is needed to the ffs manages can get the master address if the config says to use it

ffs/manager/manager.go
```diff
@@ -85,8 +88,40 @@ func (m *Manager) Create(ctx context.Context) (ffs.APIID, string, error) {
...
+		}
+		a, err := address.NewFromString(res)
+		if err != nil {
+			return ffs.EmptyInstanceID, "", fmt.Errorf("decoding newly created addr: %s", err)
+		}
+		addr = a
+	}
+
+	addrInfo := api.AddrInfo{Name: "Initial Address", Addr: addr.String()}
+	switch addr.Protocol() {
```
asutula, line 108: Kind of regretting having `AddInfo.Type`, but to not break that API, I'm detecting the address type here, so in the case of a master address passed in from Lotus, we still have the `Type` field set. 

In retrospect, and maybe for v2 APIs, I think we should keep addresses as `address.Address` everywhere possible. It's a nice type, like `cid.Cid`, that serializes and has some useful methods on it.

ffs/manager/manager_test.go
```diff
@@ -41,11 +41,32 @@ func TestCreate(t *testing.T) {
 	require.True(t, id.Valid())
 }
 
+func TestCreateWithFFSMasterAddr(t *testing.T) {
```
asutula, line 44: This seems like a sufficient test. Create the manager with the `ffsUseMasterAddr = true`, create an FFS instance, verify that the FFS instance's address list and default storage config contain the master address.

ffs/interfaces.go
asutula: I guess we could also make the master address a parameter for creating the `Manager`. Let me know which you prefer.

README.md
```diff
@@ -144,6 +144,7 @@ Usage of ./powd:
       --autocreatemasteraddr      Automatically creates & funds a master address if none is provided
       --debug                     Enable debug log level in all loggers.
       --devnet                    Indicate that will be running on an ephemeral devnet. --repopath will be autocleaned on exit.
+      --ffsusemasteraddr          Use the master address as the initial address for all new FFS instances instead of creating a new unique addess for each new FFS instance.
```
jsign, line 147: I'm not completely sure about the name of the flag, but I can't figure out a better one.
Coming back always to would be good to have a more hierarchical order of flags, and being _file_ compatible too.

ffs/interfaces.go
jsign: I think this is OK. The owner of the master addr is the Wallet manager.
This API should cover the case there isn't a master addr, since that situation is possible. 
Might be good to explain in the comments that's a possible case, and explain might return a default value? (yet to see in the review)

ffs/manager/manager.go
```diff
@@ -85,8 +88,40 @@ func (m *Manager) Create(ctx context.Context) (ffs.APIID, string, error) {
 	defer m.lock.Unlock()
 
 	log.Info("creating instance")
+
+	var addr address.Address
+	if m.ffsUseMasterAddr {
+		addr = m.wm.MasterAddr()
```
jsign, line 94: Important: We should cover here that the WalletManager could not have a master address.
As a refresher, there're three possibilities:
- The masteraddr flag nor autocreatemaster are set, in that case there's no masteraddr. That means that newly created FFS instance won't be autofunded. (Since there isn't any addr as the source of the transaction).
- In any other case, there's a masteraddr. Being directly set with the masteraddr flag, or with the autocreatemaster.

So the new flag of this PR should check that if `true`, then at least one of both flags are set. 
Sounds that should happen in `server.go` as a validation while bootstrapping things. 

I'd also add some logic here to also see if `.MasterAddr()` returns the default value (or maybe change signature to also return a bool to indicate if exist). As mentioned in the above comment, that API should have a clear way of saying: _There isn't any configured masteraddr._

ffs/manager/manager.go
jsign: Yeah. Actually if v2 also includes supporting other clients than Lotus, then the _types_ problem is much broader since `address.Address` is a Lotus specific type. 
We can think of re-using Lotus specific type as the base type for Powergate... but that's hard to know.

ffs/manager/manager_test.go
jsign: Would be good to have some extra test to cover the case of having an error when indicating using the masteraddr for all FFS instances, but WalletManager not having one. Quite short one since is checking that errs, but just to cover.

README.md
asutula: Agree here.

ffs/manager/manager.go
asutula: 👍

ffs/manager/manager.go
asutula: `address.Address` is in `github.com/filecoin-project/go-address`, so maybe not lotus specific?

ffs/manager/manager.go
asutula: Decided to move the check to the `manager.New()` so that if there's a problem the error happens on start up rather than waiting for someone to create an FFS instance. This still is really just a backup because there is also not a check in `server.go` that should catch this problem even sooner.

api/server/server.go
```diff
@@ -111,6 +112,10 @@ type Config struct {
 
 // NewServer starts and returns a new server with the given configuration.
 func NewServer(conf Config) (*Server, error) {
+	if conf.FFSUseMasterAddr && !(len(conf.LotusMasterAddr) > 0 || conf.AutocreateMasterAddr) {
+		return nil, fmt.Errorf("FFSUseMasterAddr requires LotusMasterAddr or AutocreateMasterAddr to be provided")
+	}
+
```
asutula, line 118: Here's the check to make sure we have a master address or will create on if `FFSUseMasterAddr` is `true`

jsign: 👍

ffs/manager/manager.go
```diff
@@ -52,30 +53,35 @@ type Manager struct {
...
 
 	closed bool
 }
 
 // New returns a new Manager.
-func New(ds datastore.Datastore, wm ffs.WalletManager, pm ffs.PaychManager, drm ffs.DealRecordsManager, sched *scheduler.Scheduler) (*Manager, error) {
+func New(ds datastore.Datastore, wm ffs.WalletManager, pm ffs.PaychManager, drm ffs.DealRecordsManager, sched *scheduler.Scheduler, ffsUseMasterAddr bool) (*Manager, error) {
+	if ffsUseMasterAddr && wm.MasterAddr() == address.Undef {
+		return nil, fmt.Errorf("ffsUseMasterAddr requires that master address is defined")
+	}
```
asutula, line 70: This is already being checked in `NewServer`, but as a backup here as well.

ffs/manager/manager_test.go
```diff
@@ -113,32 +176,38 @@ func TestDefaultStorageConfig(t *testing.T) {
...
+	defer require.NoError(t, cls())
 	c3 := m.GetDefaultStorageConfig()
 	require.Equal(t, c, c3)
 }
 
-func newManager(t *testing.T, ds datastore.TxnDatastore) (*Manager, func()) {
-	client, addr, _ := tests.CreateLocalDevnet(t, 1)
-	wm, err := walletModule.New(client, addr, *big.NewInt(4000000000), false, "")
-	require.NoError(t, err)
+func newManager(client *apistruct.FullNodeStruct, ds datastore.TxnDatastore, masterAddr address.Address, ffsUseMasterAddr bool) (*Manager, func() error, error) {
```
asutula, line 195: Kind of annoying refactor here, but needed to do it to be able to test that creating the manager fails in certain cases.

ffs/manager/manager_test.go
```diff
@@ -23,30 +25,85 @@ func TestMain(m *testing.M) {
...
+func TestNewManagerMasterAddrFFSUseMasterAddr(t *testing.T) {
+	t.Parallel()
+	client, addr, _ := tests.CreateLocalDevnet(t, 1)
+	m, cls, err := newManager(client, tests.NewTxMapDatastore(), addr, true)
+	require.NoError(t, err)
+	require.NotNil(t, m)
+	defer require.NoError(t, cls())
+}
+
+func TestNewManagerNoMasterAddrNoFFSUseMasterAddr(t *testing.T) {
```
asutula, line 44: Test to make sure we can create the manager with no master address and not use the master address for all ffs instances

ffs/manager/manager_test.go
```diff
@@ -23,30 +25,85 @@ func TestMain(m *testing.M) {
...
+	t.Parallel()
+	client, _, _ := tests.CreateLocalDevnet(t, 1)
+	m, cls, err := newManager(client, tests.NewTxMapDatastore(), address.Undef, false)
+	require.NoError(t, err)
 	require.NotNil(t, m)
-	defer cls()
+	defer require.NoError(t, cls())
+}
+
+func TestNewManagerNoMasterAddrFFSUseMasterAddr(t *testing.T) {
```
asutula, line 53: Test to make sure creating the manager fails with no master address and use the master address for all ffs instances

PR #541, opened 2020-07-30
Export TokenAuth secure so it can be used elsewhere

asutula:
This lets the `TokenAuth` be used in other client implementations.

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -55,8 +55,8 @@ func (t TokenAuth) RequireTransportSecurity() bool {
 	return t.Secure
 }
 
-// NewClient creates a client.
-func NewClient(target string, optsOverrides ...grpc.DialOption) (*Client, error) {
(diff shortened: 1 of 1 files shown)


api/client/client.go
```diff
@@ -52,11 +52,11 @@ func (t TokenAuth) GetRequestMetadata(ctx context.Context, _ ...string) (map[str
...
 // RequireTransportSecurity specifies if the connection should be secure.
 func (t TokenAuth) RequireTransportSecurity() bool {
-	return t.secure
+	return t.Secure
 }
 
-// NewClient creates a client.
-func NewClient(target string, optsOverrides ...grpc.DialOption) (*Client, error) {
+// CreateClientConn creates a gRPC connection with sensible defaults and the provided overrides.
+func CreateClientConn(target string, optsOverrides ...grpc.DialOption) (*grpc.ClientConn, error) {
```
asutula, line 59: @jsign I just added this as well so the logic around options doesn't have to be re-created elsewhere.

PR #559, opened 2020-08-05
Stage folders support

### LINKED ISSUE #504: pow/cli: consider adding folders helper
Consider supporting `pow ffs addToHot` to folders.

jsign:
This adds functionality to `pow ffs stage` to support adding folders to IPFS.

The `pow` cli automatically detects if the added path is a file or folder. If it is a folder, it adds the folder using the IPFS HTTP API.

I didn't like the idea of exposing the IPFS node as an extra configuration knob, and also because that would mean it will be totally unauthenticated. To solve this, I created a reverse proxy in `powd`. This reverse proxy demands that all request have an authentication token which should be a valid _FFS token_, so any IPFS HTTP API call should come from an authenticated FFS instance.

Since `pow` is using the IPFS HTTP client library, I added slightly modified HTTP client transport which always adds the FFS token, so to satisfy the reverse proxy demands without building a different client.

This reverse proxy can be disabled, so `powd` isn't forced to enable this feature since I think in advance scenarios applications will be doing this actions directly in the IPFS node.

An example flow:
```
$ pow ffs create
<grab the ffs token>
$ pow ffs stage -t <ffs-token> myfolder
<grab folder cid>
$ pow ffs config push -t <ffs-token> <folder-cid>
<... wait for saving data>
$ pow ffs get -t <ffs-token> -f <folder_cid> <output-dir>
```

Some notes:
- `pow ffs stage` autodetect if the target path is a folder or dir using `os` package and acting accordingly.
- `pow ffs get` has a new flag `-f` to indicate that the Cid to be retrieved is a folder. Even if that flag is missed and is tried to retrieve as a file, an error indicating that "Sorry, this is a folder" will appear so the user can notice.
- The folder content is placed in `<output-dir>`.

Closes #504

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ The binary will be placed automatically in `$GOPATH/bin` which in general is in 
 You can read the [generated CLI docs](https://github.com/textileio/powergate/blob/master/cli-docs/pow/pow.md) in this repo, or run `pow` with the `--help` flag to see the available commands:
 
 ```
+
 $ pow --help
 A client for storage and retreival of powergate data
 
@@ -83,22 +84,18 @@ Usage:
 
 Available Commands:
   asks        Provides commands to view asks data
-  deal        Interactive storage deal
-  deals       Provides commands to manage storage deals
+  faults      Provides commands to view faults data
   ffs         Provides commands to manage ffs
   health      Display the node health status
   help        Help about any command
-  init        Initializes a config file with the provided values or defaults
   miners      Provides commands to view miners data
   net         Provides commands related to peers and network
   reputation  Provides commands to view miner reputation data
-  faults      Provides commands to view faults data
   wallet      Provides commands about filecoin wallets
 
 Flags:
-      --config string          config file (default is $HOME/.powergate.yaml)
   -h, --help                   help for pow
-      --serverAddress string   address of the powergate service api (default "/ip4/127.0.0.1/tcp/5002")
+      --serverAddress string   address of the powergate service api (default "127.0.0.1:5002")
 
 Use "pow [command] --help" for more information about a command.
 ```
@@ -140,7 +137,7 @@ make build-powd
 You can run the `-h` flag to see the configurable flags:
 ```bash
 $ powd -h
-Usage of ./powd:
+Usage of powd:
       --autocreatemasteraddr      Automatically creates & funds a master address if none is provided
       --debug                     Enable debug log level in all loggers.
       --devnet                    Indicate that will be running on an ephemeral devnet. --repopath will be autocleaned on exit.
@@ -149,6 +146,7 @@ Usage of ./powd:
       --grpchostaddr string       gRPC host listening address. (default "/ip4/0.0.0.0/tcp/5002")
       --grpcwebproxyaddr string   gRPC webproxy listening address. (default "0.0.0.0:6002")
       --ipfsapiaddr string        IPFS API endpoint multiaddress. (Optional, only needed if FFS is used) (default "/ip4/127.0.0.1/tcp/5001")
+      --ipfsrevproxyaddr string   IPFS reverse proxy listen address (default "127.0.0.1:6003")
       --lotushost string          Lotus client API endpoint multiaddress. (default "/ip4/127.0.0.1/tcp/1234")
       --lotusmasteraddr string    Existing wallet address in Lotus to be used as source of funding for new FFS instances. (Optional)
       --lotustoken string         Lotus API authorization token. This flag or --lotustoken file are mandatory.
@@ -156,7 +154,6 @@ Usage of ./powd:
       --maxminddbfolder string    Path of the folder containing GeoLite2-City.mmdb (default ".")
       --repopath string           Path of the repository where Powergate state will be saved. (default "~/.powergate")
       --walletinitialfund int     FFS initial funding transaction amount in attoFIL received by --lotusmasteraddr. (if set) (default 4000000000000000)
-pflag: help requested
 ```
 
 We'll soon provide better information about Powergate configurations, stay tuned! 📻
--- a/api/client/ffs.go
+++ b/api/client/ffs.go
@@ -4,9 +4,17 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net/http"
+	"os"
+	"strings"
 	"time"
 
 	cid "github.com/ipfs/go-cid"
+	files "github.com/ipfs/go-ipfs-files"
+	httpapi "github.com/ipfs/go-ipfs-http-client"
+	"github.com/ipfs/interface-go-ipfs-core/options"
+	ipfspath "github.com/ipfs/interface-go-ipfs-core/path"
+	"github.com/multiformats/go-multiaddr"
 	"github.com/textileio/powergate/deals"
 	"github.com/textileio/powergate/ffs"
 	"github.com/textileio/powergate/ffs/api"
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -149,14 +146,14 @@ Usage of ./powd:
       --grpchostaddr string       gRPC host listening address. (default "/ip4/0.0.0.0/tcp/5002")
       --grpcwebproxyaddr string   gRPC webproxy listening address. (default "0.0.0.0:6002")
       --ipfsapiaddr string        IPFS API endpoint multiaddress. (Optional, only needed if FFS is used) (default "/ip4/127.0.0.1/tcp/5001")
+      --ipfsrevproxyaddr string   IPFS reverse proxy listen address (default "127.0.0.1:6003")
```
jsign, line 149: The IPFS HTTP API reverse proxy can be configured.

api/server/server.go
```diff
@@ -260,12 +264,74 @@ func NewServer(conf Config) (*Server, error) {
...
+	}
+
+	go func() {
+		if err := rp.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Fatalf("serving index http: %v", err)
+		}
+	}()
+	return rp, nil
+}
+
```
jsign, line 334: So here is the reverse proxy in `powd`, which is only enabled if a non-empty `IPFSRevProxyAddr` was provided on startup.

If that's the case, it creates a reverse proxy to the IPFS HTTP API with an extra tweak of enforcing a valid FFS token is provided in all HTTP calls.

cli-docs/pow/pow_ffs_stage.md
```diff
@@ -1,10 +1,10 @@
 ## pow ffs stage
 
-Temporarily cache data in the Hot layer in preparation for pushing a cid storage config
+Temporarily stage data in the Hot layer in preparation for pushing a cid storage config
```
jsign, line 3: Had the sensation that only using _stage_ will be better than introducing _cache_ which might have other semantic assumptions.

cmd/pow/cmd/ffs_stage.go
```diff
@@ -31,18 +39,84 @@ var ffsStageCmd = &cobra.Command{
...
+	return &ffsHeaderDecorator{
+		ffsToken: ffsToken,
+	}
+}
+
+func (fhd ffsHeaderDecorator) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header["x-ipfs-ffs-auth"] = []string{fhd.ffsToken}
+
+	return http.DefaultTransport.RoundTrip(req)
+}
```
jsign, line 122: All http calls made by the IPFS client lib, will be hooked to add an auth header to be authenticated.
The server-side reverse proxy checks for this header, before redirecting the request to  the IPFS node.

cmd/powd/main.go
```diff
@@ -280,6 +282,7 @@ func setupFlags() error {
 	pflag.String("ipfsapiaddr", "/ip4/127.0.0.1/tcp/5001", "IPFS API endpoint multiaddress. (Optional, only needed if FFS is used)")
 	pflag.Int64("walletinitialfund", 4000000000000000, "FFS initial funding transaction amount in attoFIL received by --lotusmasteraddr. (if set)")
 	pflag.String("gatewayhostaddr", "0.0.0.0:7000", "Gateway host listening address")
+	pflag.String("ipfsrevproxyaddr", "127.0.0.1:6003", "IPFS reverse proxy listen address")
```
jsign, line 285: I'm not a big fan of having this feature enabled, so my first thought was doing `""` to let be disabled.
Since some people have requested this featured, I opted to leave it open for localhost only.

api/server/server.go
```diff
@@ -260,12 +264,74 @@ func NewServer(conf Config) (*Server, error) {
...
+func newHTTPFFSAuthInterceptor(cont http.Handler, m *manager.Manager) *ffsHTTPAuth {
+	fha := &ffsHTTPAuth{
+		cont:       cont,
+		ffsManager: m,
+	}
+
+	return fha
+}
+
+func (fha *ffsHTTPAuth) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
```
jsign, line 295: `ffsHTTPAuth` is an "interceptor" which verifies that all request passing through the IPFS reverse proxy are authenticated with an FFS token.

api/server/server.go
```diff
@@ -384,17 +450,26 @@ func startIndexHTTPServer(s *Server) *http.Server {
 
 // Close shuts down the server.
 func (s *Server) Close() {
+	if s.ipfsRevProxy != nil {
```
jsign, line 453: Closing `ipfsRevProxy` is conditional since it can be disabled.

cmd/pow/cmd/ffs_stage.go
```diff
@@ -31,18 +39,94 @@ var ffsStageCmd = &cobra.Command{
...
+	}
+	pth, err := ipfs.Unixfs().Add(context.Background(), files.ToDir(ff), opts...)
+	if err != nil {
+		return cid.Undef, err
+	}
+
+	return pth.Cid(), nil
+}
+
+func newDecoratedIPFSAPI(ipfsRevProxyMaddr, ffsToken string) (*httpapi.HttpApi, error) {
```
jsign, line 107: This returns an IPFS HTTP api, but with a custom client which adds the auth header on every HTTP call.
This header will be checked by the server to allow the request to pass through the IPFS node. If that isn't the case, the request will be Unauthorized (401).

api/server/server.go
```diff
@@ -260,12 +264,74 @@ func NewServer(conf Config) (*Server, error) {
 	}
 
 	s.indexServer = startIndexHTTPServer(s)
+	if err != nil {
```
asutula, line 267: I don't think this is checking an error from starting the index server

cmd/pow/cmd/ffs_get.go
```diff
@@ -3,18 +3,23 @@ package cmd
...
 	"github.com/ipfs/go-cid"
+	files "github.com/ipfs/go-ipfs-files"
+	ipfspath "github.com/ipfs/interface-go-ipfs-core/path"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
 func init() {
 	ffsGetCmd.Flags().StringP("token", "t", "", "token of the request")
+	ffsGetCmd.Flags().String("ipfsrevproxy", "/ip4/127.0.0.1/tcp/6003", "Powergate IPFS reverse proxy multiaddr")
```
asutula, line 21: Think we should make this flag (and the one in `stage`) normal `host:port` just to be consistent with how we specify the server address.

api/server/server.go
jsign: Uhm, looks like a merge problem. Will fix.

cmd/pow/cmd/ffs_get.go
jsign: Sure, agree. Will change.

jsign:
> This is a conditional approval, explained below.
> 
> If you think this should all just live in the `pow` CLI, then ok I approve, left a couple nit comments to fix.
> 
> But, I feel like this should be incorporated into the `Client`. I think it would be cool to just make the `Client.FFS.Stage` method take in a folder/file path, and move all this logic in there. That way, other things that use the `Client` could add folders.
> 
> Further, we could just use the IPFS client to add folders, but also single files as well, and then we could remove that code from our backend FFS and RPC implementation.
> 
> Open to discussion here for sure, and also open to making these changes later if we need to get this CLI feature out ASAP.

I've should added a bit more context to this PR since has some previous convos around this topic.
First of all, I agree that moving the logic to the client would be useful. I can do that now.

Regarding using the same "logic" to add files, yeah that's an option since this is a rather generic solution.
But doing that has the assumption to always use IPFS in the backend, which might be the most probable thing but doing that would be a full confirmation (which I'm not 100% sure yet). In any case, if we decide to do it would be quite easy with all this plumbing.

jsign:
> Sounds good. I agree about the IPFS commitment part.
> 
> Will be curious how you solve the `Client` integration since the current `Stage` method takes an `io.Reader` and the IPFS client wants a `string` path. I guess `Stage` can always just take a `string` path, and if we detect that it's a single file, create a `Reader` internally and upload it. Little less flexible than using a `Reader` as the func input, oh well.

Yeah, I'm solving that now. Not an easy decision since changing `io.Reader` to other things is breaking APIs, plus might be less flexible for the underlying source as you noticed.

I leaned towards a new client API for adding folders. Not entirely ideal too, but I keep believing that all this "add to host storage" functionality are just helpers about what the real app should be doing itself.

api/client/ffs.go
```diff
@@ -371,6 +379,24 @@ func (f *FFS) Remove(ctx context.Context, c cid.Cid) error {
 	return err
 }
 
+// GetFolder retrieves to outputDir a Cid which corresponds to a folder.
+func (f *FFS) GetFolder(ctx context.Context, ipfsRevProxyAddr string, c cid.Cid, outputDir string) error {
```
jsign, line 383: Unfortunately the `Get` API isn't very compatible with folders, since it returns an `io.Reader`.
As we mentioned in the PR conv, looks like we should keep `io.Reader` for generic usages.
All these things are helpers.

Another ugly point, why `ipfsRevProxyAddr` is a param?. Ideally that value would live in `FFS` client, but that means adding a param to `NewClient` which:
- Breaks APIs
- 99.9% of people won't have a clue of what that is.
- Would be nice to allow that config to be a optional param, but we already used the optional slot for gRPC options. Ideally for some V2 we should have our own optional, and one of them being th gRPC one, so we can keep extending the API without breaking it or forcing things which doesn't make sense for multiple cases. Thoughts?

asutula: Yea though one. I think having a less than ideal api is fine for now, especially considering HFS last week. For V2, there are all sorts of changes we can make, including breaking `NewClient` and/or making `Stage` and `Get` more general but less flexible by using file paths (since they are just helpers), or we could maintains separate functions. Sky is the limit.

PR #561, opened 2020-08-05
Single Proxy

asutula:
I combined the gRPC web proxy and the IPFS reverse proxy. Less code to maintain, and for the JS client, the user won't have to provide an "ipfs reverse proxy" argument to construct the client or make calls to `stageFolder` and `getFolder` (it's still necessary in the Go client because the Go gRPC client uses the native gRPC transport which has to listen on a separate address 😞).

Only downside I see is that the reverse proxy is always running on the server now. If we want, we could provide a config flag `bool` that would leave the `ffsHTTPAuth` as `nil` and no proxy handling for IPFS requests would happen in that case.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -230,11 +230,11 @@ func NewServer(conf Config) (*Server, error) {
 	log.Info("Starting gRPC, gateway and index HTTP servers...")
 	grpcServer := grpc.NewServer(conf.GrpcServerOpts...)
 	wrappedGRPCServer := wrapGRPCServer(grpcServer)
-	httpFFSAuthInerceptor, err := newHTTPFFSAuthInterceptor(conf, ffsManager)
+	httpFFSAuthInterceptor, err := newHTTPFFSAuthInterceptor(conf, ffsManager)
 	if err != nil {
 		return nil, fmt.Errorf("creating ffsHTTPAuth: %s", err)
 	}
-	webProxy := createProxyServer(wrappedGRPCServer, httpFFSAuthInerceptor, conf.GrpcWebProxyAddress)
+	webProxy := createProxyServer(wrappedGRPCServer, httpFFSAuthInterceptor, conf.GrpcWebProxyAddress)
 
 	gateway := gateway.NewGateway(conf.GatewayHostAddr, ai, mi, si, rm)
 	gateway.Start()
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -88,12 +88,12 @@ type Server struct {
 	hs         ffs.HotStorage
 	l          *cidlogger.CidLogger
 
-	grpcServer   *grpc.Server
-	grpcWebProxy *http.Server
+	grpcServer *grpc.Server
 
-	gateway      *gateway.Gateway
-	indexServer  *http.Server
-	ipfsRevProxy *http.Server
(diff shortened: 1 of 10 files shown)


api/server/server.go
```diff
@@ -299,39 +309,28 @@ func (fha *ffsHTTPAuth) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
-	}
+func (fha *ffsHTTPAuth) IsIPFSRequest(r *http.Request) bool {
+	return len(r.Header.Get("x-ipfs-ffs-auth")) > 0
+}
 
-	go func() {
-		if err := rp.ListenAndServe(); err != nil && err != http.ErrServerClosed {
-			log.Fatalf("serving index http: %v", err)
+func createProxyServer(wrappedGRPCServer *grpcweb.WrappedGrpcServer, fha *ffsHTTPAuth, webProxyAddr string) *http.Server {
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
asutula, line 317: In here we detect if we're dealing with an IPFS request or a gRPC request and handle accordingly.

api/server/server.go
```diff
@@ -299,39 +309,28 @@ func (fha *ffsHTTPAuth) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
-		return nil, fmt.Errorf("generating IPFS URL for reverse proxy: %s", err)
-	}
-	rph := httputil.NewSingleHostReverseProxy(urlIPFS)
-	fha := newHTTPFFSAuthInterceptor(rph, m)
-	rp := &http.Server{
-		Addr:    conf.IPFSRevProxyAddr,
-		Handler: fha,
-	}
+func (fha *ffsHTTPAuth) IsIPFSRequest(r *http.Request) bool {
+	return len(r.Header.Get("x-ipfs-ffs-auth")) > 0
```
asutula, line 313: Simple detection of an IPFS request. I like this because it's very specific.

api/server/server.go
```diff
@@ -229,7 +228,13 @@ func NewServer(conf Config) (*Server, error) {
...
 
 	log.Info("Starting gRPC, gateway and index HTTP servers...")
-	grpcServer, grpcWebProxy := createGRPCServer(conf.GrpcServerOpts, conf.GrpcWebProxyAddress)
+	grpcServer := grpc.NewServer(conf.GrpcServerOpts...)
+	wrappedGRPCServer := wrapGRPCServer(grpcServer)
+	httpFFSAuthInerceptor, err := newHTTPFFSAuthInterceptor(conf, ffsManager)
+	if err != nil {
+		return nil, fmt.Errorf("creating ffsHTTPAuth: %s", err)
+	}
+	webProxy := createProxyServer(wrappedGRPCServer, httpFFSAuthInerceptor, conf.GrpcWebProxyAddress)
```
jsign, line 237: Typo: Inerceptor

jsign:
> Only downside I see is that the reverse proxy is always running on the server now. If we want, we could provide a config flag `bool` that would leave the `ffsHTTPAuth` as `nil` and no proxy handling for IPFS requests would happen in that case.
Yeah, at least is the rev proxy is authenticated. But eventually would be good to maybe have that extra flag.

PR #564, opened 2020-08-07
Various updates

asutula:
Things I found needed to be tweaked while integrating Powergate into the JS client update.


api/client/client.go
```diff
@@ -84,31 +84,36 @@ func CreateClientConn(target string, optsOverrides ...grpc.DialOption) (*grpc.Cl
...
+// Host returns the client host address.
+func (c *Client) Host() string {
 	return c.conn.Target()
 }
 
+// BuildInfo returns build info about the server.
+func (c *Client) BuildInfo(ctx context.Context) (*buildinfoRpc.BuildInfoResponse, error) {
+	return c.buildInfoClient.BuildInfo(ctx, &buildinfoRpc.BuildInfoRequest{})
+}
+
```
asutula, line 116: Moving this to a top level func in the client since there are no other build info funcs that justify its own sub module.

api/server/server.go
```diff
@@ -321,6 +321,8 @@ func createProxyServer(wrappedGRPCServer *grpcweb.WrappedGrpcServer, fha *ffsHTT
 			wrappedGRPCServer.IsAcceptableGrpcCorsRequest(r) ||
 			wrappedGRPCServer.IsGrpcWebSocketRequest(r) {
 			wrappedGRPCServer.ServeHTTP(w, r)
+		} else {
+			http.NotFound(w, r)
```
asutula, line 325: Handing the case where the proxy doesn't match any handlers to the request

PR #570, opened 2020-08-14
support calibration 8.13.X & persist logs to file

### LINKED ISSUE #571: powd: persiste logs to file
Add the feature to persist all printed logs to a file, so there's log history upon restarts.

jsign:
This PR adds supports for the new Calibration Network 8.13.X. 

At 8.13.1 there's still a problem with provider collateral in storing some deals. 
The issue is being investigated now. This is affecting using `lotus-devent` with bigsectors enabled, but hopefully, this will get fixed soon. ~We'll see if that fix happens fast enough to be included in this PR.~ Doesnt.

So `local-devnet` wise, most work involved importing some deps that are now part of Lotus instead of Go Modules dependencies ([here](https://github.com/textileio/lotus-devnet/commit/3632248a728059b2b6807c887947649c1033cb0d)). Some other commits happened later to adjust some other tricky things.

Regarding Powergate, the fact that now some deps were vendored in Lotus required some `replace` magic in our Go-mod. This since the original Lotus `go.mod` has `replace` rules to vendored folders. Our `replace` targets a problematic dependency to a remote-commit that should be the same. After some talk, looks like they'll try to fix this vendoring complexity so we could eventually avoid this.

I also took advantage and included that all generated logs gets persisted in `powd.logs` file in the repo folder. That's useful for having the log history on `powd` resets. (Closes #571)


cmd/powd/main.go
```diff
@@ -166,10 +167,17 @@ func setupInstrumentation() (func(), error) {
...
-		return err
+func setupLogging(repoPath string) error {
+	if err := os.MkdirAll(repoPath, os.ModePerm); err != nil {
+		return fmt.Errorf("creating repo folder: %s", err)
+	}
+	cfg := logging.Config{
+		Level:  logging.LevelError,
+		Stderr: true,
+		Stdout: true,
+		File:   filepath.Join(repoPath, "powd.log"),
```
jsign, line 178: Saving logs in file.

docker/Makefile
```diff
@@ -26,7 +26,7 @@ localnet:
 
 
 calibration-up: calibration-down
-	LOTUS_IMAGE_TAG=ntwk-calibration-8.8.0 \
+	LOTUS_IMAGE_TAG=ntwk-calibration-8.13.1-f094273f \
```
jsign, line 29: There's not tag for a hot-fix that solves some problems in big file retrievals.
I kind of made a custom version to provide best possible experience in the network.

ffs/integrationtest/general/general_test.go
```diff
@@ -105,12 +106,45 @@ func TestInfo(t *testing.T) {
...
 		r := rand.New(rand.NewSource(22))
 		n := 2
+		retriesPerDeal := 5
 		for i := 0; i < n; i++ {
-			cid, _ := it.AddRandomFile(t, r, ipfs)
-			jid, err := fapi.PushStorageConfig(cid)
-			require.NoError(t, err)
-			it.RequireJobState(t, fapi, jid, ffs.Success)
-			it.RequireStorageConfig(t, fapi, cid, nil)
+			for j := 0; j < retriesPerDeal; j++ {
```
jsign, line 111: So `TestInfo` is mostly the only remaining flaky test. The `RunFlaky(..)` trick doesn't quite work here for the following reason: This test make two deals per test, so if any of both deals fails then a retry should be made. But the probability of failure is much bigger than other tests, since if _any_ of those fail that would be enough for failing.

This might remediate that... which tries to have some extra attempts "per retry" in each deal that should be made. A bit of a stretch. I lost quite a lot of time making this pass, so a bit tired of its behaviour.

If still gets flaky, we should make `n := 1` and adjust the test. It wouldn't be totally ideal from my POV, but it would check the important thing of the test anyway. I'm very frustrated about this local-devnet flakyness.

go.mod
```diff
@@ -57,3 +57,5 @@ require (
 )
 
 replace github.com/dgraph-io/badger/v2 => github.com/dgraph-io/badger/v2 v2.0.1-rc1.0.20200718033852-37ee16d8ad1c
+
+replace github.com/filecoin-project/sector-storage => github.com/filecoin-project/sector-storage v0.0.0-20200812222704-c3077fb85119
```
jsign, line 61: Magic trick for solving the `replace` Lotus with custom vendored folder.

deals/module/deals.go
```diff
@@ -188,10 +188,19 @@ func (m *Module) retrieve(ctx context.Context, waddr string, cid cid.Cid, ref *a
 		return ErrRetrievalNoAvailableProviders
 	}
 	for _, o := range offers {
-		if err = m.api.ClientRetrieve(ctx, o.Order(addr), ref); err != nil {
+		upts, err := m.api.ClientRetrieve(ctx, o.Order(addr), ref)
```
jsign, line 191: Now `ClientRetrieve` returns a channel. They push updated about the retrieval, and close the channel when it's done. I adjusted our code accordingly.

Eventually might be interesting to surface this returned-channel info to the _DealsModule_? Something to think about.

docker/docker-compose-localnet.yaml
```diff
@@ -23,7 +23,7 @@ services:
       - 5001:5001
 
   lotus:
-    image: textile/lotus-devnet:ntwk-calibration-8.8.0
+    image: textile/lotus-devnet:ntwk-calibration-8.13.1.f0942
```
asutula, line 26: Just noticing this version is different than the one in `docker/Makefile`, making sure that is intended.

tests/ldevnet.go
```diff
@@ -39,7 +39,7 @@ func LaunchDevnetDocker(t TestingTWithCleanup, numMiners int, ipfsMaddr string,
 	}
 
 	repository := "textile/lotus-devnet"
-	tag := "ntwk-calibration-8.8.0"
+	tag := "ntwk-calibration-8.13.1.f0942"
```
asutula, line 42: also just flagging this one to check

jsign: Oh, yeah, they're the same. Just that one tag name got a bit longer/shorter than the other.
One is the `textile/local-devnet` and the other `textile/lotus`.

The prefix of both is the same. I'm understanding correctly your point? (just want to double-check that was your point)

asutula: Oh yea they are the same, didn't notice. All good!

PR #573, opened 2020-08-17
support nerpa

jsign:
Adds support for `ntwk-nerpa` with `make nerpa-up`.
This network supports `POWD_AUTOCREATEMASTERADDR=true`.

I took a look at the commit history of `ntwk-nerpa` and is almost at the same level that we our `ntwk-calibration` support, so all should work fine. 

I ran `make nerpa-up` locally and all things look good. Unfortunately, the network doesn't have any miner replying Storage Asks, so maybe tomorrow there might be some.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -44,3 +44,24 @@ calibration-down:
 		-f powergate-build-context.yaml \
 		down
 .PHONY: calibration-down
+
+
+nerpa-up: nerpa-down
+	LOTUS_IMAGE_TAG=nerpa-53c7594 \
+	docker-compose \
+		-p nerpa \
+		-f docker-compose.yaml \
+		-f ipfs-image.yaml \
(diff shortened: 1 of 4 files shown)


docker/Makefile
```diff
@@ -44,3 +44,24 @@ calibration-down:
 		-f powergate-build-context.yaml \
 		down
 .PHONY: calibration-down
+
+
+nerpa-up: nerpa-down
+	LOTUS_IMAGE_TAG=nerpa-53c7594 \
```
jsign, line 50: Unfortunately, there isn't a version TAG in the `ntwk-nerpa` Lotus branch (yet?).
So the used image from `textileio/lotus-build` is `nerpa-53c7594` where `53c7594` is the _sha1_ of the commit of the `lotus@ntwk-nerpa`.

PR #574, opened 2020-08-18
Localnet docker compose: enable PubSub on IPFS node, open Lotus RPC port

ukstv:
`make localnet` is a helpful command that starts the whole Powergate suite, including Lotus and IPFS. Yet, IPFS node does not have pubsub enabled, which is quite useful to have, and Lotus have its RPC port blocked.

This pull request enables pubsub for IPFS, and opens Lotus RPC port. With the latter, one could run lotus cli against the node started by `make localnet`.

### DIFF
--- a/docker/docker-compose-localnet.yaml
+++ b/docker/docker-compose-localnet.yaml
@@ -19,6 +19,8 @@ services:
     restart: unless-stopped
 
   ipfs:
+    entrypoint: ""
+    command: /bin/sh -c "/usr/local/bin/start_ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin '[\"*\"]' && /usr/local/bin/start_ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods '[\"POST\",\"GET\",\"PUT\"]' && /usr/local/bin/start_ipfs daemon  --enable-pubsub-experiment"
     ports:
       - 5001:5001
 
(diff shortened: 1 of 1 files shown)


jsign:
Lotus doesn't have its RPC blocked, if that was the case the `lotus-devnet` would be totally worthless :P
The port 7777 you see in the port-mapping _is_ the RPC. Notice how Powergate is configured as `POWD_LOTUSHOST=/dns4/lotus/tcp/7777`. If you want `1234` (or any other) you can just edit the port-mapping in the compose file.

The `make localnet` is meant for some minimal configuration to run tests locally with default values. I'm not sure if everyone would be interested in having this feature enabled, or why not another experimental one, or CORS setup. That was my logic behind using the default which is some unopinionated configuration for things.

ukstv:
ORLY? [Lotus API Documentation](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api.md) states 1234 is JSON RPC indeed.

CORS might be too much, yet pubsub is pretty common these days in the wild.

jsign:
> ORLY? [Lotus API Documentation](https://github.com/filecoin-project/lotus/blob/master/documentation/en/api.md) states 1234 is JSON RPC indeed.

Notice that `make localnet` runs `textile/lotus-devnet`, which is a Lotus node with a mocked sector-builder. The listening port of the API is 7777 [by design](https://github.com/textileio/lotus-devnet/blob/66d3bc06799b162c450d93274c44c2f4d77d48cb/ldevnet/ldevnet.go#L175) because I preferred not to collide to another possibly running Lotus node, but all that is easy to change if it's a problem (e.g: edit the docker-compose).

> CORS might be too much, yet pubsub is pretty common these days in the wild.
> 
> It would be fine to just locally edit docker-compose if the setup is totally isolated, and does not mess with default IPFS/Lotus ports.

Agree, but why make that assumption? Changing the default docker-compose should make sense in the general case. Of course, anyone can tune their configs in whatever context make sense for them.

PR #580, opened 2020-08-29
--enable-pubsub-experiement in ipfs-image.yaml

aphelionz:
Similar to #574 but a more minimal change. I'm working on https://github.com/orbitdb/orbit-db-powergate-io and OrbitDB requires pubsub for DB replication.

### DIFF
--- a/docker/ipfs-image.yaml
+++ b/docker/ipfs-image.yaml
@@ -4,3 +4,4 @@ services:
 
   ipfs:
     image: ipfs/go-ipfs:v0.6.0
+    command: ["daemon", "--migrate=true", "--enable-pubsub-experiment"]


jsign:
I still stress in my argument of not considering external project preferences for the general Powergate configuration. If tomorrow someone wants to switch to another experimental flag because it makes sense for them, I don't want that to be forced in your project. I think projects should use _their_ configuration as needed. If you entangle your project with whatever Powergate decides the configuration to be, I think that isn't a good thing for you in the long-run.

Despite that, I also want to be pragmatic and help people if now working this way would unblock things. You're the second user that wants that flag to be enabled by default, so I think that's a good reason to provided as default if helps in general.

I just want to make the caveat that if in the future some other user wants to push another configuration flag, this PR being merged won't be a convincing argument on its own means. It's a slippery slope. :) 

It seems you need to accept the DCO, mind doing that?

aphelionz:
I do see where you're coming from, and I really appreciate the consideration. TBH I'm not sure why pubsub isn't enabled by default by now in go-ipfs, but here we are :shrug: 

Anyway, thanks for accepting this and I'll accept the DCO right now

PR #585, opened 2020-09-07
Support testnet v0.6.1 & avoid single lotus client & increase stage timeout

jsign:
This PR updates to using v0.6.0 version of Lotus.

It also addresses a problem of using a single Lotus client in all the application lifespan. The problem was already reported.
Now the code provides a client builder to all the modules that depend on making Louts API calls, so we can create short-scoped Lotus client to avoid a long running instance to become faulty and halt all Powergate functionality.
Even if the underlying problem in the Lotus client is fixes, this might be a good idea to keep anyway.

This PR also disables the Fault index builder since the underlying Lotus API was commented out. It will be re-enabled when that gets fixed.

Additionally, the `stage` ffs command timeout was increased to being able to upload bigger files.

The code changes also have some temporary changes to support a particular set of trusted miners in current `testnet`. 

Dockerfile now uses latest Go 1.15.1 version.


jsign:
Merging this, not really much fun stuff to review.
This code is already being run live in `testnet` now.

PR #586, opened 2020-09-10
slow down indexes and update to 0.6.2

jsign:
- Make indexes build information in higher intervals since Louts can't handle the load without affecting other processes.
- Update client to v0.6.2.
- Restrict the scope of lock in wallet creation since lotus API call takes time.


ffs/manager/manager.go
```diff
@@ -135,6 +133,8 @@ func (m *Manager) Create(ctx context.Context) (ffs.APIID, string, error) {
 	if err != nil {
 		return ffs.EmptyInstanceID, "", fmt.Errorf("generating auth token for %s: %s", fapi.ID(), err)
 	}
+	m.lock.Lock()
+	defer m.lock.Unlock()
```
jsign, line 137: Move the lock closer to protecting the concurrent write in the map. Before it included the API call to lotus which can have deals, making concurrent users creating wallets wait too much time.

asutula (approved):

PR #589, opened 2020-09-10
Decrease indexing load & fail in unknown Jobs

jsign:
In short:
- Decrease concurrency while building the indexes. I want to be as sure as possible the sync process isn't messed up.
- If we get a `WatchJobs` API call with unknown Jobs, fail fast. In theory this shouldn't be a problem, but now considering a situation where the Hub could ask for Jobs from a previous (reset) Powergate instance, we want to give a clear indication that we don't have a clue of those Jobs anymore. (See https://github.com/textileio/textile/pull/330)

### DIFF
--- a/index/faults/module/faults.go
+++ b/index/faults/module/faults.go
@@ -123,9 +123,6 @@ func (s *Index) Close() error {
 // start is a long running job that keeps the index up to date with chain updates.
 func (s *Index) start() {
 	defer close(s.finished)
-	if err := s.updateIndex(); err != nil {
-		log.Errorf("initial updating faults index: %s", err)
-	}
 	for {
 		select {
 		case <-s.ctx.Done():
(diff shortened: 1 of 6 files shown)


index/faults/module/faults.go
```diff
@@ -123,9 +123,6 @@ func (s *Index) Close() error {
 // start is a long running job that keeps the index up to date with chain updates.
 func (s *Index) start() {
 	defer close(s.finished)
-	if err := s.updateIndex(); err != nil {
-		log.Errorf("initial updating faults index: %s", err)
-	}
```
jsign, line 128: Since the Fault Index is disabled, we shouldn't consider firing updating this data on startup.

index/miner/module/meta.go
```diff
@@ -18,9 +18,9 @@ import (
 )
 
 var (
```
jsign, line 20: Metadata information got the largest cut. At this moment, this information has 0 value for FFS miner selection, keep  these updates contribute as tiny load as possible for now.

andrewxhill (approved):

PR #590, opened 2020-09-14
Arbitrary retrievals preparation & new features and improvements

jsign:
This PR started with the quest of preparing to support arbitrary retrievals of deals from the Filecoin network, and ended up improving things or adding other features.

At this point the external API hasn't changed, but it has a lot of internal changes and breaking changes at the Datastore namespace layer, so merging this PR should bump a minor version.

My first approach to solve the "arbitrary retrievals" feature, was somewhat including that concept in the existing "Job" concept. Tried hard but found dead ends. The main problem is that these "arbitrary retrievals" allows to use an IPLD selector to retrieve partial information. So even if you're retrieving from a stored PayloadCid, since you're using a selector you aren't possibly retrieving that Cid, but maybe some partial set of the data.

In fact, you could do multiple retrievals of the same Cid with different selectors, and the underlying data you'll be getting has a different Cid (say, different folders of the same UnixFS dag root)... so all the concepts we had about configurations and information glued to a particular "Cid" breaks apart. In these retrievals the PayloadCid isn't that important, but the "PayloadCid" and the selector. 

At the end, went to the longer but clearer route. Have two types of Jobs. `StorageJob` which are the ones we always had, and new `RetrievalJob` which provides this new feature. Not forcing both features be provided by the same `Job` concept allows to less complex code, since they're completely different features and each kind of `Job` needs different data to execute and save its results.

So, since the original `Job` was moved to `StorageJob`, and we have more than one job type (now also `RetrievalJob`), this had some important cascading changes in code structure. Also, got greedy and ended up doing some other good changes.

Here's some summary of things:
- The Docker image of Powergate now also includes the `pow` CLI. This is helpful when you run a bash inside the Docker container, since you also have the `pow` to run CLI commands against `powd`. Is very useful to debug stuff.
- The _HotConfig_ part of _StorageConfig_, now has an extra attribute _UnfreezeMaxPrice_ which the user can set to provide the max price to pay for unfreezing the data from the FIlecoin network. This is still not supported in Lotus, but will come soon in the API that supports IPLD selectors... so getting a bit ahead.
- `CidLogger` was changed to `JobLogger`, since now a Job can be a `StorageJob` or a `RetrievalJob`, so it can log things from both types of actions. The name `CidLogger` before implied _logging messages of stored Cids_, but that isn't accurate now that we have jobs that retrieve potentially partial data in a Deal.
- Now instead of having a single `JobStore`, we have a `StorageJobStore` and a `RetrievalJobStore`, as the repositories for persisting data of Jobs types.
- The _ColdConfig_ part of _StorageConfig_, has a new attribute _FastRetrieval_. This allows the user to set if want to make deals with the _FastRetrieval_ feature, which means the miner will hold an unsealed copy of the deals data to allow fast retrievals of deals data.
- The _DealsModule_ retrievals APIs switched to another signature that provides more information about whats going on in the retrieval. They now return a channel that produces events on how the deal retrieval is unfolding, such as providing how much funds are being sent and how many bytes are being transferred while the retrieval process is moving forward. This will be useful to log appropriate data in `RetrievalJobs`, so to give some information of what's happening while the retrievals are executing.
- Per the recommendation of Peter, we store also the `PieceCid` as information in `CidInfo`. In theory, `PayloadCid` would be enough, but considering some future changes that might happen this was a recommended thing to store handy.
- New feature in a FFS instance: We have a new feature to import deals to Powergate. That's to say: say I stored a Cid in 3 deals in Filecoin using the Lotus CLI. Now there's an `Import(...)` API to feed that data as if those things were done in Powergate. From that moment forward, you can keep changing the storage config and make Powergate do stuff. Imported deals are imported with a  _StorageConfig_ with _Hot Storage_ disabled (for obvious reasons that we're only importing deals data, not data). If later the user wants to unfreeze, it can push a new _StorageConfig_ as usual. TBH, this feature started as a potential solution to the "make arbitrary retrievals" feature but was a dead end. But looks like a useful feature anyway for other things.
- New feature in a FFS instance: `StartRetrieval(...)`. This is the original intention of the PR. You provide a `PayloadCid`, `PieceCid`, IPLD selector, and list of miners (plus some opts params), and it will schedule a `RetrievalJob` to make that happen and make the data available to retrieve. This PR doesn't have the core part of the retrieval yet, so it has the skeleton of most of the "Jobs" part, scheduler, etc. But all that is what triggered all these other changes mentioned. To be clear, this `StartRetrieval` isn't providing the feature yet.
- The deals renewal daemon now behaves the same as the repair daemon. It schedules Jobs to check for deals of stored Cids that should be renewed. Before this, it did things separately of running Jobs... which I didn't like since we should always try to make a unique action for a stored Cid at the moment. Having exceptions to that rule might cause problems if deal renewal is being made for a Cid, and some processing of an updated _StorageConfig_ was done for the same Cid. Now that renewals are also done as auto-scheduled _StorageConfig_ jobs, all storage/renewal/repair work is in a main trunk: just execute jobs. This is something that was in my todo for a long time.
- Before we had an `Action` which basically was data that a Job needed to do its job. For example, for previous Job it had a stored Action that contained the _StorageConfig_ information, etc (since that isn't stored in the Job itself). Now the original `Action` was renamed to `StorageAction`, and there's a new `RetrievalAction`. In short: `StorageJobs` will retrieve their corresponding `StorageAction` (just as before `Job` had its corresponding `Action`), and the new `RetrievalJobs` would have their corresponding `RetrievalAction`. Again, I chose to not use the same single `Action` for both things since is confusing, and depending on which kind of job you want to execute, data needed is different. So this is cleaner. If in the future we have new job types, then they'll have their own new action types with the data they need for executing.
- A new component was made which contains which Cids had enabled repair or renew flags. Before this was resolved by looking at _StorageConfigs_ which was confusing and a bit error-prone. Now all Cids that need some kind of background tracking, are registered in this new "Tracker" component which makes the work of the repair and renews daemons easier.

All these changes didn't have breaking changes at the gRPC API levels. There're some new attributes in protos, but they aren't breaking. Since the originally intended arbitrary retrieval isn't ready yet, all the gRPC/CLI/etc part of the work isn't done. So from the perspective of a user, this update is a breaking change at the datastore level. Their apps or CLI usage is unchanged. In fact, it can be noticed that all the integration tests were unchanged and working.

To be clear, the arbitrary retrievals part isn't ready yet... which is a bit funny since was the original intention of the PR. But it already set up the stage to focus on that work having all the necessary changes, refactors and skeleton ready.

I think the above bullet points provide a good summary of changes, so won't be adding comments to code... I don't like making these dense reviews, so I'd recommend to take it light and have handy the bullet points to see the intention of changes.


asutula (approved):
Wow great updates! The PR description was really helpful, and that all sounds good and makes sense. Gave the code changes a decent read through, but so much going on there... let's get this merged and start using it.

PR #592, opened 2020-09-15
support Mongo as underlying datastore

### LINKED ISSUE #591: datastore: support go-ds-mongo
Now that we created an implementation of `Datastore`, `TxnDatastore` and other siblings backed by MongoDB, provide necessary to choose between Badger or Mongo as the backing store.

jsign:
Two new flags are available: `--mongouri`, `--mongodb`.
If `--mongouri` is set, Powergate will use MongoDB as the underlying store for all `github.com/ipfs/go-datastore` interfaces. if that's the case `--mongodb` is mandatory to indicate the name of the MongoDB database it should use.

If `--mongouri` is empty, it will use Bager as usual (default).

Closes #591


asutula (approved):
Nice!

PR #598, opened 2020-09-17
Get storage job API

### LINKED ISSUE #594: Feature request: current job status
As opposed to using `watchJobs` and thus keeping a function running, I'd like to just get an instantaneous snapshot of a job state that I can check periodically. Is something like this possible?

Thank you :pray:

asutula:
Pretty self explanatory. Closes #594


ffs/api/api_jobs.go
```diff
@@ -8,6 +8,18 @@ import (
 	"github.com/textileio/powergate/ffs/scheduler"
 )
 
+// GetStorageJob returns the current state of the specified job.
+func (i *API) GetStorageJob(ctx context.Context, jid ffs.JobID) (ffs.StorageJob, error) {
+	job, err := i.sched.GetJob(jid)
+	if err == scheduler.ErrNotFound {
+		return ffs.StorageJob{}, fmt.Errorf("job id %s not found", jid)
```
asutula, line 15: Having this urge to only return machine readable custom error types so that the rpc layer can check what error it is and start returning proper grpc error codes, and do the nice string wrapped error messages at that level. It's tough because I know this file would be considered the public API for a go developer embedding Powergate in their program. But so far, the number of developers we have doing that is 0, and this is making it hard to return proper errors in the RPC layer where we have lots of developers using that API.

ffs/integrationtest/integrationtest.go
```diff
@@ -182,8 +182,18 @@ func NewFFSManager(t require.TestingT, ds datastore.TxnDatastore, clientBuilder
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	job, err := fapi.GetStorageJob(ctx, jid)
+	require.NoError(t, err)
+	require.Contains(t, statuses, job.Status)
+	return job
+}
+
+// RequireEventualJobState watches a Job for a desired status.
+func RequireEventualJobState(t require.TestingT, fapi *api.API, jid ffs.JobID, status ffs.JobStatus) ffs.StorageJob {
```
asutula, line 196: Most file changes in this PR are because of this rename.

jsign (approved):

PR #600, opened 2020-09-18
Better release builds

asutula:
Main goal here is to funnel _all_ builds through the `Makefile`. This has some advantages:
- Good scripting code isn't locked away in just the GH action config files and can be used for any purpose from the `Makefile`
- We can use the Bingo makefile helper variables to make sure that the correct version of any build tool is installed in GOPATH and used.
- Takes advantage Bingo Makefile variables that ensure correct build tools are used without polluting local git state
- This means no more `-dirty` git status in the build info
- Simplifies release builds using `make`
- Generation of all release artifacts is done with just `make build-releases` so this cleans up GH action configuration a lot
- Publishing release artifacts for `powd` and `powbench` now because it's easy now that building and publishing releases is a reusable/parameterized workflow.


.bingo/gomplate.mod
```diff
@@ -0,0 +1,5 @@
+module _ // Auto generated by https://github.com/bwplotka/bingo. DO NOT EDIT
+
+go 1.14
+
+require github.com/hairyhenderson/gomplate/v3 v3.8.0 // cmd/gomplate
```
asutula, line 5: Using Gomplate to create the install script for the different release artifacts.

.github/workflows/release.yml
```diff
@@ -52,46 +52,14 @@ jobs:
...
-              if [ "${EXT}" == ".exe" ]; then
-                  zip pow_${VERSION}_${OSARCH}.zip LICENSE README.md pow${EXT}
-                  mv pow_${VERSION}_${OSARCH}.zip ../${OUT}/
-              else
-                  tar -czvf pow_${VERSION}_${OSARCH}.tar.gz LICENSE README.md install pow
-                  mv pow_${VERSION}_${OSARCH}.tar.gz ../${OUT}/
-              fi
-          done
-          cd .. && rm -rf tmp
-          echo $(ls ./release/cli)
```
asutula, line 90: All this logic is now in the `Makefile` and is simplified there.

.github/workflows/release.yml
```diff
@@ -52,46 +52,14 @@ jobs:
...
+          POW_VERSION=${GITHUB_REF##*/} make build-releases
```
asutula, line 57: Simple!

Dockerfile
```diff
@@ -6,10 +6,8 @@ WORKDIR /app
 COPY go.mod go.sum ./
 RUN go mod download
 COPY . .
-RUN go build -modfile .bingo/govvv.mod -o=. github.com/ahmetb/govvv
-RUN pkg="$(go list ./buildinfo)" && govvvflags="$(./govvv -flags -pkg $pkg)" && CGO_ENABLED=0 GOOS=linux go build -ldflags="$govvvflags" -o powd cmd/powd/main.go
-RUN pkg="$(go list ./buildinfo)" && govvvflags="$(./govvv -flags -pkg $pkg)" && CGO_ENABLED=0 GOOS=linux go build -ldflags="$govvvflags" -o pow cmd/pow/main.go
-
+RUN BUILD_FLAGS="CGO_ENABLED=0 GOOS=linux" make build-powd
+RUN BUILD_FLAGS="CGO_ENABLED=0 GOOS=linux" make build-pow
```
asutula, line 10: More simple!

Makefile
```diff
@@ -1,26 +1,73 @@
...
+	$(BUILD_FLAGS) go install -ldflags="${GOVVV_FLAGS}" ./cmd/pow
+.PHONY: install-pow
+
+install-powd: $(GOVVV)
+	$(BUILD_FLAGS) go install -ldflags="${GOVVV_FLAGS}" ./cmd/powd
+.PHONY: install-powd
+
+install-powbench: $(GOVVV)
+	$(BUILD_FLAGS) go install -ldflags="${GOVVV_FLAGS}" ./cmd/powbench
+.PHONY: install-powbench
```
asutula, line 39: Renamed "build" to "install" so we can use "build" for another purpose above.

Makefile
```diff
@@ -1,26 +1,73 @@
...
+	for release in *; do \
+		cp ../../LICENSE ../../README.md $${release}/; \
+		if [[ $${release} != *"windows"* ]]; then \
+  		POW_FILE=$(2) $(GOMPLATE) -f ../../dist/install.tmpl -o "$${release}/install"; \
+			tar -czvf ../dist/$${release}.tar.gz $${release}; \
+		else \
+			zip -r ../dist/$${release}.zip $${release}; \
+		fi; \
+	done
+endef
```
asutula, line 54: Reusable function that builds all the releases and tar/zip files

Makefile
```diff
@@ -1,26 +1,73 @@
...
+
+build-pow-release: $(GOX) $(GOVVV) $(GOMPLATE)
+	$(call gen_release_files,./cmd/pow,pow,"linux/amd64 linux/386 linux/arm darwin/amd64 windows/amd64")
```
asutula, line 57: Calling that function, pass in the package, desired executable name, and os/arch you want to build

Makefile
```diff
@@ -1,26 +1,73 @@
...
+
+build-powd-release: $(GOX) $(GOVVV) $(GOMPLATE)
+	$(call gen_release_files,./cmd/powd,powd,"linux/amd64 darwin/amd64")
+.PHONY: build-release
+
+build-powbench-release: $(GOX) $(GOVVV) $(GOMPLATE)
+	$(call gen_release_files,./cmd/powbench,powbench,"linux/amd64 linux/386 linux/arm darwin/amd64 windows/amd64")
+.PHONY: build-release
+
+build-releases: build-pow-release build-powd-release build-powbench-release
```
asutula, line 68: build all the releases. added support for `powd` and` powbench` because it is now trivial.

dist/install.tmpl
```diff
@@ -8,16 +8,16 @@ set -Eeuo pipefail
 
 INSTALL_DIR="$(dirname "$0")"
 
-pow="$INSTALL_DIR/pow"
+file="$INSTALL_DIR/{{ .Env.POW_FILE }}"
```
asutula, line 11: this is not a Gomplate template.

.github/workflows/release.yml
jsign: Nice... easier to test changes locally than artificially running releases.

jsign (approved):
LGTM, very nice!

PR #606, opened 2020-09-22
support urls in ffs stage command

b5:
this makes `pow ffs stage https://url-to-file` work, expanding the potential sources for staging into powergate. I've smoke-tested with a 250Mb file stored in an S3 bucket, can confirm it's very nice to have.

Urls *only* write to files (no folder support), which I think covers most url-oriented use-cases.

One caveat: if a user has a file that starts with 'http', it'll be a problem. The easiest fix is to refer to the file with a `./` prefix, such as: `./http`


jsign (approved):
Looks good!

Do you mind doing `make pow-docs` to regenerate the docs and commit?
That would fix the CI pipeline red flag.

b5:
Sure thing. `make docs-pow` run, branch rebased. Should be good to go

PR #608, opened 2020-09-22
Enable release drafter

asutula:
This will maintain a draft release as PRs are merged. I enabled all the bells and whistles that will autosuggest the next version number and categorize the release notes based on PR tags. 

Supported tags for version number calculation are:
- `major`
- `minor`
- `patch`

Supported categories and the tags to put something into that category are:
- Features
  - `feature`
  - `enhancement`
- Bug Fixes
  - `fix`
  - `bugfix`
  - `bug`
- Maintenance
  - `chore`

Those can be changed as needed.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,37 @@
+name-template: "v$RESOLVED_VERSION 🌈"
+tag-template: "v$RESOLVED_VERSION"
+categories:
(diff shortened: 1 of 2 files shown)


asutula:
So this PR will bump the minor version and be categorized as Maintenance because of the tags I added.

jsign (approved):

PR #609, opened 2020-09-23
Bump default autofund amount & fail fast if not enough balance for autofunding

jsign:
PR:
- Bump default autofunding amount to what was decided for SR2.
- Fail creating new addresses if they will most probably fail if the master address doesn't have enough balance.

### DIFF
--- a/cmd/powd/main.go
+++ b/cmd/powd/main.go
@@ -295,7 +295,7 @@ func setupFlags() error {
 	pflag.String("repopath", "~/.powergate", "Path of the repository where Powergate state will be saved.")
 	pflag.Bool("devnet", false, "Indicate that will be running on an ephemeral devnet. --repopath will be autocleaned on exit.")
 	pflag.String("ipfsapiaddr", "/ip4/127.0.0.1/tcp/5001", "IPFS API endpoint multiaddress. (Optional, only needed if FFS is used)")
-	pflag.Int64("walletinitialfund", 4000000000000000, "FFS initial funding transaction amount in attoFIL received by --lotusmasteraddr. (if set)")
+	pflag.Int64("walletinitialfund", 250_000_000_000_000_000, "FFS initial funding transaction amount in attoFIL received by --lotusmasteraddr. (if set)")
 	pflag.String("gatewayhostaddr", "0.0.0.0:7000", "Gateway host listening address.")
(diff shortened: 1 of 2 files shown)


cmd/powd/main.go
```diff
@@ -295,7 +295,7 @@ func setupFlags() error {
 	pflag.String("repopath", "~/.powergate", "Path of the repository where Powergate state will be saved.")
 	pflag.Bool("devnet", false, "Indicate that will be running on an ephemeral devnet. --repopath will be autocleaned on exit.")
 	pflag.String("ipfsapiaddr", "/ip4/127.0.0.1/tcp/5001", "IPFS API endpoint multiaddress. (Optional, only needed if FFS is used)")
-	pflag.Int64("walletinitialfund", 4000000000000000, "FFS initial funding transaction amount in attoFIL received by --lotusmasteraddr. (if set)")
+	pflag.Int64("walletinitialfund", 250_000_000_000_000_000, "FFS initial funding transaction amount in attoFIL received by --lotusmasteraddr. (if set)")
```
jsign, line 298: Was decided to default to 0.25 FIL in SR2.

wallet/module/wallet.go
```diff
@@ -92,6 +95,13 @@ func (m *Module) NewAddress(ctx context.Context, typ string) (string, error) {
 	}
 
 	if m.masterAddr != address.Undef {
+		balance, err := client.WalletBalance(ctx, m.masterAddr)
+		if err != nil {
+			return "", fmt.Errorf("getting balance from master addr: %s", err)
+		}
+		if balance.LessThan(types.BigAdd(types.BigInt{Int: m.iAmount}, types.NewInt(feeThreshold))) {
```
jsign, line 102: Fail fast if most prob the txn will fail if we don't have enough funds. The txn is sent async since is a blocking process in the Lotus API now since has to estimate feeds and that takes time.

We still have two improvements to do here:
- Instead of using `feeThreshold` as a constant, we could craft a more dynamic calculation of estimated fees. (#610)
- Think about a possibly more reliable way to send the funding txn. If for some wild reason the sent txn failed, we'd get some unwanted result of a 0 balance addr. So looks like we should have a better tracking mechanism about this... not sure though what might be a good solution apart from just retrying with some exponential backoff. (#611)

asutula (approved):

PR #613, opened 2020-09-23
Publish Docker image on release publish

asutula:
This should work with the new draft/publish flow.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,9 +1,10 @@
 name: Docker Image
 on:
-  create:
+  release:
+    types: [published]
   push:
     branches:
-        - master
+      - master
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -1,9 +1,10 @@
 name: Docker Image
 on:
-  create:
```
asutula, line 3: Let me know if you still want this @jsign. `create` is triggered any time someone creates a branch or tag. Seems like the most important tags are releases, and those are now covered by `release:` so I removed this.

jsign: LGTM.

jsign (approved):
LGTM.

PR #614, opened 2020-09-24
SR2 miner selector & safer default StorageConfig config

jsign:
This PR creates a new `MinerSelector` implementation, which executes the planned strategy for selecting miners in SR2.
In a nutshell, when this miner selector is used:
- A remote .json file is automatically fetched and parsed which contains which miners are going to be used to make deals.
- An example of that file is [here](https://raw.githubusercontent.com/filecoin-project/slingshot/0b07ab9da98bc3c8f45827edfb40f3f4cdd0d49b/miners/list.json)
- What Powergate will do is to select `Amount` random miners from each miners bucket. 

This allows to:
- Update the miner list in `github.com/filecoin-project/slingshot` repository, and automatically update every running Powerate without restarts. Every time a list of miners needs to be selected, it will fetch the current `list.json` in the `master` branch of that repo.
- If we think is a good/bad idea to increase/decrease the replication factor, it can be easily done by increase/decreasing the `Amount` attribute of each miners bucket in the json file.

By default now we configure Poweragate to use this miner selector strategy. A new flag was included if someone wants to keep using the previous one; so at the end of the day is configurable. The URL where this JSON will be fetched is also configurable via flags/envs, so if someone wants to use their own JSON file, that's OK too.

So in short, the default now is to have the right configuration for the SR2 contest.

Some other lateral changes:
- The default _StorageConfig_ has _HotStorage_ disabled as a more reasonable configuration to avoid disk usage for SR2.
- The `AddTimeout` is hardcoded to a bigger value ignoring the _StorageConfig_. This is a temporal solution until we have a bit more control to make general _StorageConfig_ changes in the Hub.
- `lotus-devnet` was updated to v0.7.2.
- Default deal duration decreased to 6 months as decided today in SR2 call.
- Keep using `reputation` minerselector, since SR2 want to hold a bit using the new one created in this PR.


api/server/server.go
```diff
@@ -531,3 +538,22 @@ func createDatastore(conf Config) (datastore.TxnDatastore, error) {
 	}
 	return ds, nil
 }
+
+func getMinerSelector(conf Config, rm *reputation.Module, ai *ask.Runner, cb lotus.ClientBuilder) (ffs.MinerSelector, error) {
+	var ms ffs.MinerSelector
+	var err error
+
+	switch conf.MinerSelector {
```
jsign, line 546: `MinerSelector` configuration depending on flags/envs.

cmd/powd/main.go
```diff
@@ -301,6 +308,8 @@ func setupFlags() error {
 	pflag.String("maxminddbfolder", ".", "Path of the folder containing GeoLite2-City.mmdb")
 	pflag.String("mongouri", "", "Mongo URI to connect to MongoDB database. (Optional: if empty, will use Badger)")
 	pflag.String("mongodb", "", "Mongo database name. (if --mongouri is used, is mandatory")
+	pflag.String("ffsminerselector", "sr2", "Miner selector to be used by FFS: 'sr2', 'reputation'")
```
jsign, line 311: `MinerSelector` can be:
- `sr2`: the default now. Does what was commented in PR description.
- `reputation`: The previous miner selector that uses the reputation index, and still directly targets peter's miners as trusted.

ffs/manager/manager.go
```diff
@@ -31,7 +31,7 @@ var (
 	// recover its last configured default StorageConfig from the datastore.
 	zeroConfig = ffs.StorageConfig{
 		Hot: ffs.HotConfig{
-			Enabled: true,
+			Enabled: false,
```
jsign, line 34: Decided with @asutula  to avoid disk usage explosion. 
If hte user want's HotStorage enabled, should be set explicitly on pushing a config.

ffs/minerselector/sr2/sr2.go
```diff
@@ -0,0 +1,143 @@
+package sr2
```
jsign, line 1: Summary of the selector:
- Fetch the json file
- Parse it
- Get covered on some possibly weird values (amount < 0, etc)
- Use the ask cache index as much as possible. Some miners might not be there if they have 0 power. So instead of changing the ask index to include also miners with 0 power (which are a lot and would affect index building times), I decided that if a miner in this trusted list hasn't a cached ask, we'd do that directly here.

ffs/minerselector/sr2/sr2.go
```diff
@@ -0,0 +1,149 @@
...
+	_, err := ms.getMiners()
+	if err != nil {
+		return nil, fmt.Errorf("verifying sr2 url content: %s", err)
+	}
+
+	return ms, nil
+}
+
+// GetMiners returns miners from SR2.
+func (ms *MinerSelector) GetMiners(n int, f ffs.MinerSelectorFilter) ([]ffs.MinerProposal, error) {
```
jsign, line 56: Important fact. This MinerSelector will ignore `n` and `f`. 
The replication factor will be the sum of `Amount` of buckets, and not filtering is done.

So this is a very aggressive miner selector kind of forcing what was decided in SR2. No custom _StorageConfig_ will change this fact, unless the user runs with `--ffsminerselector=reputation`.

ffs/scheduler/scheduler_storage.go
```diff
@@ -202,7 +202,9 @@ func (s *Scheduler) executeHotStorage(ctx context.Context, curr ffs.CidInfo, cfg
 		return ffs.HotInfo{Enabled: false}, nil
 	}
 
-	sctx, cancel := context.WithTimeout(ctx, time.Second*time.Duration(cfg.Ipfs.AddTimeout))
+	// ToDo: this is a hot-fix to force a big timeout until we have a
+	// migration tool to make this tunable again.
+	sctx, cancel := context.WithTimeout(ctx, time.Second*300)
```
jsign, line 207: See [this](https://textile-public.slack.com/archives/GQRACAG3G/p1600958777011300).

asutula (approved):
Awesome. Just one question for my own understanding.

ffs/minerselector/sr2/sr2.go
jsign: Yes, is ignored. Now will be controled in this remote file now that `--ffsminerselector=sr2` default miner selector.
If someone wants to come back to not being ignored, they could specify `--ffsminerselector=reputation` and things will work as before.

PR #627, opened 2020-09-29
add flag maxrecvmsgsize to set MaxRecvMsgSize of gRPC server

ysh0566:
Add flag maxrecvmsgsize to set MaxRecvMsgSize of gRPC server so that grpc client can upload bigger than 4mb files.

### DIFF
--- a/cmd/powd/main.go
+++ b/cmd/powd/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/textileio/powergate/buildinfo"
 	"github.com/textileio/powergate/util"
 	"go.opencensus.io/plugin/runmetrics"
+	"google.golang.org/grpc"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


asutula:
Nice! Let me ask a couple questions to make sure we're on the same page:

1. Are you hitting a size limit using the Go Powergate client or Javascript Powergate client?
2. The `Stage` function that allows uploading of data that will be added to Powergate is a _streaming_ gRPC function... is it your understanding that this flag applies to each individual request ("chunk") of the stream, or the total amount of data received by the server for a streaming request (made up of many individual requests)?

If you are using the Javascript client, that client isn't doing any chunking of data, it's just sending the entire file in a single request. And if this flag applies to individual stream requests (which I'm guess it does), I think the default of 4MB requests is fine. The fix would be to make the Javascript client chunk the data.

ysh0566:
> Nice! Let me ask a couple questions to make sure we're on the same page:
> 
> 1. Are you hitting a size limit using the Go Powergate client or Javascript Powergate client?
> 2. The `Stage` function that allows uploading of data that will be added to Powergate is a _streaming_ gRPC function... is it your understanding that this flag applies to each individual request ("chunk") of the stream, or the total amount of data received by the server for a streaming request (made up of many individual requests)?
> 
> If you are using the Javascript client, that client isn't doing any chunking of data, it's just sending the entire file in a single request. And if this flag applies to individual stream requests (which I'm guess it does), I think the default of 4MB requests is fine. The fix would be to make the Javascript client chunk the data.

ysh0566:
I think you are right. The better way is making the Javascript client chunk the data.

asutula:
Ok, I can take a look at that today.

asutula:
Ok @ysh0566 the js client has been updated. It chunks data now. I've tested it with a 330MB file, no problem. Also updates all dependencies

https://github.com/textileio/js-powergate-client/releases/tag/v2.0.1

PR #633, opened 2020-09-30
FFS scheduler improvements and knobs

jsign:
- Include a `CreatedAt` field in `StorageJob`. This is used to sort queued Jobs to dequeue them in order. When we used badger, this was the default query order (insertion order), but now we should be more explicit since can't have that guarantee in multiple `Datastore` implementations. For existing `StorageJob`, they'll have a 0 unix-time, they'll have the highest priority until are all done.
- `--schedmaxparallel`: Configurable maximum amount of jobs to run in parallel.
- `--dealwatchpollduration`: frequency in which in-progress deals are queried to discover a new status change. Increasing this number decreases the total load on calling Lotus APIs, so allow more parallel Jobs.
- `--ffsdealfinalitytimeout`: If after this time a deal doesn't report a status change, will be considered failed. This is a solution for getting unstuck from Lotus reporting `DealSealing` forever even if the miner abandoned/errored the deal.
- When printing status changes in `pow ffs log`, include the Deal ID as soon as is available and not only when the deal reach success.
- Accommodate UX to consider the replication factor if SR2 miner selector is used. A bit tricky since this miner selector violates the _StorageConfig_. Found an easy solution explained in the review comments.
- Prints scheduler stats (rate-limiting situation, number of queued and executing jobs) every time the queue is evaluated.

The idea of the first two mentioned flags is to tune the load the scheduler creates on the Lotus node. The less frequent each deal is querying for a status change, the more parallel jobs we could run in theory. 

Of course, this doesn't account for the initial cost of running a Job which is sending the data... that is an unavoidable load. But after that part happens and the deal is Sealing for hours, we can now tune how much load a single deal creates on the Lotus node. Hope makes sense.

The finality timeout flag, is just to avoid that a deal will be watched forever if the Lotus node somewhat gets stuck in some deal status. A deal watched forever is a problem since it eats one unit of the max parallel Jobs running. Whenever Lotus creates some API to cancel tracking a deal, we also should cancel it in the Lotus side (because if we don't do this, we would have forever Sealing things in the Lotus node; but currently isn't possible to do it)

The current default values of the new flags are:
- MaxParallel: 1000 (x10 compared to before)
- DealWatchPollDuration: 5 min (x0.1 compared to before)
- FFSDealFinalityTimeout: 24hs (Not sure about this number, may be a bit bigger)

Also, some deps updating.


api/server/server.go
```diff
@@ -193,7 +196,10 @@ func NewServer(conf Config) (*Server, error) {
 	if err != nil {
 		return nil, fmt.Errorf("creating faults index: %s", err)
 	}
-	dm, err := dealsModule.New(txndstr.Wrap(ds, "deals"), clientBuilder, deals.WithImportPath(filepath.Join(conf.RepoPath, "imports")))
+	if conf.Devnet {
+		conf.DealWatchPollDuration = time.Second
```
jsign, line 200: If we're in Devnet, ignore the configured value and just use 1 second.

api/server/server.go
```diff
@@ -224,7 +230,12 @@ func NewServer(conf Config) (*Server, error) {
 	if err != nil {
 		return nil, fmt.Errorf("creating coreipfs: %s", err)
 	}
-	sched, err := scheduler.New(txndstr.Wrap(ds, "ffs/scheduler"), l, hs, cs)
+
+	var sr2rf func() (int, error)
+	if ms, ok := ms.(*sr2.MinerSelector); ok {
+		sr2rf = ms.GetReplicationFactor
+	}
```
jsign, line 237: This should be temporary... but is for UX reasons, will explain later in review.

deals/module/deals.go
```diff
@@ -331,7 +332,7 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.S
 			select {
 			case <-ctx.Done():
 				return
-			case <-time.After(util.AvgBlockTime):
+			case <-time.After(m.pollDuration):
```
jsign, line 335: When watching deals, use the configured polling duration instead of a fixed `AvgBlockTime` (30s). Despite this will give a very fast status updating, considering sealing takes hours doesn't make sense to be so demanding.
Now the default is 5 min, which actually is quite low too... but now can be easily changed via flags/envs.

ffs/filcold/filcold.go
```diff
@@ -268,36 +269,52 @@ func (fc *FilCold) makeDeals(ctx context.Context, c cid.Cid, size uint64, cfgs [
...
-				StartEpoch:      di.StartEpoch,
-				EpochPrice:      di.PricePerEpoch,
+	var last deals.StorageDealInfo
+Loop:
+	for {
+		select {
+		case <-time.After(timeout):
+			msg := fmt.Sprintf("DealID %d with miner %s tracking timed out after waiting for %.0f hours.", last.DealID, last.Miner, timeout.Hours())
+			fc.l.Log(ctx, msg)
+			return ffs.FilStorage{}, ffs.DealError{ProposalCid: proposal, Message: msg}
```
jsign, line 287: So now in `WaitForDeal` we keep watching for the channel to report new status changes of a deal. But we also put a cap of the amount of time we wait between two status changes.

If for example, last reported status was `DealSealing` (which is usually the one that takes more time to change to a next status), and we didn't receive any status update in `timeout` duration, then we'll consider this deal abandoned/errored/etc.

This timeout is what the flag/env names `--ffsdealfinalitytimeout`: _A deal can't be in a non-final status for more than X hours_.

ffs/filcold/filcold.go
```diff
@@ -268,36 +269,52 @@ func (fc *FilCold) makeDeals(ctx context.Context, c cid.Cid, size uint64, cfgs [
...
+				log.Errorf("deal %d failed with state %s: %s", di.DealID, storagemarket.DealStates[di.StateID], di.Message)
+				fc.l.Log(ctx, "DealID %d with miner %s failed and won't be active on-chain: %s", di.DealID, di.Miner, di.Message)
+
+				return ffs.FilStorage{}, ffs.DealError{ProposalCid: di.ProposalCid, Miner: di.Miner, Message: di.Message}
+			default:
+				if di.DealID != 0 {
+					fc.l.Log(ctx, "Deal %d with miner %s changed state to %s", di.DealID, di.Miner, storagemarket.DealStates[di.StateID])
+				} else {
+					fc.l.Log(ctx, "Deal with miner %s changed state to %s", di.Miner, storagemarket.DealStates[di.StateID])
+				}
```
jsign, line 316: Before we printed the Deal ID only when the deal reached success.
Now we print it in any intermediary status that already has a value > 0, which usually is a status where the deal is on-chain but not yet active.

This is useful for debugging stuff, or talking with the miner asking: "Hey, what is happening with Deal X? Is still sealing". Since before this ID was only printed on success, it was a bit hard to do this.

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -87,29 +98,17 @@ func (s *Store) Finalize(jid ffs.JobID, st ffs.JobStatus, jobError error, dealEr
 func (s *Store) Dequeue() (*ffs.StorageJob, error) {
```
jsign, line 98: Now we keep a `queued` slice as a cache of all pending Jobs. 
This slice is ordered by `CreatedAt` field, so dequeuing is always easy and also avoids querying Datastore.

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -87,29 +98,17 @@ func (s *Store) Finalize(jid ffs.JobID, st ffs.JobStatus, jobError error, dealEr
...
-		var j ffs.StorageJob
-		if err := json.Unmarshal(r.Value, &j); err != nil {
-			return nil, fmt.Errorf("unmarshalling job: %s", err)
-		}
-		_, ok := s.executingCids[j.Cid]
-		if j.Status == ffs.Queued && !ok {
-			j.Status = ffs.Executing
-			if err := s.put(j); err != nil {
+
+	for _, job := range s.queued {
```
jsign, line 102: So the logic is: start from older to newer, and execute the oldest Job with a data Cid that isn't already executing.
(Recall that we only execute on job per data Cid at a time).

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -280,12 +298,38 @@ func (s *Store) put(j ffs.StorageJob) error {
 	if err := s.ds.Put(makeKey(j.ID), buf); err != nil {
 		return fmt.Errorf("saving to datastore: %s", err)
 	}
-	s.notifyWatchers(j)
+
+	// Update executing cids cache.
 	if j.Status == ffs.Executing {
 		s.executingCids[j.Cid] = j.ID
```
jsign, line 304: If the updated Job is `Executing`, should be in the executing cache list.

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -280,12 +298,38 @@ func (s *Store) put(j ffs.StorageJob) error {
...
-	} else if j.Status == ffs.Failed || j.Status == ffs.Success || j.Status == ffs.Canceled {
-		delete(s.executingCids, j.Cid)
+	} else {
+		execJobID, ok := s.executingCids[j.Cid]
+		// If the executing job is not longer executing,
+		// take out from executing cache.
+		if ok && execJobID == j.ID {
+			delete(s.executingCids, j.Cid)
+		}
```
jsign, line 311: If the updated Job isn't executing, no matter which this state is, shouldn't be in the executing cache.

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -280,12 +298,38 @@ func (s *Store) put(j ffs.StorageJob) error {
...
+	}
+
+	// Update queued cids cache.
+	if j.Status == ffs.Queued {
+		// Every new Queued job goes at the tail since has
+		// the biggest CreatedAt value.
+		s.queued = append(s.queued, j)
```
jsign, line 318: If the job status is `Queued` should be in the queued cache.

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -280,12 +298,38 @@ func (s *Store) put(j ffs.StorageJob) error {
...
+		delIndex := -1
+		for i, job := range s.queued {
+			if j.ID == job.ID {
+				delIndex = i
+				break
+			}
+		}
+		if delIndex != -1 {
+			s.queued = append(s.queued[:delIndex], s.queued[delIndex+1:]...)
+		}
```
jsign, line 329: If the updated Job isn't queued, this job shouldn't be in the queued cache.

ffs/scheduler/scheduler.go
```diff
@@ -235,9 +240,10 @@ func (s *Scheduler) run() {
 			log.Infof("scheduler daemon terminated")
 			return
 		case <-s.sd.evaluateQueue:
-			log.Debug("evaluating storage job queue...")
+			log.Infof("storage job push rate limit: %d/%d", len(s.sd.rateLim), cap(s.sd.rateLim))
+			stats := s.sjs.GetStats()
+			log.Infof("storage job total queued: %d, total executing: %d", stats.TotalQueued, stats.TotalExecuting)
```
jsign, line 245: Every time the job queue is evaluated to see if any queued job is runnable, we print some extra information like:
- How close we're to the rate limit
- queued jobs
- executing jobs

I preferred to be more verbose to include these things in logs to understand what's happening.

ffs/scheduler/scheduler_storage.go
```diff
@@ -74,6 +75,9 @@ func (s *Scheduler) push(iid ffs.APIID, c cid.Cid, cfg ffs.StorageConfig, oldCid
 	if err := s.sjs.Enqueue(j); err != nil {
 		return ffs.EmptyJobID, fmt.Errorf("enqueuing job: %s", err)
 	}
+	if jid := s.sjs.GetExecutingJob(c); jid != nil {
+		s.l.Log(ctx, "Job %s is already being executed for the same data, this job will be queued until it finishes or is canceled.")
+	}
```
jsign, line 80: If you push a new storageconfig for a Cid that already has an executing Job, we print an extra log line to explain that this job might take a while to get executed... since need the other job to finish (success, canceled, failed) before can be considered runnable.

This might help some people understand if they're in this situation.

ffs/scheduler/scheduler_storage.go
```diff
@@ -384,6 +388,12 @@ func (s *Scheduler) executeColdStorage(ctx context.Context, curr ffs.CidInfo, cf
 	// whatever extra deals we need to make that true.
 
 	// Do we need to do some work?
+	if s.sr2RepFactor != nil {
+		cfg.Filecoin.RepFactor, err = s.sr2RepFactor()
+		if err != nil {
+			return ffs.ColdInfo{}, nil, fmt.Errorf("getting SR2 replication factor: %s", err)
+		}
+	}
```
jsign, line 396: OK, a tricky thing about UX.

Without this change, the scheduler would always consider the replication factor in the StorageConfig to know what's the intentional replication factor. That isn't true if the SR2 miner selector is used.

What's more, in SR2 this value is dynamic since lives in a remote file. So the scheduler gets initialized with a `sr2RepFactor` function, that returns the current replication factor of the file, and overrides the StorageConfig with that value. If the SR2 miner selector isn't used then this function would be nil and will work as usual.

This logic will live until we take out this custom miner selector... so would be just deleting this code.

index/ask/runner/runner.go
```diff
@@ -261,15 +261,7 @@ func getMinerStorageAsk(ctx context.Context, api *apistruct.FullNodeStruct, addr
...
 	}
-	ribasMiners := []string{"t016303", "t016304", "t016305", "t016306", "t016309"}
-	var forceQueryAsk bool
-	for _, rm := range ribasMiners {
-		if addr.String() == rm {
-			forceQueryAsk = true
-			break
-		}
-	}
-	if !forceQueryAsk && power.MinerPower.RawBytePower.IsZero() {
```
jsign, line 272: Some extra things to remove about peter's miners.

ffs/scheduler/scheduler_storage.go
asutula: Not totally sure of the details, but isn't `RepFactor` normally (not in SR2) used as _input_ to the miner selector so the selector knows how many miners to select? In the case of SR2, the miner selector ignores the `RepFactor` input. Is this change just to force `RepFactor` to be consistent with what the miner selector will do anyways? That is a good thing to do for the sake of consistency and UX etc, but maybe I'm missing something else here?

ffs/filcold/filcold.go
jsign: Well, yes. Actually, both things are a bit equivalent in practice since mostly the Sealing state is the one that takes hours. The precise description is the first one, and since only Sealing takes hours then the second is true by practical means.

ffs/scheduler/scheduler_storage.go
jsign: The replication factor is also in the remote JSON file. What this does is doing an alignment of the `RepFactor` value in the `StorageConfig` such that the rest of the scheduler logic takes the correct decisions.

ffs/filcold/filcold.go
asutula: I guess I was seeing that the timeout doesn't get reset each time a new non-final status comes in. So that's why it seems like the second description is more accurate.

jsign: Every time the select is evaluated in the for look, `time.After(..)` returns a new timer with `timeout` duration.
So it gets reset on every received status change.

Makes sense?

asutula: Aaaahhh yea, got it. Cool thanks for clarifying.

PR #659, opened 2020-10-06
Deal Insight + New API Foundation

### LINKED ISSUE #386: ffs/manager: add auth token & unpublished api

asutula:
LOTS going on here. It's a ridiculous amount of changed files, but I commented on every file that contains consequential changes, and that number is large, but much smaller that 136. Focus on those commented files for sure.

- bubble up storage deal status events to somewhere in the stack where they have the chance to be consumed by a new API for viewing data storage state, can be returned along with API calls to `GetJob`,  and also sent as part of the event stream for watching jobs.
- new and useful apis for getting information about storage jobs and storage deals in real time
- new api for querying `StorageConfig` by job id
- lays foundation for a new and better API design that isn't tightly coupled to Powergate module implementations
- implements some new services and moves a couple old ones to this new api foundation to feel it out (will move everything and clean up the API in an upcoming release)
- adds admin token auth based on @jsign's original work on that
- moves ffs create and list to admin api along with some new apis for querying jobs
- #686 makes client return gRPC types and CLI mostly just print json encoded gRPC data
- #688 returns storage profile tokens along with ids in the admin api
- #689 renames `CidInfo` to `StorageInfo` and uses `CidInfo` in the API to summarize the state of a cid in the system.
- #690 cleans up and reorganizes the API, client and CLI

closes #386


api/client/ffs.go
```diff
@@ -791,10 +792,70 @@ func fromRPCStorageDealRecords(records []*rpc.StorageDealRecord) ([]deals.Storag
...
+			status = storagemarket.StorageDealUnknown
+		case rpc.StorageDealStatus_STORAGE_DEAL_STATUS_VALIDATING:
+			status = storagemarket.StorageDealValidating
+		case rpc.StorageDealStatus_STORAGE_DEAL_STATUS_VERIFYING_DATA:
+			status = storagemarket.StorageDealVerifyData
+		case rpc.StorageDealStatus_STORAGE_DEAL_STATUS_WAITING_FOR_DATA:
+			status = storagemarket.StorageDealWaitingForData
+		default:
+			return nil, fmt.Errorf("unkown deal state: %v", rpcRecord.DealInfo.Status.String())
+		}
```
asutula, line 854: Not important yet, ignore.

deals/module/deals.go
```diff
@@ -327,22 +327,29 @@ func (m *Module) Watch(ctx context.Context, proposals []cid.Cid) (<-chan deals.S
 	}
 	ch := make(chan deals.StorageDealInfo)
 	go func() {
-		defer close(ch)
+		client, cls, err := m.clientBuilder()
```
asutula, line 330: Goal of these changes is to make watching deals send a result immediately upon subscription instead of waiting `m.pollDuration` amount of time before getting the first event.

deals/module/deals.go
```diff
@@ -583,10 +590,10 @@ func (m *Module) eventuallyFinalizeDeal(dr deals.StorageDealRecord, timeout time
...
 				}
 				return
-			} else if info.StateID == storagemarket.StorageDealProposalNotFound ||
-				info.StateID == storagemarket.StorageDealProposalRejected ||
-				info.StateID == storagemarket.StorageDealFailing {
-				log.Infof("proposal cid %s failed with state %s, deleting pending deal", util.CidToString(info.ProposalCid), storagemarket.DealStates[info.StateID])
+			} else if info.Status == storagemarket.StorageDealProposalNotFound ||
+				info.Status == storagemarket.StorageDealProposalRejected ||
+				info.Status == storagemarket.StorageDealFailing {
+				log.Infof("proposal cid %s failed with state %s, deleting pending deal", util.CidToString(info.ProposalCid), storagemarket.DealStates[info.Status])
```
asutula, line 596: Not important for the feedback I'm trying to gather now. Long story short, was experimenting with using the `storagemarket` deal state enums throughout the stack since they will be a part of data exposed by the API. Not sure if using that enum directly is good or bad, or we could mirror it with our own enum type.

ffs/scheduler/scheduler.go
```diff
@@ -62,6 +63,9 @@ type Scheduler struct {
 	ctx      context.Context
 	cancel   context.CancelFunc
 	finished chan struct{}
+
+	jobStatusCache     map[ffs.APIID]map[cid.Cid]map[cid.Cid]ffs.FilStorage
```
asutula, line 67: With maintain this cache of in process deal state. It maps FFS instance id > data cid > proposal cid > `FilStorage` state (which now includes the `Status`). This cache can be queried to build responses about cid storage state or job state.

ffs/scheduler/scheduler.go
```diff
@@ -422,7 +428,29 @@ func (s *Scheduler) executeQueuedStorage(j ffs.StorageJob) {
...
+			s.jobStatusCache[j.APIID][j.Cid][update.ProposalCid] = update
+			log.Infof("\n%v", s.jobStatusCache)
+			s.jobStatusCacheLock.Unlock()
+		}
+		s.jobStatusCacheLock.Lock()
+		delete(s.jobStatusCache[j.APIID], j.Cid)
+		s.jobStatusCacheLock.Unlock()
+		log.Info("ZZZ -- Done receiving updates.")
+	}()
+	info, dealErrors, err := s.executeStorage(ctx, a, j, dealUpdates)
```
asutula, line 453: Here's the big change: We pass in an channel of `ffs.FilStorage` here. It internally gets passed down into the lower level plumbing that executes cold storage and waits for deals to become final. That lower level code now publishes events on this channel so we can subscribe to them here.

ffs/scheduler/scheduler.go
```diff
@@ -422,7 +428,29 @@ func (s *Scheduler) executeQueuedStorage(j ffs.StorageJob) {
 
 	// Execute
 	s.l.Log(ctx, "Executing job %s...", j.ID)
-	info, dealErrors, err := s.executeStorage(ctx, a, j)
+	dealUpdates := make(chan ffs.FilStorage)
+	go func() {
+		for update := range dealUpdates {
```
asutula, line 433: Handle deal updates by updating our in memory cache, and doing some debug logging for now.

ffs/scheduler/scheduler.go
```diff
@@ -422,7 +428,29 @@ func (s *Scheduler) executeQueuedStorage(j ffs.StorageJob) {
...
+			_, ok = s.jobStatusCache[j.APIID][j.Cid]
+			if !ok {
+				s.jobStatusCache[j.APIID][j.Cid] = map[cid.Cid]ffs.FilStorage{}
+			}
+			s.jobStatusCache[j.APIID][j.Cid][update.ProposalCid] = update
+			log.Infof("\n%v", s.jobStatusCache)
+			s.jobStatusCacheLock.Unlock()
+		}
+		s.jobStatusCacheLock.Lock()
+		delete(s.jobStatusCache[j.APIID], j.Cid)
```
asutula, line 449: Once the deals are all complete for this job, delete the state from the cache.

ffs/scheduler/scheduler.go
asutula: The other key thing here:

We now have events about how individual deals are progressing for a specific FFS instance and data cid here in `Scheduler`. These events are available, potentially, in the same scope as events that trigger updates to listeners of watching Jobs... Currently, `Scheduler` calls methods on `sjstore`, and those calls trigger updates to Job watchers. We could lift those event triggers and job subscriber management up out our `sjstore` to `Scheduler` and that would give us the opportunity to send job updates triggered by both `sjstore` methods calls as well as related deal events.

ffs/types.go
```diff
@@ -460,6 +461,8 @@ type FilInfo struct {
 type FilStorage struct {
 	// ProposalCid of the deal.
 	ProposalCid cid.Cid
+	// Status is the current status of the Filecoin storage deal
+	Status storagemarket.StorageDealStatus
```
asutula, line 465: This struct previously represented the final storage state for a deal proposal. I added `Status` to it so we can also use it to represent incremental state while the deal is progressing. Other option is to create a separate stuct for that, but they will be very similar.

ffs/filcold/filcold.go
```diff
@@ -269,9 +269,12 @@ func (fc *FilCold) makeDeals(ctx context.Context, c cid.Cid, size uint64, cfgs [
 // If the deal finishes successfully it returns a FilStorage result.
 // If the deal finished with error, it returns a ffs.DealError error
 // result, so it should be considered in error handling.
-func (fc *FilCold) WaitForDeal(ctx context.Context, c cid.Cid, proposal cid.Cid, timeout time.Duration) (ffs.FilStorage, error) {
+func (fc *FilCold) WaitForDeal(ctx context.Context, c cid.Cid, proposal cid.Cid, timeout time.Duration, dealUpdates chan<- ffs.FilStorage) (ffs.FilStorage, error) {
```
asutula, line 272: The `chan<- ffs.FilStorage` is passed all the way down into this low level func that waits for a deal to become final.

ffs/filcold/filcold.go
```diff
@@ -284,35 +287,38 @@ Loop:
...
+				ProposalCid:     di.ProposalCid,
+				Status:          di.Status,
+				PieceCid:        di.PieceCID,
+				Duration:        int64(di.Duration),
+				Miner:           di.Miner,
+				ActivationEpoch: di.ActivationEpoch,
+				StartEpoch:      di.StartEpoch,
+				EpochPrice:      di.PricePerEpoch,
+			}
+			dealUpdates <- proposal
```
asutula, line 306: Every time we get an update from the deals module, we create a `FilStorage` with the current state and send it on the chan.

ffs/filcold/filcold.go
```diff
@@ -284,35 +287,38 @@ Loop:
...
 				fc.l.Log(ctx, "DealID %d with miner %s failed and won't be active on-chain: %s", di.DealID, di.Miner, di.Message)
 
 				return ffs.FilStorage{}, ffs.DealError{ProposalCid: di.ProposalCid, Miner: di.Miner, Message: di.Message}
 			default:
 				if di.DealID != 0 {
-					fc.l.Log(ctx, "Deal %d with miner %s changed state to %s", di.DealID, di.Miner, storagemarket.DealStates[di.StateID])
+					fc.l.Log(ctx, "Deal %d with miner %s changed state to %s", di.DealID, di.Miner, storagemarket.DealStates[di.Status])
 				} else {
-					fc.l.Log(ctx, "Deal with miner %s changed state to %s", di.Miner, storagemarket.DealStates[di.StateID])
+					fc.l.Log(ctx, "Deal with miner %s changed state to %s", di.Miner, storagemarket.DealStates[di.Status])
```
asutula, line 321: All other handling remains the same.

asutula:
The other way to do this potentially, is similar to the cid logger: Instead of passing this channel down into the stack and subscribing to it at some high level, I could create a package and struct that manages this deal progression state and provides methods to update and query it. This struct would be injected in as a dependency wherever it can be used.

deals/module/deals.go
jsign, line 596: `s/state/status`

jsign: Might be trickier whenever we support other Filecoin clients, but maybe we can always keep the mapping.

deals/types.go
```diff
@@ -23,8 +24,7 @@ type StoreResult struct {
 // StorageDealInfo contains information about a proposed storage deal.
 type StorageDealInfo struct {
 	ProposalCid cid.Cid
-	StateID     uint64
-	StateName   string
+	Status      storagemarket.StorageDealStatus
```
jsign, line 27: This looks like a breaking change.
If this struct was saved in some place in a Datastore, looks like might be a problem. 
Is that the case?

ffs/filcold/filcold.go
jsign: Might worth adding in the comments that this method will close `dealUpdates` channel that is received as a param, so the caller shouldn't since that would panic.

ffs/filcold/filcold.go
```diff
@@ -122,7 +122,7 @@ func (fc *FilCold) IsFilDealActive(ctx context.Context, proposalCid cid.Cid) (bo
 // that got renewed with Renewed=true. New deals from renewals are added to the returned FilInfo.
 // Note: Most probably all this code should change in the future, when Filecoin supports telling the miner which deal is about to
 // expire that we're interested in extending the deal duration. Now we should make a new deal from scratch (send data, etc).
-func (fc *FilCold) EnsureRenewals(ctx context.Context, c cid.Cid, inf ffs.FilInfo, cfg ffs.FilConfig, dealFinalityTimeout time.Duration) (ffs.FilInfo, []ffs.DealError, error) {
+func (fc *FilCold) EnsureRenewals(ctx context.Context, c cid.Cid, inf ffs.FilInfo, cfg ffs.FilConfig, dealFinalityTimeout time.Duration, dealUpdates chan<- ffs.FilStorage) (ffs.FilInfo, []ffs.DealError, error) {
```
jsign, line 125: Same comment as done below about "this method closes the chan".

ffs/filcold/filcold.go
jsign: Maybe is better to change this last line to:
```
select {
   case dealUpdates <- proposal:
   default:
        log.Warnf("slow receiver for deal updates for %s", c)
}
```
If the other side is slow or abandoned the channel, it's better to skip the notification and move on than be blocked here.

ffs/scheduler/scheduler.go
```diff
@@ -422,7 +428,29 @@ func (s *Scheduler) executeQueuedStorage(j ffs.StorageJob) {
 
 	// Execute
 	s.l.Log(ctx, "Executing job %s...", j.ID)
-	info, dealErrors, err := s.executeStorage(ctx, a, j)
+	dealUpdates := make(chan ffs.FilStorage)
```
jsign, line 431: Maybe make it buffered to leave some space because on start a lot of things might get pushed fast? e.g: `make(chan ffs.FilStorage, 1000)`?

ffs/scheduler/scheduler.go
jsign: Cool, looks like might be useful to also provide structured events.

ffs/types.go
jsign: Cool. This addition might leave previous stored things in CidInfo with an "empty/default" `Status` value.
I think that shouldn't be a problem despite some _weird_ output in some APIs, right?

jsign (commented):
Nice! Looks good, I left some comments for thought.

deals/module/deals.go
asutula: Agree it could be tricky, but we're already in the same situation because we're simply using the raw enum value and string representations, all provided by `storeagemarket`... We don't have an enum of our own. We probably should.

deals/types.go
asutula: Looks like it is stored in the deal store for the storage record stuff I did a while back. Maybe for now I'll stick with the previous `StateID` and `StateName` so we don't break those records. Long term, we should create our own state enum that we map to/from the `storagemarket` one (or any other lotus client of the future)

ffs/filcold/filcold.go
asutula: Cool, I've wondered how this "slow receiver" pattern works... How does it work? What does it mean to be slow and how is that detected?

jsign: When you do `dealUpdates <- proposal`, it will block until the receiver side reads the channel.
If the receiver is too slow reading the channel, this is might be a problem since you will be blocked at the sender side.

That `select` just avoids blocking and skips the push if the receiver isn't listening... so to never block.

jsign: Maybe in this case we can assume blocking is OK since the receiver should be fast enough or never abandon reading the channel. It's usually a judgment call.

ffs/types.go
asutula: Yea, I was willing to just deal with those default values, but now that I'm switching away from `Status storagemarket.StorageDealStatus` to the `StateID int` and `StateName string` values, I don't like the default of `0` there which is misleading for sure. I think I'll look more closely at a custom object for these updates. What's cool is that some of the values that start off as empty actually do get values as the deal is unfolding. I'll make sure this custom object represents those.

ffs/filcold/filcold.go
```diff
@@ -284,12 +291,17 @@ Loop:
...
-			return ffs.FilStorage{}, ffs.DealError{ProposalCid: proposal, Message: msg}
+			return ffs.FilStorage{}, ffs.DealError{ProposalCid: proposal, Miner: last.Miner, Message: msg}
 		case di, ok := <-chDi:
 			if !ok {
 				break Loop
 			}
 			last = di
+			select {
+			case dealUpdates <- di:
+			default:
```
asutula, line 302: Hey @jsign maybe I don't know Go switch syntax well, but do I need to put some statement inside the `case dealUpdates <- di:` case? Otherwise, isn't that case getting combined with the `default` case?

jsign: No, isn't merged. Take a look [here](https://play.golang.org/p/R-JdxLmg-D-). You can commend and uncomment the goroutine to see both situations.

jsign: (Comment/Uncomment the go routine)

ffs/scheduler/scheduler.go
```diff
@@ -422,7 +422,9 @@ func (s *Scheduler) executeQueuedStorage(j ffs.StorageJob) {
 
 	// Execute
 	s.l.Log(ctx, "Executing job %s...", j.ID)
-	info, dealErrors, err := s.executeStorage(ctx, a, j)
+	dealUpdates := s.sjs.MonitorJob(j)
+	info, dealErrors, err := s.executeStorage(ctx, a, j, dealUpdates)
```
asutula, line 426: get the channel from sjstore and pass it down into the job execution/cold storage execution layer.

ffs/types.go
```diff
@@ -109,6 +110,7 @@ type StorageJob struct {
 	Cid        cid.Cid
 	Status     JobStatus
 	ErrCause   string
+	DealInfo   []deals.StorageDealInfo
```
asutula, line 113: We now store storage deal state in `StorageJob` in the `sjstore`. We were previously already storing storage job error data there, this extends that to also store the rest of the deal information.

proto/powergate/v1/powergate.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
```
asutula, line 1: New proto file for the main api. This will expand over time as I move definitions out of the old "rpc" module definitions.

wallet/rpc/rpc.proto
```diff
@@ -40,6 +32,5 @@ message SendFilResponse {
 service RPCService {
     rpc NewAddress(NewAddressRequest) returns (NewAddressResponse) {}
     rpc List(ListRequest) returns (ListResponse) {}
-    rpc Balance(BalanceRequest) returns (BalanceResponse) {}
```
asutula, line 43: Moved this into `powergate.proto` just to see how it feels. Feels good! You'll see this change reflected in the Go client code.

asutula:
Hey @jsign when you get in here, keep an eye out for what might be causing a Lotus RPC client timeout. I'm seeing this when running the integration tests on this branch, but master works fine:
```
2020-10-14T10:34:33.113-0500    ERROR   rpc     go-jsonrpc@v0.1.2-0.20201008195726-68c6a2704e49/websocket.go:654        Connection timeout      {"remote": "127.0.0.1:33162"}
```
It hangs there until the test itself times out. Not clear to me what the problem could be.

jsign:
> Hey @jsign when you get in here, keep an eye out for what might be causing a Lotus RPC client timeout. I'm seeing this when running the integration tests on this branch, but master works fine:
> 
> ```
> 2020-10-14T10:34:33.113-0500    ERROR   rpc     go-jsonrpc@v0.1.2-0.20201008195726-68c6a2704e49/websocket.go:654        Connection timeout      {"remote": "127.0.0.1:33162"}
> ```
> 
> It hangs there until the test itself times out. Not clear to me what the problem could be.

[This](https://github.com/textileio/powergate/pull/671) other PR that has other changes also passes correctly (it had to rerun all the suite since has a new devnet), so this sounds like a code change in this PR.
Do you know what test fails? Does that test runs in your machine?

asutula:
Ok I think I fixed it @jsign. Likely some double mutex locking problem. Have to run, but will try to understand exactly why later.

asutula:
Pushed the fix, let's see what happens.

ffs/types.go
jsign: Cool. Important to tell people that this is Job related deals, not complete photo of things which is CidInfo, right?

e.g: 
- Cid1 is first executed by Job1 making 1 deal.
- Cid1 pushes a new config with rep factor 2. Job2 executes that config and creates 1 extra deal.

Result:
- Job1.DealInfo has 1 dea..
- Job2.DealInfo has 1 deals, since this job only created 1 deal.
- CidInfo should have the 2 deals information.

Same page?

jsign (approved):
LGTM! I'd definitely like to review this PR with more time.

To be sure about something really important: this PR doesn't contain struct changes that might break datastore persistence. While reviewing I think I didn't see any (after the one I pointed in the first review that was reverted), but you might have things fresher about changes here.

api/client/client.go
```diff
@@ -89,15 +103,18 @@ func NewClient(host string, optsOverrides ...grpc.DialOption) (*Client, error) {
...
 		Miners:          &Miners{client: minerRpc.NewRPCServiceClient(conn)},
 		Faults:          &Faults{client: faultsRpc.NewRPCServiceClient(conn)},
-		Wallet:          &Wallet{client: walletRpc.NewRPCServiceClient(conn)},
+		Wallet:          &Wallet{walletClient: walletRpc.NewRPCServiceClient(conn), powergateClient: powClient},
 		Reputation:      &Reputation{client: reputationRpc.NewRPCServiceClient(conn)},
 		FFS:             &FFS{client: ffsRpc.NewRPCServiceClient(conn)},
 		Health:          &Health{client: healthRpc.NewRPCServiceClient(conn)},
 		Net:             &Net{client: netRpc.NewRPCServiceClient(conn)},
+		Jobs:            &Jobs{client: powClient},
+		Admin:           &Admin{client: adminProto.NewPowergateAdminServiceClient(conn)},
```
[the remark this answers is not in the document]
asutula: Yea, just related to future changes that are in my head... I think there will just be `proto` and `adminProto` packages for the gRPC services. I'm just using package alias names that are more similar to the names of those new packages.

ffs/api/istore.go
```diff
@@ -84,48 +84,71 @@ func (s *instanceStore) removeStorageConfig(c cid.Cid) error {
...
-	}
-	var conf ffs.StorageConfig
-	if err := json.Unmarshal(buf, &conf); err != nil {
-		return ffs.StorageConfig{}, fmt.Errorf("unmarshaling cid config from datastore: %s", err)
-	}
-	return conf, nil
-}
-
-func (s *instanceStore) getCids() ([]cid.Cid, error) {
+func (s *instanceStore) getStorageConfigs(cids ...cid.Cid) (map[cid.Cid]ffs.StorageConfig, error) {
```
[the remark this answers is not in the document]
asutula: correct

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -51,18 +59,59 @@ type watcher struct {
...
+				s.jobStatusCache[j.APIID] = map[cid.Cid]map[cid.Cid]deals.StorageDealInfo{}
+			}
+			_, ok = s.jobStatusCache[j.APIID][j.Cid]
+			if !ok {
+				s.jobStatusCache[j.APIID][j.Cid] = map[cid.Cid]deals.StorageDealInfo{}
+			}
+			s.jobStatusCache[j.APIID][j.Cid][update.ProposalCid] = update
+			job, err := s.get(j.ID)
+			if err != nil {
+				log.Errorf("getting job: %v", err)
```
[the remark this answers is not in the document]
asutula: I think you're assuming I had a reason for this :)

Just an oversight, will  add `continue`, that is the correct thing.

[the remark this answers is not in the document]
asutula: went with this just to be safe:
```
			job, err := s.get(j.ID)
			if err == ErrNotFound {
				log.Errorf("job not found: %v", err)
				s.lock.Unlock()
				return
			}
			if err != nil {
				log.Errorf("getting job: %v", err)
				s.lock.Unlock()
				continue
			}
```

ffs/types.go
asutula: Yea same page. I want to be really clear about the difference between jobs and storage state. In an upcoming PR, I'd like to rename `CidInfo` to `StorageInfo` for that reason. So then we have `StorageJob`s and `StorageInfo`

ffs/scheduler/internal/sjstore/sjstore.go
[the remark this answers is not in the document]
jsign: Uhm, yeah tricky to `return`. The usual problem with returning a channel that can indicate that something bad happens... so the caller can't distinguish between "all is done and good, just finished" from "this was closed earlier since an error happened".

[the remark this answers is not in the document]
jsign: How bad that situation of misinterpreting a "success" call from an "errored" call be for clients that are using this API?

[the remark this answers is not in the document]
asutula: Ah yea. In this case, I'd say it isn't very important to convey that error to listeners. Removed the `return` and just keeping `continue`. 

The possible error is very unlikely, or impossible, and if it does happen, the worse case is that listeners don't get called back with updates. The error is logged, so we'll have something to look for if a user reports not receiving job updates.

PR #662, opened 2020-10-08
minerselector/sr2: filter by maxprice and always query ask && FFS MinPieceSize condition

### LINKED ISSUE #664: minerselector/sr2: consider configured MaxPrice
Make SR2 miner selector consider a `MaxPrice` greater than 0 condition in _StorageConfig_.

### LINKED ISSUE #665: ffs: allow specifying minimum piece size for Cold Storage
Allow configuring FFS to only allow to enable Cold Storage for data which piece size is at least some specified minimum.

Recall that the piece size isn't the data size, since data is padded to the next power of 2. e.g: data of size 100MiB will have a piece size of 128MiB.

jsign:
- Closes #664 
- Closes #665
- Force always query-asking miners since now in SR2 miners are changing prices too frequently and using the ask index cache might have stalled prices.
- Update Louts client to v0.9.0


ffs/integrationtest/filters/filters_test.go
```diff
@@ -56,7 +56,7 @@ func TestFilecoinTrustedMiner(t *testing.T) {
 
 		r := rand.New(rand.NewSource(22))
 		cid, _ := it.AddRandomFile(t, r, ipfsAPI)
-		trustedMiner := "t01001"
+		trustedMiner := "f01001"
```
jsign, line 59: In this transition phase to mainnet, now miners start with `f` not `t`.

ffs/minerselector/sr2/sr2.go
```diff
@@ -41,8 +39,8 @@ type bucket struct {
 }
 
 // New returns a new SR2 miner selector.
-func New(url string, ai *askindex.Runner, cb lotus.ClientBuilder) (*MinerSelector, error) {
-	ms := &MinerSelector{url: url, ai: ai, cb: cb}
+func New(url string, cb lotus.ClientBuilder) (*MinerSelector, error) {
```
jsign, line 42: Ugly diff, but basically:
- Always query ask miners in the list, since they're changing prices fast and the cache might give stalled values. Also, the ask index might be disabled temporarily so can also have stalled values by other reason.
- Take into account `MaxPrice` if set. If a miner doesn't satisfy that constraint, is skipped and it looks for a next one.

asutula (approved):
Looks good!

PR #668, opened 2020-10-12
SR2 miner selector price cap & index refactor

jsign:
Put a hard cap on max price from remote list, and do some good index refactors.

### DIFF
--- a/ffs/minerselector/sr2/sr2.go
+++ b/ffs/minerselector/sr2/sr2.go
@@ -82,7 +82,7 @@ func (ms *MinerSelector) GetMiners(n int, f ffs.MinerSelectorFilter) ([]ffs.Mine
 			bucket.Amount = len(miners)
 		}
 		var regionSelected int
-		for i := 0; i < len(miners); i++ {
+		for i := 0; regionSelected < bucket.Amount && i < len(miners); i++ {
 			sask, err := getMinerQueryAsk(c, miners[i])
 			if err != nil {
 				log.Warnf("sr2 miner %s query-ask errored: %s", miners[i], err)
--- a/ffs/minerselector/sr2/sr2.go
+++ b/ffs/minerselector/sr2/sr2.go
@@ -89,7 +89,7 @@ func (ms *MinerSelector) GetMiners(n int, f ffs.MinerSelectorFilter) ([]ffs.Mine
 				continue
 			}
 			if sask > maxSR2Price {
-				log.Warnf("skipping miner %s since has price above maximum allowed for SR2")
+				log.Warnf("skipping miner %s since has price %d above maximum allowed for SR2", miners[i], sask)
 				continue
 			}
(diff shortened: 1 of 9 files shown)


ffs/minerselector/sr2/sr2.go
```diff
@@ -83,6 +88,10 @@ func (ms *MinerSelector) GetMiners(n int, f ffs.MinerSelectorFilter) ([]ffs.Mine
 				log.Warnf("sr2 miner %s query-ask errored: %s", miners[i], err)
 				continue
 			}
+			if sask > maxSR2Price {
+				log.Warnf("skipping miner %s since has price %d above maximum allowed for SR2", miners[i], sask)
+				continue
+			}
```
jsign, line 94: Main intention of this PR.

ffs/minerselector/sr2/sr2.go
```diff
@@ -150,9 +159,27 @@ func getMinerQueryAsk(c *apistruct.FullNodeStruct, addrStr string) (uint64, erro
...
+	}()
+
+	select {
+	case <-time.After(time.Second * 10):
+		return 0, fmt.Errorf("query asking timed out")
+	case r := <-chAsk:
+		if r.Error != "" {
+			return 0, fmt.Errorf("query ask had controlled error: %s", r.Error)
+		}
+		return r.Ask.Price.Uint64(), nil
```
jsign, line 183: An unfortunate change that I have to do to work around what I fixed [here](https://github.com/filecoin-project/go-fil-markets/pull/429). Since the API isn't respecting cancelations, I had to "abandon" the call manually.
Without this the selection takes a lot of time if some miners don't reply to the query-ask, since Lotus retries with exponential backoff and isn't respecting the cancellation.

ffs/minerselector/sr2/sr2_test.go
```diff
@@ -0,0 +1,55 @@
+package sr2
```
jsign, line 1: These are two tests that are skipped by default. They're useful as wired tests when connecting to an external Lotus node. It's useful to have them handy to tests things in testnet since it's impossible to test this selector in a devnet environment.

index/miner/module/meta.go
```diff
@@ -18,51 +18,53 @@ import (
 )
```
jsign, line 18: All changes in the indexes aren't that interesting. I'm simplifying some logic and having a bit more ordered logic since I did some exploration about some situations... then took advantage and improved things a bit.

index/miner/module/miner_test.go
```diff
@@ -57,6 +58,25 @@ func TestFullRefresh(t *testing.T) {
 	}
 }
 
+func TestIntegration(t *testing.T) {
```
jsign, line 61: Same "external Lotus" node test that is handy to have, but skipped by default.

asutula (approved):

PR #671, opened 2020-10-13
index/miners: make corrupted checkpoint a recoverable error & Lotus v0.10.1 update

jsign:
Makes a corrupted checkpoint on indexes a recoverable error.

This was motivated by a user that failed to load the Miner Index since got some invalid json stored in a Badger key, which looks impossible. Running on this branch got:
`2020-10-13T20:31:05.555Z	WARN	chainstore	chainstore/chainstore.go:115	corrupted checkpoint:  in deal activation: failed to set up called handler: called check error (h: 128743): client: failed to` (see lines changed in this PR to understand where is it).

That means that we expected `in deal activation: failed to set up called handler: called check error (h: 128743): client: failed to` to be a JSON string, which looks too weird. I can't make sense of this in any way, since we always save things result of `json.Marshal(...)`.

[Here](https://github.com/textileio/powergate/blob/d1840e23251c53b03457d8eb33cc72baf7b0253f/index/miner/module/onchain.go#L88) is the only call to `Save(...)` that can write values in that key. And `chainIndex` is defined as `var chainIndex miner.ChainIndex`, so no chance a `json.Marshal(...)` can get a string value as the above.

I thought that maybe some Datastore namespacing wrapping might be missing on `server.go` to make sense its reading the wrong place, but all seems good. The user this wasn't a super old Powergate in which some version upgrading wasn't compatible or similar.

@asutula, can you imagine how this can make sense?

This PR just discards saved checkpoints which are invalid, which results in just rebuilding the index and move forward. The user with this situation recovered his instance correctly with this.

---

Also updates Powergate to Lotus v0.10.1 which has a breaking change in wallet creation. I had also to update the `lotus-devnet` since the breaking API would make the CI fail (and `make localnet` too).


chainstore/chainstore.go
```diff
@@ -110,7 +114,8 @@ func (s *Store) load(tsk types.TipSetKey, v interface{}) error {
 		return fmt.Errorf("error getting key %s: %s", key, err)
 	}
 	if err := json.Unmarshal(buf, v); err != nil {
-		return err
+		log.Warnf("corrupted checkpoint: %s", buf)
```
jsign, line 117: So somehow here `buf` is a plain string (mentioned in PR description) instead of a `chainIndex` json.

asutula (approved):
Very strange. I have no idea what it could be. This code has been quite static for a long time.

PR #682, opened 2020-10-19
update go-ds-mongo and account for errors in queries

jsign:
Updates `go-ds-mongo` version which has bug fixes which caused panics in very unlucky situations of where the MongoDB database timeout.

Also, account for checking always errors in queries!


asutula (approved):
Good find!

PR #683, opened 2020-10-19
Completely remove SendFil

asutula:
Had previously removed it from some apis, but not all. Now it's gone gone.


ffs/api/api_wallet.go
```diff
@@ -81,14 +80,6 @@ func (i *API) VerifyMessage(ctx context.Context, addr string, message, signature
 	return ok, nil
 }
 
-// SendFil sends fil from a managed address to any another address, returns immediately but funds are sent asynchronously.
```
jsign, line 84: Do we need to delete it from the internal API? Sounds to me that only at the gRPC level is enough.

asutula: Yea, actually realized that and reversing now. Same in wallet module.

andrewxhill: doh... this just made me wonder. couldn't it be a flag? seems like those running their own Pow would still want these.

asutula: hmmmm, I'm not sure. thinking.

jsign (approved):

asutula:
closing in favor of https://github.com/textileio/powergate/pull/684

PR #685, opened 2020-10-20
pow ffs log [show log with interval]

ipfsauth:
pow ffs log only show the latest log.

### DIFF
--- a/cmd/pow/cmd/ffs_log.go
+++ b/cmd/pow/cmd/ffs_log.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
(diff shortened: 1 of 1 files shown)


asutula:
Can you explain a little more what your goal is here. Your description is vague. From looking at the code, it seems like your change will cause the watching of logs to stop after `-i` seconds? Is that right?

ipfsauth:
> Can you explain a little more what your goal is here. Your description is vague. From looking at the code, it seems like your change will cause the watching of logs to stop after `-i` seconds? Is that right?

YES, sometimes i just need to see the latest log, i don't need watch the log.

first i stage a lot of files.
then i push all Cids to Filecoin.
finally i check all Cids one by one use the log interface, BUT the log is watch it.

PR #686, opened 2020-10-20
Client/CLI simplifications and improvements

asutula:
This PR is targeting my `asutula/deal-insight` branch, goal being to accumulate big changes into that PR and do one big breaking release.

Main goal of this one is to simplify and standardize the client api and cli.

- Client methods always return gRPC types, no more converting back to the various go types we use on the back end. This keeps the client feeling simple, reduced dependencies on the consumer of the client, and provides encoding/decoding to/from json that should be more consistent across platforms/languages.
- The CLI is simplified, and in most cases simply just prints out the json version of the object received from the client. Removed spinners and other fancy stuff. This could all be considered a regression in some ways, but I think it's a better fit for the current state of the `pow` cli... We have some users that are having trouble scripting `pow` commands because of the spinners etc, and simply printing the data, all the data, fulfills the general need for maximum information possible in these crazy times of bugs and network instability. 

Lots of files changed, but those types of changes are the only things that were done.


jsign (approved):
LGTM!

PR #689, opened 2020-10-21
Repurpose CidInfo

asutula:
`CidInfo` has always represented the current storage state of a cid. A better name for it is `StorageInfo`. An that frees up the name `CidInfo` to be used in the API as an object that provides general information about the state of a cid in the system.


ffs/scheduler/internal/cistore/cistore.go
```diff
@@ -45,8 +45,8 @@ func (s *Store) Get(c cid.Cid) (ffs.CidInfo, error) {
 	return ci, nil
 }
 
-// Put saves a new storing state for a Cid.
-func (s *Store) Put(ci ffs.CidInfo) error {
+// Put saves a new storage state for a Cid.
+func (s *Store) Put(ci ffs.StorageInfo) error {
```
jsign, line 49: So we got lucky that the JSON serialization is naming agnostic.

jsign (approved):
LGTM. Getting a bit nervous about such a big PR and if we're missing some breaking change at the datastore layer; shouldn't be the case.

ffs/scheduler/internal/cistore/cistore.go
asutula: Exactly ;)

asutula:
I've really gone out of my way to make sure there are no breaking datastore changes in all this work. And all my remaining big changes are going to be strictly API related.... so I _thiiiink_ we're ok.

PR #694, opened 2020-10-28
Auto-throttling plus many improvements & fixes.

jsign:
This PR contains multiple improvements and updates.

Lotus load handling:
- Allow flag/env cap on the number of maximum parallel deal preparations that can be executed in parallel. Deal preparation (piece size calculation and CommP calculation) are the most resource-intensive tasks for Lotus. This allows to bump the total parallel job constraints in a safer way since the real intensive part of the Job is limited in a surgical way.
- Powergate will generate *warnings in the log if detects that Lotus is falling behind in syncing*. This will help operators to understand that something might be going wrong, or simply to be aware of this fact. e.g log: `WARN    lotus-client    lotus/metrics.go:93     Louts behind in syncing with height diff 39, todo: 6`. So, it also mentions how much of the difference is and which is the current progress (similar to `lotus sync wait`).
- Leveraging the above mechanism, Powergate it *will block any new deal-making to let the Lotus node recover*. This is an important feature since is a **dynamic throttling/limiting** of the load, and not relying on some hard number specified by flags/envs which is hard to tune correctly. If this situation is happening, the operator can see in the logs messages indication this is the case: e.g log: `WARN    ffs-filcold     filcold/filcold.go:296 backpressure from unsynced Lotus node`.

To be clear about the two last points since it might look similar. The former is a general monitoring control that will warn if the Lotus node is falling behind at any moment (maybe you're not even making deals). The latter is indicating that Powergate is trying to make deals, but is throttling since detected the Lotus node is syncing... whenever the Lotus node gets in sync, it will continue with its work. 

Enhancements:
- Stop using our custom `calculatePieceSize` logic. Originally, we needed to do this since there wasn't an API to resolve it. Now use the existing Lotus API which might give future-proof size calculation.
- Allow tuning how many retries are needed to consider that connecting to Lotus is an error. Use a big default. This default is nice if the underlying Lotus is restarted or has some problem, so we give Lotus time to recover and not break the flow of whatever Powergate was doing, which in many cases could be undesired.
- Workaround known problem of `ClientGetDealInfo` not reporting active deals. (until the real problem is fixed in Lotus).

Indices:
- Include in Ask index `MaxPieceSize` information.

Miner selection:
- Consider in `reptop` and `sr2` miner selection, the `MinPieceSize`, and `MaxPieceSize` of miners to avoid selecting miners which will reject the deal depending on deals data size.

Chore:
- Update the Lotus client to v1.1.2.
- Update Lotus docker image to v1.1.2.


api/client/asks.go
```diff
@@ -54,6 +54,7 @@ func askFromPbAsk(a *rpc.StorageAsk) ask.StorageAsk {
 	return ask.StorageAsk{
 		Price:        a.GetPrice(),
 		MinPieceSize: a.GetMinPieceSize(),
+		MaxPieceSize: a.GetMaxPieceSize(),
```
jsign, line 57: Include max piece size in the ask index.

api/server/server.go
```diff
@@ -64,8 +64,7 @@ import (
 )
 
 const (
-	datastoreFolderName    = "datastore"
-	lotusConnectionRetries = 10
```
jsign, line 68: Now a config attribute. Line 115.

api/server/server.go
```diff
@@ -152,13 +153,16 @@ func NewServer(conf Config) (*Server, error) {
 	}
 
 	var err error
-	clientBuilder, err := lotus.NewBuilder(conf.LotusAddress, conf.LotusAuthToken, lotusConnectionRetries)
+	clientBuilder, err := lotus.NewBuilder(conf.LotusAddress, conf.LotusAuthToken, conf.LotusConnectionRetries)
 	if err != nil {
 		return nil, fmt.Errorf("creating lotus client builder: %s", err)
 	}
-	lotus.MonitorLotusSync(clientBuilder)
+	lsm, err := lotus.NewSyncMonitor(clientBuilder)
```
jsign, line 160: What before was some background running daemon that generated Prometheus metrics for the grafana dashboard regarding Lotus sync height, now is a more formal component to provide Lotus syncing information for decision making.

api/server/server.go
```diff
@@ -250,7 +254,7 @@ func NewServer(conf Config) (*Server, error) {
 	if conf.Devnet {
 		conf.FFSMinimumPieceSize = 0
 	}
-	cs := filcold.New(ms, dm, ipfs, chain, l, conf.FFSMinimumPieceSize)
+	cs := filcold.New(ms, dm, ipfs, chain, l, lsm, conf.FFSMinimumPieceSize, conf.FFSMaxParallelDealPreparing)
```
jsign, line 257: This new component is used by Cold Storage, plus a new flag that indicates how many data-intensive tasks we throw at lotus in parallel. (More on this later).

deals/module/deals.go
```diff
@@ -31,7 +31,7 @@ import (
 const (
 	chanWriteTimeout = time.Second
 
-	defaultDealStartOffset = 72 * 60 * 60 / util.EpochDurationSeconds // 72hs
+	defaultDealStartOffset = 48 * 60 * 60 / util.EpochDurationSeconds // 48hs
```
jsign, line 34: Still, a question mark which is the right value for this. 
Eventually, we can move it to the Storage Config and let people have the change to play with it, but most prob they won't know the consequences yet.

deals/module/deals.go
```diff
@@ -173,11 +173,27 @@ func (m *Module) Store(ctx context.Context, waddr string, dataCid cid.Cid, piece
 	return res, nil
 }
 
+// CalculatePieceSize calculates the data and piece size of a Cid accesible
+// by the underlying Lotus node.
+func (m *Module) CalculatePieceSize(ctx context.Context, c cid.Cid) (api.DataSize, error) {
```
jsign, line 178: We're replacing our own logic of calculating piece sizes, to an available API from Lotus.
Originally in March we had to do it ourselves since this didn't exist.

Very soon there will be a new API for also calculating CommP, which will replace this one since will allow us to make piece size calculation and CommP independent of the replication factor.. but that's will be something when the next Lotus release appears.

deals/module/deals.go
```diff
@@ -298,23 +314,21 @@ func (m *Module) retrieve(ctx context.Context, lapi *apistruct.FullNodeStruct, l
...
+// GetDealStatus returns the current status of the deal.
+func (m *Module) GetDealStatus(ctx context.Context, pcid cid.Cid) (storagemarket.StorageDealStatus, error) {
+	lapi, cls, err := m.clientBuilder(ctx)
 	if err != nil {
-		return storagemarket.StorageDealUnknown, false, fmt.Errorf("creating lotus client: %s", err)
+		return storagemarket.StorageDealUnknown, fmt.Errorf("creating lotus client: %s", err)
 	}
 	defer cls()
-	di, err := lapi.ClientGetDealInfo(ctx, pcid)
+	di, err := robustClientGetDealInfo(ctx, lapi, pcid)
```
jsign, line 324: Considering that `ClientGetDealInfo` can lie about the real status of the deal, I created `robustClientGetDealInfo` with a workaround logic. All places that used `ClientGetDealInfo` now use `robustClientGetDealInfo`.

deals/module/deals.go
```diff
@@ -636,6 +650,33 @@ func notifyChanges(ctx context.Context, client *apistruct.FullNodeStruct, currSt
 	return nil
 }
 
+func robustClientGetDealInfo(ctx context.Context, lapi *apistruct.FullNodeStruct, propCid cid.Cid) (*api.DealInfo, error) {
+	di, err := lapi.ClientGetDealInfo(ctx, propCid)
+	if err != nil {
+		return nil, fmt.Errorf("client get deal info: %s", err)
+	}
+
+	// Workaround ClientGetDealInfo giving unreliable status.
```
jsign, line 659: Kind of self-explanatory. 

If `ClientGetDealInfo` starts working correctly again, the workaround code will never get executed... so no rush to remove the workaround code. Of course, we should eventually do it to avoid dead code; but the point is that there's no rush plus no performance impact.

ffs/filcold/filcold.go
```diff
@@ -77,50 +81,80 @@ func (fc *FilCold) Fetch(ctx context.Context, pyCid cid.Cid, piCid *cid.Cid, wad
 	return ffs.FetchInfo{RetrievedMiner: miner, FundsSpent: fundsSpent}, nil
 }
 
+func (fc *FilCold) calculatePieceSize(ctx context.Context, c cid.Cid) (api.DataSize, error) {
```
jsign, line 84: The two places that need to calcualte the piece size now:
- Use the Lotus API instead of our custom logic.
- Are rate limited by the deal-preparation-limit flag.
- Even if they pass the rate-limiting, they're blocked if the Lotus sync monitor is claiming the lotus node is having a hard time syncing, so it can recovers.

ffs/filcold/filcold.go
```diff
@@ -77,50 +81,80 @@ func (fc *FilCold) Fetch(ctx context.Context, pyCid cid.Cid, piCid *cid.Cid, wad
...
+	size, err := fc.calculatePieceSize(ctx, c)
+	if err != nil {
+		return nil, nil, 0, fmt.Errorf("getting cid cummulative size: %s", err)
+	}
+	fc.l.Log(ctx, "Calculated piece size is %d MiB.", size.PieceSize/1024/1024)
+	pieceSize := uint64(size.PieceSize)
+
+	if pieceSize < fc.minPieceSize {
+		return nil, nil, 0, fmt.Errorf("Piece size is below allowed minimum %d MiB", fc.minPieceSize/1024/1024)
+	}
```
jsign, line 128: Calculating piece size is moved before query-asking miners, so if the deal size is lower than limit we avoid doing work of doing miner selection.

Apart from performance reasons, we now use the piece size to select miners, so this change was needed anyway.

ffs/filcold/filcold.go
```diff
@@ -244,6 +279,28 @@ func (fc *FilCold) renewDeal(ctx context.Context, c cid.Cid, size uint64, p ffs.
...
+		if fc.lsm.SyncHeightDiff() < unsyncedThreshold {
+			break
+		}
+		log.Warnf("lotus backpressure from unsynced node")
+		select {
+		case <-ctx.Done():
+			return nil, nil, fmt.Errorf("canceled by context")
+		case <-time.After(time.Minute):
+		}
+	}
```
jsign, line 302: Similar limiting done in `calculatePieceSize`. I decided not to extract the limiting logic since this particular piece of code will go away in Lotus v1.1.3 when we have the new mentioned API.

When that lands, we don't have to rate limit things here... since we'll use what was calculated in `calculatePieceSize` when making deals, so Lotus doesn't have to recalculate heavy things, thus no need to rate-limit here anymore.

ffs/filcold/filcold.go
```diff
@@ -326,53 +383,6 @@ Loop:
...
-	// Consider padding.
-	paddedSize := padreader.PaddedSize(totalSize).Padded()
-
-	return uint64(paddedSize), nil
-}
-
-func (fc *FilCold) getLinks(ctx context.Context, c cid.Cid) ([]*format.Link, error) {
-	return fc.ipfs.Object().Links(ctx, path.IpfsPath(c))
-}
-
```
jsign, line 375: I was proud of doing this code since there wasn't a good way to calculate the right piece size... but has to go now.

ffs/minerselector/reptop/reptop.go
```diff
@@ -49,6 +49,9 @@ func (rt *RepTop) GetMiners(n int, f ffs.MinerSelectorFilter) ([]ffs.MinerPropos
 		if f.MaxPrice > 0 && sa.Price > f.MaxPrice {
 			continue
 		}
+		if f.PieceSize < sa.MinPieceSize && f.PieceSize > sa.MaxPieceSize {
+			continue
+		}
```
jsign, line 54: Consider min and max piece size of miners.

ffs/minerselector/sr2/sr2.go
```diff
@@ -88,17 +88,20 @@ func (ms *MinerSelector) GetMiners(n int, f ffs.MinerSelectorFilter) ([]ffs.Mine
...
 			}
-			if f.MaxPrice > 0 && sask > f.MaxPrice {
-				log.Warnf("skipping miner %s with price %d higher than max-price %d", miners[i], sask, f.MaxPrice)
+			if f.MaxPrice > 0 && sask.Price.Uint64() > f.MaxPrice {
+				log.Warnf("skipping miner %s with price %d higher than max-price %d", miners[i], sask.Price, f.MaxPrice)
 				continue
 			}
+			if f.PieceSize < uint64(sask.MinPieceSize) || f.PieceSize > uint64(sask.MaxPieceSize) {
+				log.Warnf("skipping miner %s since needed piece size %d doesn't fit bounds (%d, %d)", miners[i], f.PieceSize, sask.MinPieceSize, sask.MaxPieceSize)
+			}
```
jsign, line 101: Same with SR2 miner selector.

index/ask/rpc/rpc.go
```diff
@@ -28,6 +28,7 @@ func (s *RPC) Get(ctx context.Context, req *GetRequest) (*GetResponse, error) {
 		storage[key] = &StorageAsk{
 			Price:        ask.Price,
```
jsign, line 29: @asutula , I just did the usual RPC/CLI changes to be coherent.. but this will go away anyway in your changes.

lotus/metrics.go
```diff
@@ -13,4 +19,123 @@ var (
 		Description: "Current height of Lotus node",
 		Aggregation: view.LastValue(),
 	}
+
+	metricHeightInterval = time.Second * 120
 )
+
+// SyncMonitor provides information about the Lotus
```
jsign, line 26: TL;DR: SyncMonitor analyzes the sync status every 2 minutes, and provides cached value to whoever wants to know about it.

lotus/metrics.go
```diff
@@ -13,4 +19,123 @@ var (
...
+	return lsm, nil
+}
+
+// SyncHeightDiff returns the height difference between the tip of the chain
+// and the current synced height.
+func (lsm *SyncMonitor) SyncHeightDiff() int64 {
+	lsm.lock.Lock()
+	defer lsm.lock.Unlock()
+	return lsm.heightDiff
+}
```
jsign, line 62: Provide cached value... we don't want people asking lotus about its sync status to add load to it.
Since the syncing process is rather slow, a value 2 minutes old is usually good enough to make decisions.

lotus/metrics.go
```diff
@@ -13,4 +19,123 @@ var (
...
+				maxHeightDiff = heightDiff
+				remaining = int64(as.Target.Height() - as.Height)
+			}
+		}
+	}
+
+	lsm.lock.Lock()
+	lsm.heightDiff = maxHeightDiff
+	lsm.remaining = remaining
+	lsm.lock.Unlock()
```
jsign, line 138: The Lotus syncing process has many workers that are trying to sync different possible branches.
I decided to go pessimistic and consider the worst-case scenario. I think this is the best decision.

deals/module/deals.go
asutula: Very nice.

ffs/filcold/filcold.go
```diff
@@ -77,50 +81,80 @@ func (fc *FilCold) Fetch(ctx context.Context, pyCid cid.Cid, piCid *cid.Cid, wad
 	return ffs.FetchInfo{RetrievedMiner: miner, FundsSpent: fundsSpent}, nil
 }
 
+func (fc *FilCold) calculatePieceSize(ctx context.Context, c cid.Cid) (api.DataSize, error) {
+	// Limiting deal size calculation since is very resource intensive.
+	// In the next Lotus version (v1.1.3) a new API will be used which also calculates
+	// CommP, so we can help Lotus avoid recalculating size and CommP when deals are made.
+	select {
+	case fc.semaphDealPrep <- struct{}{}:
```
asutula, line 89: I am sure I'm just missing some understanding about how this works, but I would have expected to see some goroutine spawned off after the `fc.semaphDealPrep` is used here to block execution until a slot opens up. As is, it looks to me like calls to `calculatePieceSize` are run serially. Like I said, I'm sure I just don't understand, but I'd like to understand this because it seems awesome!

ffs/filcold/filcold.go
asutula: 😿

ffs/minerselector/reptop/reptop.go
asutula: Think you mean `||`

index/ask/rpc/rpc.go
asutula: 👍

asutula (approved):
This is very exciting, no doubt that things are going to work much much better than before. 

I found one actual bug you need to fix, then just left a couple comments/questions for my own understanding of things. 

Very nice work.

ffs/filcold/filcold.go
jsign: When the scheduler decides to run a Job, it [spawns](https://github.com/textileio/powergate/blob/4b54223d97c4be65fb089c368b0033760975f064/ffs/scheduler/scheduler.go#L379) a new goroutine and let that goroutine execute all the logic.

This means that at this point, when executing cold storage logic, it should only wait for the "OK" of this semaphore to keep doing the work it was doing.

If you see what's happening in a bigger scope, you might see multiple goroutines (Jobs) waiting in this semaphore to be allowed to move on executing heavy stuff.

The good part of this design is that you shouldn't think about the job execution logic methods as single goroutine doing things for multiple jobs. You can always assume you're doing things for a Job already in its own goroutine.

That's why you see that `calculatePieceSize` is serial. Is serial in a single Job execution, but there're multiple goroutines one per executing Job. Whatever buffer this semaphore allows, are parallel executing goroutine Jobs that will be running in parallel calculating their size.

Makes sense?

ffs/minerselector/reptop/reptop.go
jsign: Oops, correct.

ffs/filcold/filcold.go
asutula: Yea that totally makes sense. I forgot that gouroutines for jobs are created quite early and high up the stack. Yea this is a really nice design because it makes this code look very simple and easy to understand.

jsign: Exactly!

PR #704, opened 2020-11-11
API nits + refactor

asutula:
While integrating the Powergate v1.0.0 update into Hub, I felt a few things I wanted to change here before the v1.x Powergate become widely deployed/used.

These updates mostly are related to trying to get the proto files and generated code perfect so they can become stable and we can start doing proper backward compatible proto updates moving forward (minimize breaking changes).

There are just a few small but important changes here that led to a hundred files being updated simply because of renaming of a couple things and moving the generated proto code (most file updates are generated cli docs, and every gRPC service file and client file was updated because of new import paths for generated proto code). Here is what changed:

- In the gRPC API, renamed `Job` to `StorageJob` just since we're being clear about storage jobs vs retrieval jobs elsewhere in the codebase.
- Moved away from "storage profile" wording in the API and client/cli to "user". I originally wanted to avoid using "user" because it implies some use case, but in reality, it is the main use case and it just makes everything easier to talk about and understand.
- Updated two client methods to avoid stuttering: `Client.Users.Create()` and `Client.Users.List()` are the new names.
- Adopted some improvements for proto file organization and naming. The two proto files now define the "user" API (the main api that most users use) and the "admin" api. 
- Moved generated proto code to the `api/gen` directory. This is nice because the proto types are part of the public api, and now all client and server imports come from `github.com/textileio/powergate/api` and its sub packages.
- While adopting more proto best practices, I checked in on Buf updates and saw they now support proto code generation. This allowed me to update our Go generation to use dependencies managed by Bingo (`buf`, `protoc-gen-go` and `protoc-gen-go-grpc`)... no more `buildtools` folder and script! Much cleaner.
- Using `buf`'s code generation also made it easy to update to the new/current versions of `protoc-gen-go` and `protoc-gen-go-grpc` which I've been meaning to do for a while. Our previous code generation shell script would have needed some updating that wasn't quite clear to me, but using `buf`'s declarative code generation configuration made this easy.


.bingo/Variables.mk
```diff
@@ -52,3 +52,15 @@ $(PROTOC_GEN_BUF_CHECK_LINT): .bingo/protoc-gen-buf-check-lint.mod
...
+	@# Install binary/ries using Go 1.14+ build command. This is using bwplotka/bingo-controlled, separate go module with pinned dependencies.
+	@echo "(re)installing $(GOBIN)/protoc-gen-go-grpc-v1.0.1"
+	@cd .bingo && $(GO) build -modfile=protoc-gen-go-grpc.mod -o=$(GOBIN)/protoc-gen-go-grpc-v1.0.1 "google.golang.org/grpc/cmd/protoc-gen-go-grpc"
+
+PROTOC_GEN_GO := $(GOBIN)/protoc-gen-go-v1.25.0
+$(PROTOC_GEN_GO): .bingo/protoc-gen-go.mod
+	@# Install binary/ries using Go 1.14+ build command. This is using bwplotka/bingo-controlled, separate go module with pinned dependencies.
+	@echo "(re)installing $(GOBIN)/protoc-gen-go-v1.25.0"
+	@cd .bingo && $(GO) build -modfile=protoc-gen-go.mod -o=$(GOBIN)/protoc-gen-go-v1.25.0 "google.golang.org/protobuf/cmd/protoc-gen-go"
+
```
asutula, line 66: `protoc-gen-go` and `protoc-gen-go-grpc` now managed by and accessed through `bingo`

.github/workflows/release.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           replace-with: ""
       - name: Generate JS gRPC bindings
         run: |
-          ./scripts/gen-js-protos.sh ${{steps.makeversion.outputs.replaced}} . ./js-grpc
+          ./scripts/gen-js-protos.sh ${{steps.makeversion.outputs.replaced}} ./proto ./js-grpc
```
asutula, line 97: `./proto/` is now the "root" of our proto files... `./proto/` itself is not part of the proto path, it's just the container for all protos and the actual proto paths are relative to this root.

.bingo/Variables.mk
jsign: Nice

Makefile
```diff
@@ -82,12 +82,8 @@ clean-protos:
...
 .PHONY: clean-protos
 
-install-protoc:
-	cd buildtools && ./protocInstall.sh
-	
-PROTOCGENGO=$(shell pwd)/buildtools/protoc-gen-go
-protos: install-protoc clean-protos
-	PATH=$(PROTOCGENGO):$(PATH) ./scripts/protoc_gen_plugin.bash --proto_path=. --plugin_name=go --plugin_out=. --plugin_opt=plugins=grpc,paths=source_relative
+protos: $(BUF) $(PROTOC_GEN_GO) $(PROTOC_GEN_GO_GRPC) clean-protos
+	$(BUF) generate --template '{"version":"v1beta1","plugins":[{"name":"go","out":"api/gen","opt":"paths=source_relative","path":$(PROTOC_GEN_GO)},{"name":"go-grpc","out":"api/gen","opt":"paths=source_relative","path":$(PROTOC_GEN_GO_GRPC)}]}'
```
asutula, line 86: Using `buf` to generate the Go protos. `buf` actually supports configuring this via a config file, which is really nice, but I needed to pass the location of the `buf`, `protoc-gen-go`, and `protoc-gen-go-grpc` binaries managed by Bingo in dynamically, so I instead used `buf`'s flag for passing the config in as json on the command line.

jsign, line 86: Nice, the `protocInstall.sh` was a stretch, this looks more maintainable.

api/client/admin/admin.go
```diff
@@ -1,21 +1,21 @@
...
 import (
-	proto "github.com/textileio/powergate/proto/admin/v1"
+	adminPb "github.com/textileio/powergate/api/gen/powergate/admin/v1"
 )
 
 // Admin provides access to Powergate admin APIs.
 type Admin struct {
 	StorageJobs *StorageJobs
-	Profiles    *Profiles
+	Users       *Users
```
asutula, line 10: New "user" wording in the client. Feels much easier and makes sense.

api/server/admin/service.go
```diff
@@ -1,13 +1,15 @@
...
 import (
+	adminPb "github.com/textileio/powergate/api/gen/powergate/admin/v1"
 	"github.com/textileio/powergate/ffs/manager"
 	"github.com/textileio/powergate/ffs/scheduler"
 	"github.com/textileio/powergate/wallet"
 )
 
 // Service implements the Admin API.
 type Service struct {
+	adminPb.UnimplementedAdminServiceServer
```
asutula, line 12: Required by the new version of gRPC to help with evolving the API moving forward. That type just provides default implementations of the service that returning `unimplemented` errors.

api/client/admin/admin.go
jsign: I agree that user feels weird, but I definitely prefer it over storage profile.

buf.yaml
```diff
@@ -1,14 +1,10 @@
+version: v1beta1
 build:
   roots:
-    - .
-  excludes:
-    - buildtools
+    - proto
```
asutula, line 4: That new root

buf.yaml
```diff
@@ -1,14 +1,10 @@
...
 lint:
   use:
     - DEFAULT
-  except:
-    - PACKAGE_VERSION_SUFFIX
```
asutula, line 10: We are doing the version suffix now

proto/powergate/admin/v1/admin.proto
```diff
@@ -1,9 +1,9 @@
 syntax = "proto3";
-package proto.admin.v1;
+package powergate.admin.v1;
```
asutula, line 2: Using `proto` as the root allows the package to be more proper like this.

api/client/admin/storagejobs.go
```diff
@@ -3,26 +3,26 @@ package admin
 import (
 	"context"
 
-	proto "github.com/textileio/powergate/proto/admin/v1"
+	adminPb "github.com/textileio/powergate/api/gen/powergate/admin/v1"
```
jsign, line 6: Nice, the `/gen` subpath makes clear is gRPC releated.

jsign (approved):
Nice, I like the changes!

PR #705, opened 2020-11-12
Improve deal making efficiency & Lotus v1.1.3 update & fixes

jsign:
This PR:
- Update Lotus client  & docker image to v1.1.3
- Leverage a [PR](https://github.com/filecoin-project/lotus/pull/4584) included in 1.1.3 I made in Lotus some time ago that allows removing the `replace` directive we had in `go.mod`; GO clients will be happier.
- Leverage new APIs to only calculate once deal size and CommP. This makes Job execution load on the Lotus node be independent of the replication factor. That is if the replication factor was 5, this PR decreases the total cumulative load in the Job execution by 5x in Lotus.
- The _TotalActiveDeals_ is removed from the miner's index. Unfortunately, this was derived from an API that lists all the sectors of miners to do some extra calculation, but for big miners, this can have a huge output in the API call which isn't paged thus trigger OOMs. Total active deals were nice metrics, but honestly is very easy to fake anyway so not really convincing for decisions anyway. If we come up to think this was important later, we should research/escalate a more efficient way to derive this. Unblocks https://github.com/textileio/textile/issues/397.
- Make the index raw json output bind-address be configurable, and bind to dynamic free ports to avoid collisions with other running things in the same GH runner.


deals/module/deals.go
```diff
@@ -146,9 +146,11 @@ func (m *Module) Store(ctx context.Context, waddr string, dataCid cid.Cid, piece
 			Data: &storagemarket.DataRef{
 				TransferType: storagemarket.TTGraphsync,
 				Root:         dataCid,
+				PieceCid:     &pieceCid,
+				PieceSize:    pieceSize.Unpadded(),
```
jsign, line 150: We delegate to the client to provide `pieceSize` and `pieceCid`. This allows Lotus to avoid calculating these values, and just trusting they're correct. This is important for clients such as FFS, which create multiple deals with the same data.

deals/module/deals.go
```diff
@@ -173,18 +175,17 @@ func (m *Module) Store(ctx context.Context, waddr string, dataCid cid.Cid, piece
...
+func (m *Module) CalculateDealPiece(ctx context.Context, c cid.Cid) (api.DataCIDSize, error) {
 	lapi, cls, err := m.clientBuilder(ctx)
 	if err != nil {
-		return api.DataSize{}, fmt.Errorf("creating lotus client: %s", err)
+		return api.DataCIDSize{}, fmt.Errorf("creating lotus client: %s", err)
 	}
 	defer cls()
 
-	dsz, err := lapi.ClientDealSize(ctx, c)
+	dsz, err := lapi.ClientDealPieceCID(ctx, c)
```
jsign, line 186: We use a new API that calculates both size and PieceCID.

ffs/filcold/filcold.go
```diff
@@ -81,14 +81,14 @@ func (fc *FilCold) Fetch(ctx context.Context, pyCid cid.Cid, piCid *cid.Cid, wad
 	return ffs.FetchInfo{RetrievedMiner: miner, FundsSpent: fundsSpent}, nil
 }
 
-func (fc *FilCold) calculatePieceSize(ctx context.Context, c cid.Cid) (api.DataSize, error) {
+func (fc *FilCold) calculateDealPiece(ctx context.Context, c cid.Cid) (abi.PaddedPieceSize, cid.Cid, error) {
```
jsign, line 84: I preferred to avoid using `api` namespace here since is a Lotus package.
Also, I decided using `abi.PaddedPieceSize` since using `uint64` is ambiguous regarding if the size includes padding or not. It's a technical detail that matters in using the right size type for calculating price. For other minor things, we can still use `uint64`. It's just a type-safety mesaure.

Saying it differently: `abi.PaddedPieceSize` always indicates padded data which is the real size it will take in the miner's sector.

ffs/integrationtest/scheduler/scheduler_test.go
```diff
@@ -128,5 +128,5 @@ func TestFailedJobMessage(t *testing.T) {
 	de := job.DealErrors[0]
 	require.False(t, de.ProposalCid.Defined())
 	require.NotEmpty(t, de.Miner)
-	require.Equal(t, "failed to start deal: cannot propose a deal whose piece size (4096) is greater than sector size (2048)", de.Message)
+	require.Equal(t, "data doesn't fit in a sector", de.Message)
```
jsign, line 131: Lotus API changed error string.

go.mod
```diff
@@ -60,5 +61,3 @@ require (
 )
 
 replace github.com/dgraph-io/badger/v2 => github.com/dgraph-io/badger/v2 v2.2007.2
-
-replace github.com/filecoin-project/filecoin-ffi => github.com/filecoin-project/filecoin-ffi v0.30.4-0.20200910194244-f640612a1a1f
```
jsign, line 64: Gone.

reputation/reputation.go
```diff
@@ -252,7 +252,7 @@ func calculateScore(addr string, mi miner.IndexSnapshot, si faults.IndexSnapshot
 		askScore = 1
 	}
 
-	score := 50*faultsScore + 20*powerScore + 20*externalScore + 10*askScore + 1000*float64(miner.ActiveDeals)
+	score := 50*faultsScore + 20*powerScore + 20*externalScore + 100*askScore
```
jsign, line 255: Still arbitrary.

asutula (approved):
Awesome!

PR #712, opened 2020-11-19
GC & multitenancy

jsign:
A TL;DR of this PR, more details below:
- Now it should be safe to run GC in the IPFS node at any point in time without any coordination with Powergate: e.g: with a cronjob or similar. 
- Now the FFS subsystem is fully-multitenant, so different FFS instances can store the same Cid. This means each of them, for the same cid, will have their own log output, _StorageInfo_, reference counting and correct real pin/unpinning in hot-storage, etc.
- A new migration system was built to migrate existing non-multitenant data to new multi-tenant datastore schemas.
- Same as the above point for reference counting regarding pinning cids in go-ipfs.
- Some improvements/fixes in the wild. 

---

**Details about GC safety**

The main _problem_ with having safe IPFS GC at any point in time is all the staging data that might not be actually have ran its corresponding job.

The solution to this is to allow _HotStorage_ have two types of pins:
- Staging pins.
- Real pins.

Now every time you `pow data stage`, it gets pinned in the underlying `go-ipfs` node. Internally, Powergate keeps track of which Cids are pinned in which way.

For example, say that FFS instance IID1 stages some data with cid Cid1. From that moment Cid1 is staged in the IPFS node. Then we have two cases:
1. The user pushes a config for Cid1.
2. For some reason the user abandons that staged data and never pushes a config.

In case 1, when the Job ends up running we have two cases:
1.a - The config that was pushed had _HotStorage_ enabled. In this case, Cid1 is now switched to be a "real pin" instead of a "staging pin".
2.a - The config that was pushed had _HotStorage_ disabled. In this case, Cid1 gets unpinned from the go-ipfs node.

In summary for case 1, after the Job runs the staged Cid1 gets promoted to a "real pin" or gets unpinned. Working this way means that IPFS GC can run any point in time after the data was staged and it's always safe to do so, i.e: no _needed_ data that a queued or executing Job need will be deleted by IFPS.

In case 2, we have a situation of pinned data that would never execute a Job thus, a priori, never gets unpinned without extra logic. To handle this case, Powergate now has a "Staging GC" cron that runs every X minutes (configurable via flags) to manage these cases.

How this GC works? It basically analyzes all staged Cids that:
- Were staged at least 1hr ago.
- That Cid isn't part of a Queued or Executing job.

Those Cids look OK to be unpinned and let be GCed. Saying it differently, if we have some data which is "Stage pinned" for a "long" time, and we don't have any scheduled (queued) Job or executing one, then we can assume these staged data is abandoned. It's important this check that the Cid isn't part of a Queued job, since if we have a long queue of pending Jobs there's the chance that staged data is staged for more than 1hr.

There're knobs via flags/env to configure:
- `--ffsgcinterval`: Frequency in which the Powergate GC runs to unpin old/abandoned staged data. (15min by default)
- `--ffsgcstagegraceperiod`: How old staged data should be to be considered potentially unpinnable (1hr by default)

I also included two extra admin CLI commands for GC related things:
- `pow admin data gcstaged`: This is for doing a manual trigger of a Powergate GC unpinning run. (What gets automatically run every 15min). Might be useful if the user wants to disable automatic unpinning, and just want to run it manually (or via API) for some reason.
- `pow admin data pinnedCids`: This will dump all pinned Cids from Hot-Storage. Each Cid will say which FFS instances is pinning the data, and which kind of pin those are. e.g: we could see that Cid1 is pinned by FFS1 with a Staged Pin, and by FFS2 with a real pin. Just to have some visibility of why a Cid is pinned in IPFS, since if you see the `ipfs pin ls` output you only see that Cid1 is pinned but you might not understand why it's the case.

Exampleof `pinnedCids` output:
```
{
  "cids":  [
    {
      "cid":  "QmeBWFEUxvb25v9HrjoXe8c39i3f47Ge71wySGM6RowuwA",
      "users":  [
        {
          "userId":  "7a3c326d-5ccd-4375-bf1b-aca176e8b8ba",
          "staged":  true,
          "createdAt":  "1606242965"
        }
      ]
    }
  ]
}
```

Finally, as expected, if we have a Cid pinned by multiple FFS instances... only when all FFS instances unpin the Cid, it gets really unpinned from the go-ipfs node. Saying it differently, now _HotStorage_ does reference counting to really know when a Cid can be unpinned from the IPFS node.

---

**Details about multitenancy**

Now we support different FFS instances storing the same Cid. This involved changes in multiple internal components of the scheduler:
- _Trackstore_: The component that keeps track of which Cids have _repair_ or _renew_ enabled capabilities. Since now more than one FFS instance can store the same Cid, then we might have to run repairing or renewals for multiple FFSs for the same Cid.
- _StorageInfo_ store: Since multiple FFSs instances will save hot and cold information about the current state of storage, this can't be only keyed by Cid but (Cid,FFS instance).
- _JobLogger_: _JobLogger_ is the component that has the human friend log of a Job. Since now we have multiple FFSs instances pushing to Filecoin the same Cid, then we might have a different stream of logs for the same Cid. This fixes some weird behavior that some might see of having a second instance push a config for an existing Cid from another instance, and magically seeing Job logs for that Cid (which are from the other instance).

---

**Migrations library**

A new `migration` package was created that keeps track of which version the underlying datastore is. It also registers different `Migration` functions that transition version X to X+1. These Migrations are run in datastore transaction, so all the logic executed in the migration is atomic so it can fail safely without leaving "half-baked" data transformations.

Whenever Powergate runs, the first thing it does after initializing the underlying Datastore is ensuring the datastore version is equal to the latest version. If that isn't the case, it will run all the transformations to move the current version to the latest version.

The current version of the datastore is assumed to be in `/migration/version`.
We have to handle two special cases if Powergate recognizes that `/migration/version` key doesn't exist:
- We're running a Powergate that has data on a version before this migration setup exists. (e.g: all existing Powergates up to now). In this case, the datastore is assumed to be in version 0.
- We're running a completely clean Powergate for the first time.

In both cases, the `/migration/version` key doesn't exist. To distinguish both, the library checks if the datastore has at least 1 key. If there's at least 1 key in the datastore but `/migration/version` doesn't exist, then we're in the first case. If the datastore is completely empty, we're in the last case.

The distinction is necessary since the first case should run the necessary migrations to move from version 0, to the latest version that the newest migration migrates to. If we're in the second case (Powergate run for the firs time), we avoid any migration logic and set `/migration/key` value directly to the latest version.

As mentioned above, all existing Powergates are considered to be in version 0 now. In this PR there's a single migration function that has 4 sub-migrations:
1. JobLogger migration: This migration reads existing job logs for a Cid, and migrates them to the new keyspace format, that is from: `/ffs/joblogger/<cid>/<timestamp>` to `/ffs/joblogger/<ffsid>/<cid>/<timestamp>`. Since we can already have multiple FFS instances storing the same Cid, the migration automatically duplicates entries for all FFS instances that might be having the same Cid... so we don't leave any FFS instance "empty" on logs that it might already existed from its POV. (Hope makes sense, but I can extend if isn't clear).
2. StorageInfo migration: Same as above. The StorageInfo entries are migrated to a new key namespace, and every existing StorageInfo entry that is referenced in many FFS instances is copied once per instance, so we don't leave FFS instances orphaned on shared Cids StorageInfo.
3. TrackStore migration: Same as above. We migrate data to acknowledge that multiple FFS instances might be setting repair and renew features in their storage configs, and all should be tracked for repair/renew crons.
4. Pinstore filling: The migration detects all FFS instances that have _HotStorage_ enabled, and fills the Pinstore with that information... so to recreate the correct pin reference counting in _HotStorage_.

Most of the above logic is a bit confusing since migrations that involve transforming a non-multitenant to multitenant setup is a bit hairy. These transformations not only copy the data to a new key but also have to copy the data multiple times once for each FFS instance that makes sense.

All this is covering the full set of cases that might appear. Most probably 90% of the Powergate instances weren't pushing the same Cid in different FFS instances... but this is the correct way to do the migrations.


ffs/scheduler/scheduler.go
```diff
@@ -163,6 +175,50 @@ func (s *Scheduler) Cancel(jid ffs.JobID) error {
...
+// PinnedCids returns the pinned cids from Hot-Storage.
+func (s *Scheduler) PinnedCids(ctx context.Context) ([]ffs.PinnedCid, error) {
+	res, err := s.hs.PinnedCids(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("getting pinned cids from hot-storage: %s", err)
+	}
+	return res, nil
+}
+
+func (s *Scheduler) gcStaged(ctx context.Context, gracePeriod time.Duration) ([]cid.Cid, error) {
```
jsign, line 192: The method that gets run by cron Powergate GC or the manual API call from the admin user.

ffs/scheduler/scheduler.go
```diff
@@ -163,6 +175,50 @@ func (s *Scheduler) Cancel(jid ffs.JobID) error {
...
+	}
+	ej := s.sjs.ExecutingJobs(ffs.EmptyInstanceID)
+	for _, j := range ej {
+		cids[j.Cid] = struct{}{}
+	}
+
+	excludedCids := make([]cid.Cid, 0, len(cids))
+	for c := range cids {
+		excludedCids = append(excludedCids, c)
+	}
```
jsign, line 210: We exclude all the Cids that are in queued or executing jobs.

ffs/scheduler/scheduler.go
```diff
@@ -163,6 +175,50 @@ func (s *Scheduler) Cancel(jid ffs.JobID) error {
...
+
+	gced, err := s.hs.GCStaged(ctx, excludedCids, time.Now().Add(-gracePeriod))
+	if err != nil {
+		return nil, fmt.Errorf("hot-storage gc: %s", err)
+	}
```
jsign, line 215: We ask Hot-Storage to GC all Cids that *only* have staged-pins, aren't in the exclusion list, and were staged-pinned before the grace-period.

ffs/scheduler/scheduler.go
```diff
@@ -176,15 +232,36 @@ func (s *Scheduler) Close() error {
...
+			select {
+			case <-s.ctx.Done():
+				return
+			case <-time.After(s.gc.AutoGCInterval):
+				if _, err := s.gcStaged(s.ctx, s.gc.StageGracePeriod); err != nil {
+					log.Errorf("automatic gc: %s", err)
+				}
+			}
+		}
+	}()
```
jsign, line 258: New daemon that runs Staged pins GC.
If `AutoGCInterval` is zero, is considered disabled.

ffs/scheduler/scheduler.go
```diff
@@ -396,28 +459,29 @@ func (s *Scheduler) executeQueuedStorage(j ffs.StorageJob) {
...
 		s.cancelLock.Lock()
-		delete(s.cancelChans, j.ID)
+		delete(s.jobsCancel, j.ID)
 		s.cancelLock.Unlock()
 	}()
 
 	ctx, cancel := context.WithCancel(context.WithValue(context.Background(), ffs.CtxKeyJid, j.ID))
 	defer cancel()
 	ctx = context.WithValue(ctx, ffs.CtxStorageCid, j.Cid)
+	ctx = context.WithValue(ctx, ffs.CtxAPIID, j.APIID)
```
jsign, line 473: We add `APIID` to the ctx that is used by Jobs, so all messages logged are stored for the corresponding FFS instance.

ffs/scheduler/scheduler.go
```diff
@@ -396,28 +459,29 @@ func (s *Scheduler) executeQueuedStorage(j ffs.StorageJob) {
...
 	var canceled bool
 	go func() {
 		// If the user called Cancel to cancel Job execution,
 		// we cancel the context to finish.
 		<-cancelChan
-		cancel()
 		cancelLock.Lock()
 		canceled = true
 		cancelLock.Unlock()
+		cancel()
```
jsign, line 484: A fix here. We first should set `canceled = true` and then cancel the context. Doing otherwise might produce canceled logic to still read `canceled = true` if they run fast enough.

ffs/scheduler/scheduler_storage.go
```diff
@@ -203,34 +204,47 @@ func (s *Scheduler) StorageConfig(jid ffs.JobID) (ffs.StorageConfig, error) {
...
+	var hot ffs.HotInfo
+	if a.Cfg.Hot.Enabled {
+		s.l.Log(ctx, "Executing Hot-Storage configuration...")
+		hot, err = s.executeEnabledHotStorage(ctx, a.APIID, ci, a.Cfg.Hot, a.Cfg.Cold.Filecoin.Addr, a.ReplacedCid)
+		if err != nil {
+			s.l.Log(ctx, "Enabled Hot-Storage excution failed.")
+			return ffs.StorageInfo{}, nil, fmt.Errorf("executing enabled hot-storage: %s", err)
+		}
+		s.l.Log(ctx, "Hot-Storage configuration ran successfully.")
 	}
```
jsign, line 227: If HotStorage is enabled, then we run the hot-storage logic right away which will promote the potentially Staged-pinned Cid (from `pow data stage`) to a full pin.

I say _potentially_ since the Cid might not be staged, and just trusted to be discovered in the IPFS network.

If HotStorage isn't enabled, we run HotStorage logic after ColdStorage logic execution to avoid HotStorage logic unpin the data.

ffs/scheduler/scheduler_storage.go
```diff
@@ -203,34 +204,47 @@ func (s *Scheduler) StorageConfig(jid ffs.JobID) (ffs.StorageConfig, error) {
...
+	s.l.Log(ctx, "Cold-Storage configuration ran successfully.")
+
+	if !a.Cfg.Hot.Enabled {
+		s.l.Log(ctx, "Executing Hot-Storage configuration...")
+		if err := s.executeDisabledHotStorage(ctx, a.APIID, a.Cid); err != nil {
+			s.l.Log(ctx, "Disabled Hot-Storage execution failed.")
+			return ffs.StorageInfo{}, nil, fmt.Errorf("executing disabled hot-storage: %s", err)
+		}
+		s.l.Log(ctx, "Hot-Storage configuration ran successfully.")
+	}
```
jsign, line 244: If HotStorage wasn't enabled, now run hot-storage execution which will unpin the Stage-pin potentially set by a staging flow.

To be clear:
- If HotStorage is enabled, this code first executes HotStorage logic since it will promote the Staged-pin to a full pin... so it's still safe to run ColdStorage afterwards (since the data is still pinned).
- If HotStorage is disabled, we avoid running HotStorage logic first since that would unpin the data and potentially make ColdStorage logic be unsafe to run correctly (if IPFS GC runs in the middle). So we run it afterwards so it can unpin the data safely since cold storage things finished already.

migration/migration.go
```diff
@@ -0,0 +1,175 @@
+package migration
```
jsign, line 1: This is the new `migration` library.

migration/migration.go
```diff
@@ -0,0 +1,175 @@
...
+type Migrator struct {
+	ds         datastore.TxnDatastore
+	migrations map[int]Migration
+}
+
+// Migration runs a vA->v(A+1) migration.
+type Migration func(datastore.Txn) error
+
+// New returns a new Migrator.
+func New(ds datastore.TxnDatastore, migrations map[int]Migration) *Migrator {
```
jsign, line 29: Gets created with the underlying datastore, and a map of versions and `Migration`.
i.e:
`{1: functionThatMigratesFromV0ToV1, 2: functionThatMigratesFromV1ToV2, etc}`

migration/migration.go
```diff
@@ -0,0 +1,175 @@
...
+
+// Migrator ensures a datastore goes through all the needed
+// Migrations to upgrade its current version to the latest version.
+type Migrator struct {
+	ds         datastore.TxnDatastore
+	migrations map[int]Migration
+}
+
+// Migration runs a vA->v(A+1) migration.
+type Migration func(datastore.Txn) error
```
jsign, line 26: A `Migration` is any function that receives a `txn` and does whatever it wants to migrate from version X to (X+1).
This function doesn't have to commit the txn, just do its logic. The library will automatically commit or discard the transaction depending if the migration returned an error or not.

migration/migration.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil
+	}
+
+	for i := currentVersion + 1; i <= targetVersion; i++ {
+		log.Infof("Running %d migration...")
+		if err := m.run(i); err != nil {
+			return fmt.Errorf("running migration %d: %s", i, err)
+		}
+		log.Infof("Migration %d ran successfully")
+	}
```
jsign, line 70: Each needed migration will run in its own transaction, so every version bumping should be safe.

migration/migration.go
```diff
@@ -0,0 +1,175 @@
...
+		return 0, false, fmt.Errorf("getting version from datastore: %s", err)
+	}
+	if err := json.Unmarshal(buf, &current); err != nil {
+		return 0, false, fmt.Errorf("unmarshaling current version: %s", err)
+	}
+
+	return current.Version, false, nil
+}
+
+func (m *Migrator) run(version int) error {
```
jsign, line 106: Runs a `Migration` to move from version X to X+1.

migration/migration.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	defer txn.Discard()
+
+	script, ok := m.migrations[version]
+	if !ok {
+		return fmt.Errorf("migration script not found")
+	}
+
+	if err := script(txn); err != nil {
+		return fmt.Errorf("running migration script: %s", err)
```
jsign, line 119: If `script` (`Migrator` func) fails, it will mean that the deferred `txn.Discard()` will rollback the txn so all partial work is undone.

migration/migration.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return fmt.Errorf("marshaling new version: %s", err)
+	}
+	if err := txn.Put(keyCurrentVersion, newVerBuf); err != nil {
+		return fmt.Errorf("saving new version: %s", err)
+	}
+
+	if err := txn.Commit(); err != nil {
+		return fmt.Errorf("committing transaction: %s", err)
+	}
```
jsign, line 133: Bump the version key to the +1 version, and then commit the txn; leaving things ready for the next migration (if applies) to run in a future new txn.

migration/migration.go
```diff
@@ -0,0 +1,175 @@
...
+
+	return nil
+}
+
+func (m *Migrator) isDSEmpty() (bool, error) {
```
jsign, line 138: To detect what was mentioned in the PR description: the case of a new clean Powergate vs a Powergate with data but no datastore version key (since this migration library didn't exist).

migration/migration.go
```diff
@@ -0,0 +1,175 @@
...
+
+	all, err := res.Rest()
+	if err != nil {
+		return false, fmt.Errorf("getting query results: %s", err)
+	}
+
+	return len(all) == 0, nil
+}
+
+func (m *Migrator) getTargetVersion() int {
```
jsign, line 154: The `Migrator` always selects the target version to the latest version in which exists a `Migrator`.

migration/migration.go
```diff
@@ -0,0 +1,175 @@
...
+	newVerBuf, err := json.Marshal(newVer)
+	if err != nil {
+		return fmt.Errorf("marshaling new version: %s", err)
+	}
+	if err := m.ds.Put(keyCurrentVersion, newVerBuf); err != nil {
+		return fmt.Errorf("saving new version: %s", err)
+	}
+
+	return nil
+}
```
jsign, line 175: If we were in the case of a new clean Powergate, then just create the datastore version with the latest version.

migration/migration1.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/query"
+	"github.com/textileio/powergate/ffs"
+	"github.com/textileio/powergate/util"
+)
+
+// V1MultitenancyMigration contains the logic to upgrade a datastore from
+// version 0 to version 1.
+var V1MultitenancyMigration = Migration(func(txn datastore.Txn) error {
```
jsign, line 16: I think this filename and migrator name is quite clear. 

This is the only migration that exists now. Inside this migration, multiple sub-migrations are run. Recall that this function is an all-or-nothing executing. The "sub-migration" functions is just some name I invented in this comment, from the POV of the `Migrator` it is a single one.

migration/migration1.go
```diff
@@ -0,0 +1,137 @@
...
+	log.Infof("Starting trackstore migration...")
+	if err := migrateTrackstore(txn, cidOwners); err != nil {
+		return fmt.Errorf("migrating trackstore: %s", err)
+	}
+	log.Infof("Trackstore migration finished")
+
+	return nil
+})
+
+func v0CidOwners(txn datastore.Txn) (map[cid.Cid][]ffs.APIID, error) {
```
jsign, line 43: This function (and below ones) are helpers to the sub-migrations to know which APIIDs own Cids.
In many migrations we need to insert data for all the FFS instances that are storing a Cid.

migration/migration1_joblogger.go
```diff
@@ -0,0 +1,58 @@
+package migration
```
jsign, line 1: All migrations are a bit similar, and I added extra comments to explain.

migration/migration1_joblogger_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/textileio/powergate/tests"
+	"github.com/textileio/powergate/util"
+)
+
+func TestV1_JobLogger(t *testing.T) {
+	t.Parallel()
+
+	ds := tests.NewTxMapDatastore()
+
+	pre(t, ds, "testdata/v1_JobLogger.pre")
```
jsign, line 18: First we load key-values described in `testdata/v1_JobLogger.pre`.

migration/migration1_joblogger_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+	c1, _ := util.CidFromString("QmPewMLNZEgnLxaenjo9Q5qwQwW3zHZ7Ac973UmeJ6VWHE")
+	c2, _ := util.CidFromString("QmZTMaDfCMWqhUXYDnKup8ctCTyPxnriYW7G4JR8KXoX5M")
+	cidOwners := map[cid.Cid][]ffs.APIID{
+		c1: {ffs.APIID("ID1"), ffs.APIID("ID2")},
+		c2: {ffs.APIID("ID3")},
+	}
+	txn, _ := ds.NewTransaction(false)
+	err := migrateJobLogger(txn, cidOwners)
```
jsign, line 27: We run this sub-migration on the datastore.

migration/migration1_joblogger_test.go
```diff
@@ -0,0 +1,32 @@
...
+	require.NoError(t, err)
+	require.NoError(t, txn.Commit())
+
+	post(t, ds, "testdata/v1_JobLogger.post")
```
jsign, line 31: We assert that the end result of the datastore state is the same as described in `v1_JobLogger.post`.

This style of doing test on migrations is quite easy since most of it is creating the `pre` and `post` state files.

migration/migration_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"strings"
+	"testing"
+
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/query"
+	"github.com/stretchr/testify/require"
+	"github.com/textileio/powergate/tests"
+)
+
+func TestEmptyDatastore(t *testing.T) {
```
jsign, line 17: Here some tests of general `Migrator` library.
In this test we check the case of a new pristine Powergate. In this case, only the version will be set to the latest known version since doesn't make sense to run any migration.

migration/migration_test.go
```diff
@@ -0,0 +1,159 @@
...
+	t.Parallel()
+
+	ms := map[int]Migration{
+		1: func(_ datastore.Txn) error {
+			return fmt.Errorf("This migration shouldn't be run on empty database")
```
jsign, line 22: We define some migration that always fail... so this way we can test that no migrations are run and just the version is set to the lastest value.

migration/migration_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	err = m.Ensure()
+	require.NoError(t, err)
+
+	v, _, err = m.getCurrentVersion()
+	require.NoError(t, err)
+	require.Equal(t, 1, v)
+}
+
+func TestNonEmptyDatastore(t *testing.T) {
```
jsign, line 40: Test the case of a Powergate already with data, but running the migration for the first time.
(e.g: all hosted or Hub powergate).

migration/migration_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	err = m.Ensure()
+	require.NoError(t, err)
+
+	v, _, err = m.getCurrentVersion()
+	require.NoError(t, err)
+	require.Equal(t, 1, v)
+}
+
+func TestNoop(t *testing.T) {
```
jsign, line 63: Test the case of the migration running when the version is already the latest.. so most cases in general when Powergate is restarted or no new migrations exist.

migration/migration_test.go
```diff
@@ -0,0 +1,159 @@
...
+	require.NoError(t, err)
+
+	err = m.Ensure()
+	require.NoError(t, err)
+}
+
+func TestFailingMigration(t *testing.T) {
+	ms := map[int]Migration{
+		1: func(_ datastore.Txn) error {
+			return fmt.Errorf("I failed")
```
jsign, line 83: Simulate a failing migration and ensure that the version isn't bumped.

migration/testdata/v1_JobLogger.post
```diff
@@ -0,0 +1,3 @@
+/ffs/joblogger/ID1/QmPewMLNZEgnLxaenjo9Q5qwQwW3zHZ7Ac973UmeJ6VWHE/1602952162298722885,{"Cid":{"/":"QmPewMLNZEgnLxaenjo9Q5qwQwW3zHZ7Ac973UmeJ6VWHE"},"RetrievalID":"","Timestamp":1602952162298722885,"Jid":"ad6da2a0-5465-4275-a330-2537062765c8","Msg":"Deal 763168 with miner f022142 is active on-chain"}
```
jsign, line 1: `pre` and `post` files are CSV values of `(key,value)`.

Most of these cases are real that I dumped from some hosted powergates and in some cases tuned a bit more for interesting cases or easier IIDs.

migration/migration.go
```diff
@@ -0,0 +1,179 @@
...
+
+	log.Infof("Current datastore version is %d, target version %d", currentVersion, targetVersion)
+
+	if currentVersion == targetVersion {
+		return nil
+	}
+
+	if currentVersion > targetVersion {
+		return fmt.Errorf("migrations are forward only, current version %d, target version %d", currentVersion, targetVersion)
+	}
```
jsign, line 66: Currently, this `Migrator` does forward only migrations. So check here for a situation in which the user might want to go back to an older tag of Powergate which isn't compatible with a newer datastore.

We can extend this lib to do forward and backward migrations. Now I did only forward since doesn't make sense for these big migrations of the PR to be rollbacked since that's impossible to do correctly.

We could define `Migrations` as rollback-able or not, and potentially allow backward migrations depending if the migration has a backward logic or not. That would be easy to do. Mentioning it just to consider for the future.

migration/migration.go
sanderpick: IIRC, MongoDB transactions are only available to replica sets. Just mentioning in case you've implemented `TxnDatastore` with `go-ds-mongo`.

jsign: Yes, that's correct. But our MongoDB cluster is a replica set, right?

sanderpick (commented):
Very nice! Lots of work. I read the description and comments to get an idea of the changes.

migration/migration.go
sanderpick: Yep, it is. Not a problem, just mentioning. It would error and cleanly exit your migration if run on a single mongod

jsign: Oh yeah, sure. In fact, in `go-ds-mongo` tests, the tests that involve Txn have to run in a local replica-set that I documented how to spinup since was kind of hairy. (And maybe is there some simpler solution), but yeah, a single `docker run mongo` will fail.

sanderpick: Oh I see. I did a similar thing when starting the billing work, where I was relying on transactions (not any more). You can actually make a replica set with one member. I had a docker-compose file that does that, then runs the initiate command, making for a one liner. Maybe I can add that when doing the go-threads -> go-ds-mongo work.

jsign: Cool, that sounds good to add to `go-ds-mongo` easier testing yep.

Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN mkdir /app
 WORKDIR /app 
 
 COPY go.mod go.sum ./
-RUN go mod download
+RUN go mod download -x
```
jsign, line 7: Since downloading deps can take some time, I prefer seeing detailed output about what's being donwloaded.

api/server/server.go
```diff
@@ -189,13 +189,13 @@ func NewServer(conf Config) (*Server, error) {
 		}
 	}
 
-	ds, err := createDatastore(conf)
-	if err != nil {
-		return nil, fmt.Errorf("creating datastore: %s", err)
+	if err := runMigrations(conf); err != nil {
+		return nil, fmt.Errorf("running migrations: %s", err)
 	}
```
jsign: Before spinning internal components, we run migrations logic.

api/server/server.go
```diff
@@ -247,7 +247,7 @@ func NewServer(conf Config) (*Server, error) {
 		return nil, fmt.Errorf("creating miner selector: %s", err)
 	}
 
-	l := joblogger.New(txndstr.Wrap(ds, "ffs/joblogger"))
+	l := joblogger.New(txndstr.Wrap(ds, "ffs/joblogger_v2"))
```
jsign, line 250: More about the need of suffixing migrated key namespaces soon.

api/server/server.go
```diff
@@ -537,15 +537,19 @@ func (s *Server) Close() {
 	}
 }
 
-func createDatastore(conf Config) (datastore.TxnDatastore, error) {
+func createDatastore(conf Config, longTimeout bool) (datastore.TxnDatastore, error) {
```
jsign, line 540: We enable a new flag `longTimeout` which returns a datastore that has long timeouts for operations and transactions. 
Basically, the datastore that Powergate uses for operating has _false_ value. For migrations, we create use a _true_ value to allow for bigger timeouts since we do long-running operations.

api/client/utils_test.go
```diff
@@ -25,9 +24,7 @@ func defaultServerConfig(t *testing.T) server.Config {
 	gatewayHostAddr := fmt.Sprintf("0.0.0.0:%d", freePort(t))
 	indexRawJSONHostAddr := fmt.Sprintf("0.0.0.0:%d", freePort(t))
 
-	repoPath, err := ioutil.TempDir("/tmp/powergate", ".powergate-*")
-	require.NoError(t, err)
-
+	repoPath := t.TempDir()
```
jsign, line 27: In Go 1.15, we can use `(*testing.T).TempDir()` which automatically cleans the folder when the test finishes, so no need to leave folders in tmp, or return "cls" funcs, and also deletes the folder even if the test panics.
We can use this in lots of places, just changed here.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/textileio/powergate
 
-go 1.14
+go 1.15
```
jsign, line 3: Since we use `t.TempDir()` only available in >Go 1.15.

migration/migration.go
```diff
@@ -0,0 +1,198 @@
...
+	datastore.Read
+	datastore.Write
+}
+
+// Migration runs a vA->v(A+1) migration. UseTxn indicates
+// if this migration should be run in a transaction.
+type Migration struct {
+	Run    func(datastoreReaderWriter) error
+	UseTxn bool
+}
```
jsign, line 35: A `Migration` is some logic that does a transformation between version X and X+1.
It also describes if it should run in a txn or not. This is to solve that long/big-running migration can't be transactional in Bager or Mongo (Atlas) for reasons explained in Slack.

migration/migration1.go
```diff
@@ -0,0 +1,175 @@
...
+	if len(errors) > 0 {
+		for _, m := range errors {
+			log.Error(m)
+		}
+		return nil, fmt.Errorf("building cidowners had %d errors", len(errors))
+	}
+
+	return owners, nil
+}
+
```
jsign, line 95: All below `v0` functions are helpers for the migration stuff, which resolves some important questions for migrations:
- Which iids exists.
- Which iids own a Cid
- Get the storage config of a Cid for a iid.

These methods should be agnostic of internal components, since if that is the case we could handcuff improving internal components in the future since migrations rely on them.

migration/migration_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+	err = m.Ensure()
+	require.Error(t, err)
+
+	v, _, err = m.getCurrentVersion()
+	require.NoError(t, err)
+	require.Equal(t, 0, v)
+}
+
+func TestRealDataBadger(t *testing.T) {
```
jsign, line 144: I exported the `go-datastore` of one of our internal Powergate stacks locally in a Badger datastore.
And then just tested e2e with that real datastore that the migration runs correctly.

migration/migration_test.go
```diff
@@ -0,0 +1,332 @@
...
+	migrations := map[int]Migration{
+		1: V1MultitenancyMigration,
+	}
+	m := New(ds, migrations)
+	err = m.Ensure()
+	require.NoError(t, err)
+}
+
+func TestRealDataRemoteMongo(t *testing.T) {
+	t.SkipNow()
```
jsign, line 165: This test case was mostly to test the migration directly in copied datastores of many live hosted powergates.
Disabled the test since doesn't make sense for CI.

migration/migration_test.go
```diff
@@ -0,0 +1,332 @@
...
+		expected[parts[0]] = []byte(parts[1])
+	}
+
+	require.Equal(t, len(expected), len(current))
+	for k1, v1 := range current {
+		v2, ok := expected[k1]
+		require.True(t, ok)
+		require.Equal(t, v2, v1)
+	}
+}
```
jsign, line 238: `pre` is a function that reads a csv files with key,values as a initial state of a datastore for tests.
`post` is a function that asserts if a datastore has the same state as a csv file with key,values.

migration/testdata/v1_JobLogger.post
```diff
@@ -0,0 +1,3 @@
+/ffs/joblogger_v2/ID1/QmPewMLNZEgnLxaenjo9Q5qwQwW3zHZ7Ac973UmeJ6VWHE/1602952162298722885,{"Cid":{"/":"QmPewMLNZEgnLxaenjo9Q5qwQwW3zHZ7Ac973UmeJ6VWHE"},"RetrievalID":"","Timestamp":1602952162298722885,"Jid":"ad6da2a0-5465-4275-a330-2537062765c8","Msg":"Deal 763168 with miner f022142 is active on-chain"}
```
jsign, line 1: Here the pre and post files for each test case.
A new datastore is populated with `.pre` file key-values, the migration is run, and then it's asserted if the datastore is exactly in the `.post` key-values state.

migration/migration1_joblogger.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+			timestampStr := originalKey.Namespaces()[3]
+
+			owners := cidOwners[cid]
+
+			// Step 1/2:
+			// For each cid owner, we create the same registry
+			// in the new key version.
+			for _, iid := range owners {
+				newKey := datastore.NewKey("/ffs/joblogger_v2/").ChildString(iid.String()).ChildString(cidStr).ChildString(timestampStr)
```
jsign, line 52: Here the detail about `_v2` suffixes.
What's happening in many migrations is that we try doing something like:
- Original key: `/hello/world/<cid>`
- New key: `/hello/world/<iid>/<cid>`

For doing the migration, we scan all original keys with `Query{Prefix:"/hello/world"}`.

Depending on the underlying way that the `go-datastore` implementation works, that can produce problems; and that was the case, since the new keys we're `.Put`ing are also a result of the query itself.

That's to say, we start getting `/hello/world/<cid>` keys, but at the same time we're iterating that query we're inserting others `/hello/world/<iid>/<cid>` which will appear later in the query iteration and is *not* something we want (new keys satisfy the prefix from the query). We only want to iterate the original keys.

Not a big deal and is solved with the `_v2` suffixing to avoid the iteration to consider the new keys.

In the Badger implementation of `go-datastore` this doesn't happen, since Badger has snapshot capabilities on the query execution and you won't see new keys that appear after the query was started. But in Mongo that isn't the case. It's just different ways that queries work in different databases.

ffs/scheduler/internal/trackstore/trackstore.go
```diff
@@ -18,41 +18,26 @@ import (
...
-	}
-	if err := s.loadCaches(); err != nil {
-		return nil, fmt.Errorf("loading renewable/repairable caches: %s", err)
-	}
-	return s, nil
+// TrackedCid contains tracked storage configs for a Cid.
+type TrackedCid struct {
+	Cid     cid.Cid
+	Tracked []TrackedStorageConfig
 }
```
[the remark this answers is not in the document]
asutula: I guess I'm not familiar with the implementation of repair/renew, but just curious why storing a list per cid is better than using a `instanceId/cid` key with a `StorageConfig` value, avoiding that complexity you mentioned.

ffs/scheduler/scheduler.go
```diff
@@ -163,6 +175,50 @@ func (s *Scheduler) Cancel(jid ffs.JobID) error {
 	return nil
 }
 
+// GCStaged runs a unpinned garbage collection of stage-pins.
+func (s *Scheduler) GCStaged(ctx context.Context) ([]cid.Cid, error) {
+	return s.gcStaged(ctx, 0)
```
asutula, line 180: Maybe I'm missing something here, but would expect `0` to be `s.gc.StageGracePeriod`

migration/migration1_joblogger.go
asutula: Tricky

asutula (approved):
Awesome work. This all makes good sense to me. Left a couple questions, one potential issue.

asutula:
Oh, meant to ask about `Client.StageFolder`... That thing uses the IPFS HTTP client directly to the IPFS API we're proxying. So that would bypass all the pin counting logic. Problem?

ffs/scheduler/internal/trackstore/trackstore.go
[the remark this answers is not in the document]
jsign: Yeah, this was mostly for avoiding changing the original keyspace namespace structure since at that time I didn't have all the migration library stuff that did afterward; so thought originally was a better idea. After having the migration library, looks like might be a better idea.

Anyway, I don't like the idea of this `Trackstore` in general. I believe in the future we should take it out, and maybe only save the `iid`, and find a reasonable way to query the corresponding API for the `StorageConfig`. Since here we're kind of duplicating that info. Still feels weird since that kind of violates the Scheduler knowing about a higher-level concept such as `API` but dunno.

ffs/scheduler/scheduler.go
jsign: Actually, this is on purpose. This API is for the manual GC call that I feel should clean all staged stuff (grace period 0).
I'd imagine that the user would like the manual option to be more direct clean.

ffs/coreipfs/internal/pinstore/pinstore.go
```diff
@@ -0,0 +1,338 @@
...
+// New returns a new Store.
+func New(ds datastore.TxnDatastore) (*Store, error) {
+	cache, err := populateCache(ds)
+	if err != nil {
+		return nil, fmt.Errorf("populating cache: %s", err)
+	}
+	return &Store{ds: ds, cache: cache}, nil
+}
+
+// AddStaged sets  c a stage-pin by iid.
```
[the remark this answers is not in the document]
jsign: Yep, improving in next commit.

jsign:
> Oh, meant to ask about `Client.StageFolder`... That thing uses the IPFS HTTP client directly to the IPFS API we're proxying. So that would bypass all the pin counting logic. Problem?

Yeah, any external API calls to go-ipfs are out of the path of tracking. What we can do is create some other API to stage-pin a Cid. So after it does `ipfs add` externally, the Powergate client calls that new API to stage-pin the resulting Cid.
That would be quite clean and easy. Thoughts?

asutula:
Yea that sounds like a good idea.

ffs/scheduler/scheduler.go
asutula: 👍

api/client/data.go
```diff
@@ -97,13 +97,18 @@ func (d *Data) StageFolder(ctx context.Context, ipfsRevProxyAddr string, folderP
 	defer func() { _ = ff.Close() }()
 	opts := []options.UnixfsAddOption{
 		options.Unixfs.CidVersion(1),
-		options.Unixfs.Pin(false),
+		options.Unixfs.Pin(true),
```
jsign, line 100: So for staging folders we also pin now.

api/client/data.go
```diff
@@ -97,13 +97,18 @@ func (d *Data) StageFolder(ctx context.Context, ipfsRevProxyAddr string, folderP
...
 	pth, err := ipfs.Unixfs().Add(context.Background(), files.ToDir(ff), opts...)
 	if err != nil {
 		return "", err
 	}
 
+	_, err = d.client.StageCid(ctx, &userPb.StageCidRequest{Cid: pth.Cid().String()})
+	if err != nil {
+		return "", fmt.Errorf("stage pinning cid: %s", err)
+	}
+
```
jsign, line 111: We call a new API to stage Cids instead of streams of data.

asutula: Awesome

PR #718, opened 2020-12-02
feat(api): add batch task queue

andrewxhill:
Signed-off-by: Andrew Hill <andrew@textile.io>


cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,297 @@
...
+func resultsToStdOut(rc chan TaskResult) (int, int, int) {
+	jobs := 0
+	deals := 0
+	errs := 0
+	for {
+		select {
+		case res, ok := <-rc:
+			if ok {
+				if res.err != nil {
+					Message("Error: %s %s %s", res.task.path, res.stage, res.err.Error())
```
jsign, line 121: Nit: `%s` automatically calls `.Error()` if available, so you can use `res.err`.

cmd/fts/cmd/staging.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func (sc *StagingConfig) Ready(bytes int64) (bool, error) {
+	sc.m.Lock()
+	defer sc.m.Unlock()
+	if bytes > sc.maxStagingBytes {
+		err := fmt.Errorf("Request larger than available staging limit: %d needed out of %d.", bytes, sc.maxStagingBytes)
+		return false, err
+	}
+	if sc.cachedBytes+bytes >= sc.maxStagingBytes {
```
jsign, line 22: Nit: Use `>` instead of `>=`?

cmd/fts/cmd/tasks.go
```diff
@@ -0,0 +1,232 @@
...
+	}()
+
+	return res
+}
+
+func ProcessTask(task Task, config RunConfig, ch chan bool, res chan<- TaskResult, wg *sync.WaitGroup, sc *StagingConfig) {
+	defer wg.Done()
+
+	// Fill in the channel buffer slot.
+	ch <- true
```
jsign, line 76: I think you'd like to do `defer func() { <- ch}` in the line below. That will get you covered of returning the slot in any code path that happens in the method. Maybe you were careful enough to `break` and be sure of never early returning, but if you do the above, you can forget about messing around with that situation. Usually safer when some other person comes to change the code and doesn't have that in mind.

cmd/fts/cmd/tasks.go
```diff
@@ -0,0 +1,232 @@
...
+
+	sc := StagingConfig{
+		maxStagingBytes: config.maxStagingBytes,
+		minDealBytes:    config.minDealBytes,
+	}
+
+	// wg is used to force the application wait for all goroutines to finish before exiting.
+	wg := sync.WaitGroup{}
+	// ch is a buffered channel for processing tasks
+	ch := make(chan bool, config.concurrent)
```
jsign, line 52: You can also do:
`make(chan struct{}, config.concurrent)`, and `ch < struct{}{}`.

A `struct{}` type has size 0 so consumes no memory.
Of course, just mentioning this as a curious note since you might see it in the willd... it doesn't make any difference in this program.

cmd/fts/cmd/tasks.go
```diff
@@ -0,0 +1,232 @@
...
+		if err != nil {
+			sendErr(err)
+			break
+		}
+		// wait for space
+		if !ready {
+			timeout := 1000 * time.Millisecond
+			time.Sleep(timeout)
+			continue
+		}
```
jsign, line 100: Cool, simple solution. If this becomes a problem, we can think of re-structuring the logic a bit more to only consume a slot of concurrency if the deal size will fit the limit, if not, try other pending ones so as to completely block waiting if we know there's no chance of any progress.

cmd/fts/cmd/tasks.go
```diff
@@ -0,0 +1,232 @@
...
+		res <- taskResult
+	}
+
+	for {
+		// check available space
+		ready, err := sc.Ready(task.bytes)
+		if err != nil {
+			sendErr(err)
+			break
+		}
```
jsign, line 94: Similar-ish `defer func() { sc.Done(task.Bytes) }` to avoid duplicating that in every branch, and future-proofing potentially missing or double-calling.

cmd/fts/cmd/tasks.go
```diff
@@ -0,0 +1,232 @@
...
+		taskResult.records = complete.Records
+		sc.Done(task.bytes)
+		res <- taskResult
+		break
+	}
+	// Free the channel buffer slot.
+	<-ch
+}
+
+func watchStorage(pow *client.Client, token string, jobId string) (*client.WatchStorageJobsEvent, error) {
```
jsign, line 201: We should probably change this implementation. The main problem with using stream APIs is that they have a timeout if no new data is streamed. This is a safety measure in gRPC to avoid long-lived idle connections that can lead to port exhaustion. In the case of Powergate it's highly probable that the connection will be closed and you reach the final return with error, since it can pass multiple hours without new messages.

[In Hub](https://github.com/textileio/textile/blob/master/buckets/archive/tracker.go) we do polling, with shorter intervals for initial stages, and longer ones for "Sealing". 
So basically a for loop with a sleep in this case. In Hub we do something a bit more sophisticated to allow killing the Hub process and let resuming all needed things.

jsign: The "watch log" API maybe was too biased for localnet usage since it's quite fast and has a great UX. But for real usage, it has limited application and might be even misleading as happened here. So something to think about...

jsign (commented):
Looks good!
Nice advanced concepts for a non-expert Go. I left some educational comments and mentioned a particular thing that might need to be changed (cc @asutula  for comments there too).

cmd/fts/cmd/tasks.go
andrewxhill: sweet, gone

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,297 @@
...
+			if maxDealBytes < tsk.bytes {
+				Message("data too large %s (%d)", tsk.name, tsk.bytes)
+				continue
+			}
+			tasks = append(tasks, tsk)
+		}
+
+		Message("Removed %d tasks", len(alltsks)-len(tasks))
+
+		concurrent := viper.GetInt64("concurrent")
```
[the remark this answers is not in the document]
andrewxhill: fixed up a bunch there

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,297 @@
...
+	},
+}
+
+func resultsToStdOut(rc chan TaskResult) (int, int, int) {
+	jobs := 0
+	deals := 0
+	errs := 0
+	for {
+		select {
+		case res, ok := <-rc:
```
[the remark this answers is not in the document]
andrewxhill: nice. used it in a few places. lots of updates to channels use. trying to keep UI things // staging things // watching things fanned out in the right way and only when in use

cmd/fts/cmd/tasks.go
andrewxhill: cleaned that up and created a watching channel that will poll for updates as long as deals are still out there

cmd/fts/cmd/resume.go
```diff
@@ -0,0 +1,101 @@
...
+	resumeCmd.Flags().MarkHidden("concurrent")
+	resumeCmd.Flags().MarkHidden("max-staged-bytes")
+	resumeCmd.Flags().MarkHidden("max-deal-bytes")
+	resumeCmd.Flags().MarkHidden("min-deal-bytes")
+
+	rootCmd.AddCommand(resumeCmd)
+}
+
+var resumeCmd = &cobra.Command{
+	Use:   "resume",
```
andrewxhill, line 32: realized it was more intuitive that running the same tasks again to resume. so generally just, `fts resume` in the same directory should pick up where you left off.

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,170 @@
...
+func init() {
+	runCmd.Flags().BoolVarP(&dryRun, "dry-run", "d", false, "Run through steps without pushing data to Powergate API.")
+	runCmd.Flags().BoolVarP(&mainnet, "mainnet", "n", false, "Sets staging limits based on localnet or mainnet.")
+	runCmd.Flags().BoolVarP(&retryErrors, "retry", "e", false, "Retry tasks with error status")
+	runCmd.Flags().StringVar(&taskFolder, "folder", "", "Folder with organized tasks of directories or files")
+	runCmd.Flags().BoolVarP(&hiddenFiles, "all", "a", false, "Include hidden files & folders from top level folder")
+	runCmd.Flags().StringVar(&ipfsrevproxy, "ipfsrevproxy", "127.0.0.1:6002", "Powergate IPFS reverse proxy multiaddr")
+	runCmd.Flags().StringVarP(&resultsOut, "results", "r", "results.json", "The location to store intermediate and final results.")
+
+	// Not included in the public commands
```
andrewxhill, line 36: below are mostly the dev user settings, not really helpful to have in the public interface

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,170 @@
...
+	taskFolder     string
+	resultsOut     string
+	concurrent     int64
+	ipfsrevproxy   string
+	retryErrors    bool
+)
+
+func init() {
+	runCmd.Flags().BoolVarP(&dryRun, "dry-run", "d", false, "Run through steps without pushing data to Powergate API.")
+	runCmd.Flags().BoolVarP(&mainnet, "mainnet", "n", false, "Sets staging limits based on localnet or mainnet.")
```
andrewxhill, line 29: `mainnet` flag is used below to toggle between two reasonable (?) sets of defaults for staging usage

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,170 @@
...
+
+var runCmd = &cobra.Command{
+	Use:   "run",
+	Short: "Run a storage task pipeline",
+	Long:  `Run a storage task pipeline to migrate large collections of data to Filecoin`,
+	PreRun: func(cmd *cobra.Command, args []string) {
+		err := viper.BindPFlags(cmd.Flags())
+		checkErr(err)
+		// set default variables
+		if mainnet {
```
andrewxhill, line 57: couldn't find a good way in cobra to manage two sets of options like this, so just did the logic

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+		}
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		if maxStagedBytes < maxDealBytes {
+			Fatal(fmt.Errorf("Max deal size (%d) is larger than max staging size (%d)", maxDealBytes, maxStagedBytes))
+		}
+
+		// Read all tasks from the input folder
+		unfiltered, err := pathToTasks(taskFolder)
```
andrewxhill, line 85: reads all files/folders in a folder and turns them into potential tasks.

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,170 @@
...
+		checkErr(err)
+		Message("Input tasks: %d", len(unfiltered))
+		allTasks := filterTasks(unfiltered)
```
andrewxhill, line 88: drops all hidden files (per `.` prefix for now)

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,170 @@
...
+
+		// Ensure there aren't existing tasks in the same queue
+		knownTasks, err := openResults(resultsOut)
```
andrewxhill, line 91: grabs the output file to see if there are still pending tasks from a previous run to resume

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,170 @@
...
+		checkErr(err)
+		Message("Existing tasks: %d", len(knownTasks))
+		allTasks = mergeNewTasks(knownTasks, allTasks)
+
+		pendingTasks := cleanErrors(removeComplete(allTasks), retryErrors)
+
+		Message("Pending tasks: %d", len(pendingTasks))
+		time.Sleep(1 * time.Second)
```
andrewxhill, line 99: added a little sleep here otherwise the summary up to now can happen too fast to be seen at all

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,170 @@
...
+
+		if len(pendingTasks) == 0 {
+			return
+		}
+
+		// Init or update our intermediate results
+		err = storeResults(resultsOut, allTasks)
+		checkErr(err)
```
andrewxhill, line 107: stores a `results.json` file to start maintaining. worked really nicely this way and there is only one writer, so all incremental results get written and can be resumed if the job is killed etc.

cmd/fts/cmd/run.go
```diff
@@ -0,0 +1,170 @@
...
+		var progressWaitGroup sync.WaitGroup
+
+		progress := uiprogress.New()
+		progress.Start()
+		progressBars := make(map[string](chan Task))
+
+		taskUpdates := Start(pendingTasks, conf)
+
+		j := 1
+		for task := range taskUpdates {
```
andrewxhill, line 143: all results are streamed back over this channel and all UI can be managed here. don't love the look right now, but didn't want to get hung up on it before people have tried this thing.

cmd/fts/cmd/staging.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+type StagingStatus struct {
+	cachedBytes     int64
+	maxStagingBytes int64
+	minDealBytes    int64
+	m               sync.Mutex
+}
+
+func (sc *StagingStatus) Ready(bytes int64) (bool, error) {
```
andrewxhill, line 15: same as previous commit, still just tracks available resources on the remote staging

cmd/fts/cmd/task.go
```diff
@@ -0,0 +1,117 @@
...
+import (
+	"sync"
+	"time"
+
+	client "github.com/textileio/powergate/api/client"
+	userPb "github.com/textileio/powergate/api/gen/powergate/user/v1"
+)
+
+// Task is an individual data collection to be stored on Filecoin
+type Task struct {
```
andrewxhill, line 12: becomes the pipeline task object and the log file output format // resume job input format.

cmd/fts/cmd/task.go
```diff
@@ -0,0 +1,117 @@
...
+type queueConfig struct {
+	watch    chan Task
+	updates  chan<- Task
+	complete chan struct{}
+	staging  *StagingStatus
+	pipeWg   *sync.WaitGroup
+	pipe     PipelineConfig
+}
+
+func queue(task Task, conf queueConfig, isLast bool) {
```
andrewxhill, line 36: single threaded, just waits for space for the new job

cmd/fts/cmd/task.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		// wait for space
+		if !ready {
+			time.Sleep(waitTimeout)
+			waitTimeout = waitTimeout * 2
+			continue
+		}
+		// start processing
+		conf.pipeWg.Add(1)
+		go process(task, conf, isLast)
```
andrewxhill, line 56: fires off a new channel per task when ready though

cmd/fts/cmd/task.go
```diff
@@ -0,0 +1,117 @@
...
+		return
+	}
+}
+
+func process(task Task, conf queueConfig, isLast bool) {
```
andrewxhill, line 61: walks through each of the stages of a task. this is resumable, since any stage task will jump forward

andrewxhill: it could get a lot more complex if we dealt with errors a bit more, but works for now

cmd/fts/cmd/task.go
```diff
@@ -0,0 +1,117 @@
...
+	defer func() {
+		if isLast {
+			conf.complete <- struct{}{}
```
andrewxhill, line 64: fires a complete message to the Job Watcher channel, which is a single channel that polls results. This complete message doesn't tell it to stop, just that it can expect to stop getting results soon.

cmd/fts/cmd/task.go
```diff
@@ -0,0 +1,117 @@
...
+	pow, err := client.NewClient(conf.pipe.serverAddress)
+	if err != nil {
+		sendErr(err)
+		return
+	}
+
+	if task.Stage < StagingComplete {
+		sendStage(Staging)
+		if conf.pipe.dryRun {
+			dryRunStep()
```
andrewxhill, line 89: just adds some delay when dry-run

cmd/fts/cmd/task.go
```diff
@@ -0,0 +1,117 @@
...
+			if err != nil {
+				sendErr(err)
+				return
+			}
+			task.JobID = applyRes.JobId
+		}
+		sendStage(PushComplete)
+	}
+
+	conf.watch <- task
```
andrewxhill, line 116: send the task to the watcher loop for polling

cmd/fts/cmd/staging.go
asutula: So just want to understand this... you're keeping a local representation of some known server side limit for the amount of data that can be staged at once? Idea is to just never get to the point where the server side is borked because it's out of disk space?

asutula: And I guess we're assuming that `fts` has exclusive access to the powergate instance

cmd/fts/cmd/watcher.go
```diff
@@ -0,0 +1,127 @@
...
+func getStorageSummary(conf queueConfig) (*userPb.StorageJobsSummaryResponse, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour*2)
+	defer cancel()
+	ctx = context.WithValue(ctx, client.AuthKey, conf.pipe.token)
+
+	pow, err := client.NewClient(conf.pipe.serverAddress)
+	if err != nil {
+		return nil, err
+	}
+	summary, err := pow.StorageJobs.Summary(ctx)
```
asutula, line 126: I'm considering making `Summary` not return all the data because it can get out of hand. Not sure yet, but if anything changes, this may just require a couple different calls to get the same information.

go.mod
```diff
@@ -47,13 +52,16 @@ require (
...
+	github.com/pkg/errors v0.9.1
 	github.com/rs/cors v1.7.0
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.1
+	github.com/sqs/goreturns v0.0.0-20181028201513-538ac6014518 // indirect
 	github.com/stretchr/objx v0.2.0 // indirect
 	github.com/stretchr/testify v1.6.1
 	github.com/textileio/go-ds-mongo v0.1.2
+	github.com/textileio/go-threads v1.0.2
```
asutula, line 64: Not sure what's going on here. Didn't think Powergate uses threads at all...

asutula (approved):
Cool to see something like this built with Powergate! Looks like the API use is only a few calls and actually pretty simple. The rest, however is quite complex. Necessarily complex for sure. Just makes me wonder if this would be better in its own repo. In fact, everything in under the `cmd` package even though most of the code and logic aren't Cobra command files... they are a whole application that the few command files interact with. But they aren't Powergate either, hence a separate repo idea. If it helps with messaging and potential adoption, I'm fine with leaving the code here for now, but we should consider longer term moving it, especially as this code continues to grow.

cmd/fts/cmd/resume.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func init() {
+	resumeCmd.Flags().BoolVarP(&dryRun, "dry-run", "d", false, "Run through steps without pushing data to Powergate API.")
```
asutula, line 12: Would be cool to not have to repeat all this configuration for `run` and `resume`. You could make all these persistent flags at the root level, but then the `version` command would inherit them. Another idea is to make "resume" a flag of `run`, and just detect the value of that flag and fork the code path that gets run at that point.

andrewxhill: yeah, nice. resume as a flag seems like a good call. i could get rid of `run` at that point and make it all work of root

cmd/fts/cmd/staging.go
andrewxhill: first one is correct. second one is 1/2 correct. i was actually wondering if we should make it native in the admin api to set staging limits on ffs instances. there are some edge cases that make this hard (add by cid is one jsign already brought up), but it still seems better than doing it client side. 

long run, the default mode for this would be to do it over buckets. at which point, you can actually let the user set the limit to whatever they want. so didn't get to bogged down here for now.

andrewxhill: i think it _would_ be very nice to test this out with some users on a single POW where we give them their own token though... something to talk about

go.mod
andrewxhill: huh... i was going to use Threads for results collection but found that file on disk worked better at this point. will go back to it when i do a Hub version of this w/ buckets. maybe it got stuck in there though...

cmd/fts/cmd/resume.go
asutula: Yea, there are so few commands, that could work.

go.mod
asutula: Prob just need to run `go mod tidy`

cmd/fts/cmd/staging.go
asutula: Yea, long term doing any type of limiting server side is better, however it is done. From the client, it would just be a simple error you might get back from the call to `Stage`.

cmd/fts/cmd/resume.go
asutula: I think keeping the `run` command makes sense, and add the `--resume` flag to it. There will be a time when you want to add another top level command next to `run`, and that will be weird if run is just a flag on the root command.

PR #719, opened 2020-12-02
minerselector/reputation: improvements

jsign:
In this PR we change the behavior of _Reputation_ miner selector:
- _TrustedMiners_ are always query-asked even if they aren't in the reputation scoring table. If they're online and not in the table, that's because they have zero power. In some situations, this might happen if the miner misses submitting proof. For _TrustedMiners_, we avoid them being filtered by that situation.
- For non-trusted miners selected from the reputation table, we use the cached values for price and max/min piece size for initial filtering. All good candidates are query-asked again to double-check if they still satisfy the constraints in the miner selection instant. This might avoid some situations in which using cached/stalled values might produce incorrect price calculations, or proposing deals to miners that now accept bigger/smaller deals.

A shorter summary:
- _TrustedMiners_ are _more trusted_ now, even if they have zero power.
- All selected miners for deal-making were query-asked again to avoid using cached/stalled values.
- Trusted miners are always evaluated in the order defined in _StorageConfig_. (This didn't change, it's just a clarification).

I also changed the magnitude of price per epoch said in Job logs, from attoFil to FIL.

### DIFF
--- a/ffs/minerselector/reptop/reptop.go
+++ b/ffs/minerselector/reptop/reptop.go
@@ -31,10 +31,11 @@ var _ ffs.MinerSelector = (*RepTop)(nil)
 
 // New returns a new RetTop instance that uses the specified Reputation Module
 // to select miners and the AskIndex for their epoch prices.
-func New(rm *reputation.Module, ai *askRunner.Runner) *RepTop {
+func New(cb lotus.ClientBuilder, rm *reputation.Module, ai *askRunner.Runner) *RepTop {
 	return &RepTop{
 		rm: rm,
 		ai: ai,
+		cb: cb,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


ffs/minerselector/reptop/reptop.go
```diff
@@ -32,30 +45,59 @@ func (rt *RepTop) GetMiners(n int, f ffs.MinerSelectorFilter) ([]ffs.MinerPropos
...
-			continue
-		}
-		res = append(res, ffs.MinerProposal{
-			Addr:       sa.Miner,
-			EpochPrice: sa.Price,
-		})
+		res = append(res, mp)
 		if len(res) == n {
 			break
 		}
```
jsign, line 103: Ugly diff from GH, but the improtant part reg a question you had yesterday @asutula : 
- `rt.rm.Queryminers` returns all the list, it isn't capped.
- The `for` loop only early-break if `len(res)==n`

So that basically means that if potential candidates are discarded by a fresh query-ask, then the loop continues with the next.

asutula (approved):

PR #726, opened 2020-12-10
Data.Info(<cid>) get executing jobs for provided cid

asutula:
This fixes a bug where random jobs from other cids with executing jobs were being listed in the results of `Data.Info(<cid>)`.

### DIFF
--- a/api/server/user/data.go
+++ b/api/server/user/data.go
@@ -200,7 +200,7 @@ func (s *Service) CidInfo(ctx context.Context, req *userPb.CidInfoRequest) (*use
 		}
 		cidInfo.QueuedStorageJobs = rpcQueudJobs
 		executingJobs := i.ExecutingStorageJobs(cid)
-		if len(executingJobs) > 0 {
+		if len(executingJobs) == 1 {
 			// There is exactly one job in the slice because we specified a cid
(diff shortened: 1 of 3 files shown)


api/server/user/data.go
```diff
@@ -199,8 +199,10 @@ func (s *Service) CidInfo(ctx context.Context, req *userPb.CidInfoRequest) (*use
 			rpcQueudJobs[i] = rpcJob
 		}
 		cidInfo.QueuedStorageJobs = rpcQueudJobs
-		executingJobs := i.ExecutingStorageJobs()
+		executingJobs := i.ExecutingStorageJobs(cid)
 		if len(executingJobs) > 0 {
+			// There is exactly one job in the slice because we specified a cid
+			// and there can be only one executing job per cid at a time.
```
jsign, line 205: May worth doing `if len(executingJobs) == 1` and `log.Warn` in an `else`?
Maybe a bit paranoid, so dismiss depending on your feelings.

jsign (approved):
LGTM.

api/server/user/data.go
asutula: Yea I like that

PR #727, opened 2020-12-11
Deal importing & default miner change switch & other improvements

jsign:
This PR provides a way to import deals into existent and non-existent _StorageConfigs_. 
A bit more context and some use-cases.

In Filecoin, on-chain deals are uniquely identified by a DealID which is simply an integer, e.g: 1022. The same data can be stored in multiple deals if for example, some users saves the data in different miners.

Say that some user knows that deals Deal1, Deal2, and deal3 are all active deals for the same data Cid1 and PieceCid1 (we can assume Cid1 and PieceCid1 have a unique one-to-one correspondence; not entirely true but is offtopic for all this).

The user now can call the existing API for pushing a config, with a new extra option to provide an array of DealIDs. Powergate will import all the deal data and store it internally as if those deals were created by Powergate.

This means that after that moment, the user can continue to use Powergate as usual and do whatever it likes with it.

To do retrievals, the user can simply push a config with this option providing DealIDs, and specify in that first storage config Hot-Storage and AllowUnfreeze true. Powergate will first import all these deals data, and then execute a typical Hot-Storage enabling. This will use the imported deals data to do the retrievals. Recall that if the pushed replication factor is bigger than the number of imported deals, Powergate will continue doing what it says: new deals to try ensuring the desired replication factor.

In short, Powergate imports the deals and then simply executes what the storage config says. If you import deals with Hot-Storage disabled, then you're simply adding active deals information to your StorageInfo; which might be an interesting use-case maybe.

The way this works also allows a user to add more existing active deals information for data that already was stored by Powergate. The new option when pushing a config should be understood as augmenting known active deals for that data: so can be used for bootstrapping a new Cid in Powergate (as described above), or simply to keep adding external active deals for data stored in Powergate.

---

Apart from the feature described above, this PR:
- Contains other fixes
- Split some modules which were too big.

I'll add comments to improve signal/noise ratio for review.

---
This PR has [some breaking changes](https://github.com/textileio/powergate/pull/727/files#diff-f729e9861c64d6e4e0d8839bbf620e4946ce665471691e264804776027151ded) regarding a small part of a returned message in a particular API, so not a big thing but strictly breaking in two returned fields. 

I thought some time if this can be avoided, but think the answer is "no". 
What I'm referring to is two fields from `pow data info`:
- ProposalCid
- ActivationEpoch

First of all, these two fields are mostly useless.

`ProposalCid` is the cid of the proposal made to the miner; after the deal is active, can't be used to do anything on an external node. Its kind of an internal Cid that its only useful for the Lotus node that made the deals (and after it's done, it's kind of pointless).

`ActivationEpoch` is the epoch where the sector that contains the deal was activated on-chain, which might be earlier than the real Deal start epoch. Another useless field means nothing to the user. `StartEpoch` field is also returned which is the epoch at which the deal started (which is the useful thing for the user).

OK, so if those things are _useless_, could we avoided "breaking" the API in a small part?
The problem raises when importing deals. When you have a `DealID` you can't reconstruct the corresponding `ProposalCid` since that information isn't on-chain, most probably because you can't do anything with that. We could make all internal `ProposalCid` values be empty for imported deals, but I don't like exceptions in field values if possible. That introduces hidden complexity of having to think always about "this might be empty since might come from an imported deal". (Again, if it's worth for some reason, maybe; but isn't). The same happens with `ActivationEpoch`.

Those files are still tracked in internal deal-records but are not present anymore in `StorageInfo.Cold.Filecoin.Proposals` attributes, which are the ones used by the `pow data info` APIs. That `StorageInfo.Cold.Filecoin.Proposals` array is what gets augmented now with importing deal ids.

What does this mean for existing clients?
Those two fields won't be anymore. I thought about keeping the attribute number at the proto file levels, but that would mean users might silently receive default values. In theory, might be an option... but if someone is relying on them for "important" stuff I prefer the API to break and let them just update the client and make the right decision. If we don't do that and someone is using, e.g, `ProposalCid` as an ID to do something, then returning default values silently might break their things. We can recommend using the new `DealID` for that.

For Hub's particular case, I think this isn't a big deal since @asutula  started using deal-records data to store deal information in Hub. And deal-record data is independent of this changed API output. Also, that data already had DealID, which now is the important thing to provide to users to let them recover their deals. Saying it differently, Hub wasn't using "StorageInfo" but mostly deal-records for the corresponding Job, which had more info.

To be honest, I'd like to remove more things and maybe just leave the newly added `DealID`. The rest of the information can be reconstructed only with `DealID`. Why did we get to the point of adding these extra fields? Mostly because we started very early with Powergate and the meaning of these fields wasn't 100% clear.

OK, so does this require any "migrations"? I'd say no. Deleting these fields would only mean that old that might still have this info, but since isn't anymore in structs, it will be ignored while unmarshaling things from go-datastore; so not a big deal from my perspective.


ffs/scheduler/scheduler_import.go
```diff
@@ -0,0 +1,124 @@
+package scheduler
```
jsign, line 1: Here the logic of deal importing.
This code has some comments, so I'll add extra ones in review if make sense.
Feel free to ask.

ffs/scheduler/scheduler_import.go
```diff
@@ -0,0 +1,124 @@
...
+			Created: time.Now(),
+			Cold: ffs.ColdInfo{
+				Filecoin: ffs.FilInfo{
+					DataCid: payloadCid,
+				},
+			},
+		}
+	} else if err != nil {
+		return fmt.Errorf("getting current storageinfo: %s", err)
+	}
```
jsign, line 32: First, we get the current _StorageInfo_.

If there isn't one, the user is pushing a _StorageConfig_ for the first time, since we don't have any current _StorageInfo_. In this case, we create an empty one with minimal necessary information.

If there is one, the user is augmenting an existing storage with external deals.

ffs/scheduler/scheduler_import.go
```diff
@@ -0,0 +1,124 @@
...
+	var newDealIDs []uint64
+	existingDealID := map[uint64]struct{}{}
+	for _, p := range fi.Proposals {
+		existingDealID[p.DealID] = struct{}{}
+	}
+	for _, did := range deals {
+		if _, ok := existingDealID[did]; !ok {
+			newDealIDs = append(newDealIDs, did)
+		}
+	}
```
jsign, line 57: There's the possibility that the user is providing DealIDs that we already know. Just be sure to take them out for the importing logic.

ffs/scheduler/scheduler_import.go
```diff
@@ -0,0 +1,124 @@
...
+
+		fi.Proposals = append(fi.Proposals, fs)
+
+		// If our StorageInfo doesn't know about size, which is the case if we're boostrapping only
+		// knowing DealIDs, then populate that data.
+		if fi.Size == 0 {
+			fi.Size = size
+		}
+	}
+
```
jsign, line 103: The _PieceCid_ part might be a bit confusing.
Basically we have two scenarios:
- The user is bootstraping a StorageInfo for the first time with external deals.
- The user is augmenting some data already stored with Powergate with some external deals to add as information.

In any case, we want to be sure that the DealIDs we're receiving are for the same data. 
In case 1, the only thing we can do is at least validate that PieceCID is the same for all received DealIDs.
In case 2, we already know the PieceCID of the data. So we do a more strict check that all received DealIDs are for that PieceCID that we already know (and not just they're the same between each other).

All these are validations. In theory the user should not be providing wrong DealIDs... but better check as many things as possible.

ffs/scheduler/scheduler_import.go
```diff
@@ -0,0 +1,124 @@
...
+	return nil
+}
+
+// genFilStorage given a DealID it generates FilStorage and returns the piece size.
+func (s *Scheduler) genFilStorage(dealID uint64) (ffs.FilStorage, uint64, error) {
+	ctx, cls := context.WithTimeout(context.Background(), time.Second*10)
+	defer cls()
+	di, err := s.cs.GetDealInfo(ctx, dealID)
```
jsign, line 111: We get deal information from on-chain Lotus, then generate the corresponding `ffs.FilStorage` just as if Powergate stored this data.

migration/migration2.go
```diff
@@ -0,0 +1,189 @@
+package migration
```
jsign, line 1: This migration is for the changes done in _ffs.FilStorage_:
- Removed _ProposalCid_.
- Removed _ActivationEpoch_.
- Added _DealID_.

If was for the first two bullet points, we wouldn't require a migration since we can leave that data in the datastore and the unmarshaling would just ignore the data. The cause of needed a migration is adding _DealID_. If we don't do something about that, all _DealID_ of existing `ffs.FilStorage` will be 0, which is wrong.

To do this migration we basically leverage the information we have in `/deals/record/<storage-pending/storage-final>/<iid>/<proposalCid>`, so we can map _ProposalCid_ to _DealID_ to fill that attribute.

migration/migration2.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/textileio/powergate/deals"
+	"github.com/textileio/powergate/ffs"
+)
+
+// V2StorageInfoDealIDs contains the logic to upgrade a datastore from
+// version 1 to version 2. Transactionality is disabled since is a big migration.
+var V2StorageInfoDealIDs = Migration{
+	UseTxn: false,
+	Run: func(ds datastoreReaderWriter) error {
+		propDealIDMap, err := v2GenProposalCidToDealID(ds)
```
jsign, line 20: We generate a `map[ProposalCid]DealID`.

migration/migration2.go
```diff
@@ -0,0 +1,189 @@
...
+		if err != nil {
+			return fmt.Errorf("getting cid owners: %s", err)
+		}
+		log.Infof("Starting DealID filling...")
+		if err := v2MigrationDealIDFilling(ds, propDealIDMap); err != nil {
```
jsign, line 25: Do the migration using that generated mapping.

migration/migration2.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		log.Infof("DealID filling migration finished")
+
+		return nil
+	},
+}
+
+func v2GenProposalCidToDealID(ds datastoreReaderWriter) (map[cid.Cid]uint64, error) {
+	ret := map[cid.Cid]uint64{}
+	qs := []query.Query{{Prefix: "/deals/storage-pending"}, {Prefix: "/deals/storage-final"}}
```
jsign, line 36: Iterate these two keyspaces, and generate the map.

migration/migration2.go
```diff
@@ -0,0 +1,189 @@
...
+				errors = append(errors, fmt.Sprintf("unmarshaling storage info from datastore: %s", err))
+				lock.Unlock()
+				return
+			}
+			if err := json.Unmarshal(r.Value, &newCi); err != nil {
+				lock.Lock()
+				errors = append(errors, fmt.Sprintf("unmarshaling storage info from datastore: %s", err))
+				lock.Unlock()
+				return
+			}
```
jsign, line 142: A very interesting situation here: why unmarshaling two times?

The problem is the following. In this version of the code `ffs.StorageInfo` already has wiped out `ProposalCid`. So the unmarshaling using that type won't have the `ProposalCid` we need for taking a look at the `ProposalCid->DealID` map!

So I created `v1PartialStorageInfo` which is basically a partial struct of the old (current) JSON with the _ProposalCid_ field. Then we also unmashal the same JSON but with the new `ffs.StorageInfo` to fill the _DealID_ field.

migration/migration2.go
```diff
@@ -0,0 +1,189 @@
...
+			if err != nil {
+				lock.Lock()
+				errors = append(errors, fmt.Sprintf("marshaling storage info for datastore: %s", err))
+				lock.Unlock()
+			}
+			if err := ds.Put(datastore.NewKey(r.Key), buf); err != nil {
+				lock.Lock()
+				errors = append(errors, fmt.Sprintf("put storage info in datastore: %s", err))
+				lock.Unlock()
+			}
```
jsign, line 171: Notice that we iterate over `oldCi.Cold.Filecoin.Proposals`, use the _ProposalCid_ from there, and then fill the _DealID_ of **that same entry** but unmarshaled in the current `ffs.StorageInfo`.  Then we save the new filled `ffs.StorageInfo` that now has the _DealID_ information (and indirectly, deleted the old _ProposalCid_ and _ActivationEpoch_ fields since they are not part of the new structure type)

Sidenote: If we happen to not know the mapping, we remove it. We can't leave things with _DealID_ zero.

migration/migration2.go
```diff
@@ -0,0 +1,189 @@
...
+		}()
+	}
+
+	for i := 0; i < cap(lim); i++ {
+		lim <- struct{}{}
+	}
+
+	log.Infof("Couldn't be mapped %d, Could be mapped %d", cantBeMapped, canBeMapped)
```
jsign, line 179: I did test with real data and most of the things can be mapped, so I haven't discovered any big problems.

migration/migration_test.go
```diff
@@ -141,11 +141,11 @@ func TestFailingMigration(t *testing.T) {
 	require.Equal(t, 0, v)
 }
 
-func TestRealDataBadger(t *testing.T) {
+func TestRealDataBadgerV0(t *testing.T) {
 	logger.SetDebugLogging()
 	_ = logger.SetLogLevel("badger", "error")
 	tmpDir := t.TempDir()
-	err := copyDir("testdata/badgerdump", tmpDir+"/badgerdump")
+	err := copyDir("testdata/badgerdumpv0", tmpDir+"/badgerdump")
```
jsign, line 148: The `badgerdump` folder had some exported database. Now I needed a new one that was already migrated to version 1. So I renamed the old one with version number.

migration/testdata/v1_StorageInfo.post
```diff
@@ -1,4 +1,4 @@
...
+/ffs/scheduler/cistore_v2/ID3/QmZTMaDfCMWqhUXYDnKup8ctCTyPxnriYW7G4JR8KXoX5M,{"APIID":"ID3","JobID":"35a6f8fc-cad5-41f4-a410-e5c3d882f0e0","Cid":{"/":"QmZTMaDfCMWqhUXYDnKup8ctCTyPxnriYW7G4JR8KXoX5M"},"Created":"2020-10-20T12:29:12.996702455Z","Hot":{"Enabled":false,"Size":0,"Ipfs":{"Created":"0001-01-01T00:00:00Z"}},"Cold":{"Enabled":true,"Filecoin":{"DataCid":{"/":"QmZTMaDfCMWqhUXYDnKup8ctCTyPxnriYW7G4JR8KXoX5M"},"Size":67108864,"Proposals":[{"DealID":0,"PieceCid":{"/":"baga6ea4seaqhvlh3ktd4rxnbozupun776lu4wh6bprrcr6t63ec2fvkkwatn4oi"},"Renewed":false,"Duration":521469,"StartEpoch":168927,"Miner":"f010507","EpochPrice":6250000000},{"DealID":0,"PieceCid":{"/":"baga6ea4seaqhvlh3ktd4rxnbozupun776lu4wh6bprrcr6t63ec2fvkkwatn4oi"},"Renewed":false,"Duration":519893,"StartEpoch":168927,"Miner":"f022142","EpochPrice":12500000},{"DealID":0,"PieceCid":{"/":"baga6ea4seaqhvlh3ktd4rxnbozupun776lu4wh6bprrcr6t63ec2fvkkwatn4oi"},"Renewed":false,"Duration":520113,"StartEpoch":149621,"Miner":"f03491","EpochPrice":6250000000}]}}}
```
jsign, line 4: Since _StorageInfo_ changed, a previous test of v1 migration had to change a bit the data.
I'll recommend that if anyone is still behind the v1 migration, jump to that one first, and then to this one.

Making things jump for v0 to v2 for the amount of things that changed would take much more work. Mostly, replication old struct versions for migrations. (and we have some long nested ones).

migration/migration2.go
asutula: I'll be interested to see how this works in production. For a period of time, when deals started taking over 24 hrs to become active, the code that was updating the deal records would time out and give up on creating a deal record. That has since been fixed, but there may be some deals that don't have corresponding deal records for that reason.

migration/migration2.go
asutula: Cool, glad we just discard things we don't have a deal id for. Like I said, I want to know the count, logged below :)

migration/migration2.go
asutula: Oh! great.

asutula (approved):
Looks really good. I love when you add a new feature _and_ the code also gets more simple 🥇

cmd/pow/cmd/config_apply.go
```diff
@@ -19,6 +21,7 @@ func init() {
 	configApplyCmd.Flags().StringP("conf", "c", "", "Optional path to a file containing storage config json, falls back to stdin, uses the user default by default")
 	configApplyCmd.Flags().BoolP("override", "o", false, "If set, override any pre-existing storage configuration for the cid")
 	configApplyCmd.Flags().BoolP("watch", "w", false, "Watch the progress of the resulting job")
+	configApplyCmd.Flags().StringP("import-deals", "i", "", "Comma-separated list of deal ids to import")
```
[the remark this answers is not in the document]
jsign: Nice, I'll change to that.

migration/migration2.go
jsign: Yeah, I recalled that 24hs thing you mention above. Not much to do about that, but yep, as far as I tested on real that wasn't a big %.

PR #728, opened 2020-12-16
Proposed CLI package structure

asutula:
It's getting out of hand to have all CLI commands in the same package. The namespace is just too crowded and code is feeling monolithic in there. The prospect of adding some interactive text based UIs (TUIs) to some commands is just going to make this worse.

Feels nice to button up commands into their own packages. In this branch, I did that only with the `data` command and its sub commands to see how it goes. This required moving some shared code into a `common` package to avoid circular imports, so that refactor touches every `pow` CLI command file. Just know it's an organizational refactor, no functionality or logic changes.

In the next PR (targeting this branch), I'll show how I added a TUI to one of the data sub commands. So, this PR is all about CLI code organization, next one is about TUI implementation.

In this branch, the `pow` CLI directory structure now looks like:
```
➜ tree cmd/pow    
cmd/pow
├── cmd
│   ├── admin.go
│   ├── admin_data.go
│   ├── admin_jobs.go
│   ├── admin_users.go
│   ├── admin_wallet.go
│   ├── config.go
│   ├── config_apply.go
│   ├── config_default.go
│   ├── config_remove.go
│   ├── config_set_default.go
│   ├── data
│   │   ├── data.go
│   │   ├── get
│   │   │   └── get.go
│   │   ├── info
│   │   │   └── info.go
│   │   ├── log
│   │   │   └── log.go
│   │   ├── replace
│   │   │   └── replace.go
│   │   ├── stage
│   │   │   └── stage.go
│   │   └── summary
│   │       ├── summary.go
│   ├── deals.go
│   ├── deals_retrievals.go
│   ├── deals_storage.go
│   ├── docs.go
│   ├── id.go
│   ├── root.go
│   ├── storage_jobs.go
│   ├── storage_jobs_cancel.go
│   ├── storage_jobs_executing.go
│   ├── storage_jobs_get.go
│   ├── storage_jobs_latest_final.go
│   ├── storage_jobs_latest_successful.go
│   ├── storage_jobs_queued.go
│   ├── storage_jobs_storage_config.go
│   ├── storage_jobs_summary.go
│   ├── storage_jobs_watch.go
│   ├── version.go
│   ├── wallet.go
│   ├── wallet_addrs.go
│   ├── wallet_balance.go
│   ├── wallet_new_addr.go
│   ├── wallet_send.go
│   ├── wallet_sign.go
│   └── wallet_verify.go
├── common
│   └── common.go
└── main.go
```

[...]

PR #730, opened 2020-12-16
New cid summary api and simplified cid info

asutula:
The existing `Data.CidInfo` API is just too much data, different types of data, in a list, and it gets too big to consume efficiently. This creates a "summary" api that will return the most basic information about all the cids you have stored in powergate, then you can call `Data.CidInfo` with any one of those cids to still get detailed and varied information about a single cid.

Ended up merging a couple other PRs into this one. For completeness, the are:
- #728 which creates a package structure for cli code
- #748 which adds a new API for querying storage jobs


api/server/user/data.go
```diff
@@ -156,8 +156,8 @@ func (s *Service) WatchLogs(req *userPb.WatchLogsRequest, srv userPb.UserService
 	return nil
 }
 
-// CidInfo returns information about cids managed by the FFS instance.
-func (s *Service) CidInfo(ctx context.Context, req *userPb.CidInfoRequest) (*userPb.CidInfoResponse, error) {
+// CidSummary gives a summary of the storage and jobs state of the specified cid.
+func (s *Service) CidSummary(ctx context.Context, req *userPb.CidSummaryRequest) (*userPb.CidSummaryResponse, error) {
```
asutula, line 160: This method is surprisingly slow, and I need to figure out why. When I call it against our powergate instance with about 700 cids stored, it takes 10s of seconds to return the data. The amount of data is small, so it must be some processing in here that is taking a while.

proto/powergate/user/v1/user.proto
```diff
@@ -99,12 +99,20 @@ message WatchLogsResponse {
...
 }
 
-message CidInfoRequest {
+message CidSummaryRequest {
   repeated string cids = 1;
 }
 
+message CidSummaryResponse {
+  repeated CidSummary cid_summary = 1;
+}
```
asutula, line 108: The new request/response type.

proto/powergate/user/v1/user.proto
```diff
@@ -99,12 +99,20 @@ message WatchLogsResponse {
...
+
+message CidInfoRequest {
+  string cid = 1;
+}
+
 message CidInfoResponse {
-  repeated CidInfo cid_infos = 1;
+  CidInfo cid_info = 1;
```
asutula, line 115: `CidInfo` request/response is now just for a single cid.

proto/powergate/user/v1/user.proto
```diff
@@ -318,6 +327,13 @@ message AddrInfo {
   string balance = 4;
 }
 
+message CidSummary {
+  string cid = 1;
+  bool stored = 2;
+  repeated string queued_jobs = 3;
+  string executing_job = 4;
+}
```
asutula, line 335: Summary returns some simple inflation about a cid.

api/server/user/data.go
jsign: I'm not sure when you test that, but today there were some heavy processes in the Atlas db.
Maybe you got unlucky? Mentioning it since was a special day today.
Dismiss if isn't the case... might be a real prob.

api/server/user/data.go
```diff
@@ -176,63 +176,172 @@ func (s *Service) CidInfo(ctx context.Context, req *userPb.CidInfoRequest) (*use
...
+		return time
 	}
-	sort.Slice(res, func(a, b int) bool {
-		return res[a].Cid < res[b].Cid
+
+	sort.Slice(sources, func(a, b int) bool {
+		jobTimeA := extractJobTime(sources[a])
+		jobTimeB := extractJobTime(sources[b])
+		currentStorageTimeA := extractCurrentStorageTime(sources[a])
+		currentStorageTimeB := extractCurrentStorageTime(sources[b])
```
jsign, line 238: This could be moved below the two ifs to avoid calculating if any of the ifs match.

api/server/user/data.go
```diff
@@ -176,63 +176,172 @@ func (s *Service) CidInfo(ctx context.Context, req *userPb.CidInfoRequest) (*use
...
+	}
+
+	return &userPb.CidSummaryResponse{CidSummary: summaries}, nil
+}
+
+// CidInfo returns information about cids managed by the FFS instance.
+func (s *Service) CidInfo(ctx context.Context, req *userPb.CidInfoRequest) (*userPb.CidInfoResponse, error) {
+	i, err := s.getInstanceByToken(ctx)
+	if err != nil {
+		return nil, err
```
jsign, line 273: Nit: `status.Errorf(Unauthorized, "...")`?

api/server/user/data.go
```diff
@@ -176,63 +176,172 @@ func (s *Service) CidInfo(ctx context.Context, req *userPb.CidInfoRequest) (*use
...
+	if len(executingJobs) == 1 {
+		// There is exactly one job in the slice because we specified a cid
+		// and there can be only one executing job per cid at a time.
+		rpcJob, err := toRPCJob(executingJobs[0])
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "converting job to rpc job: %v", err)
+		}
+		cidInfo.ExecutingStorageJob = rpcJob
+	} else if len(executingJobs) > 1 {
+		log.Warnf("received %d executing jobs when there should be 1", len(executingJobs))
```
jsign, line 326: Is safe not to `break` or return err in this case?

api/server/user/util.go
```diff
@@ -279,10 +279,18 @@ func toRPCJob(job ffs.StorageJob) (*userPb.StorageJob, error) {
 	}, nil
 }
 
+func fromProtoCid(c string) (cid.Cid, error) {
+	res, err := util.CidFromString(c)
+	if err != nil {
+		return cid.Undef, err
+	}
+	return res, nil
+}
```
jsign, line 288: Isn't this the same of directly using `util.CidFromString`?

cmd/pow/cmd/data_summary.go
```diff
@@ -0,0 +1,43 @@
...
+
+func init() {
+	dataCmd.AddCommand(dataSummaryCmd)
+}
+
+var dataSummaryCmd = &cobra.Command{
+	Use:   "summary [optional cid1,cid2,...]",
+	Short: "Get a summary about the current storage and jobs state of cids",
+	Long:  `Get a summary about the current storage and jobs state of cids`,
+	Args:  cobra.MaximumNArgs(1),
```
jsign, line 21: Couldn't this CLI receive cid1, cid2, ..., cidN. 
`MaximumNArgs(1)` is the right setup?

jsign (approved):
LGTM

api/server/user/data.go
asutula: Tested before the downtime. Will have to dig in and see what's up.

api/server/user/data.go
asutula: It is safer to return an error. Done.

api/server/user/util.go
asutula: Yep! Deleting this.

cmd/pow/cmd/data_summary.go
asutula: Yea technically it's one string arg that contains commas, and we split the arg on the commas.

asutula:
Ok all rebased against `develop` here and changed the target of the PR to `develop`.

PR #732, opened 2020-12-17
cmd/powcfg: tool for custom StorageConfig transformations

jsign:
This is a new tool `powcfg`, under `cmd/powcfg`.
Calling it a CLI would be a stretch, it's more like a dev tool.

This tool lets you run concrete code to do whatever transformation you want in all _StorageConfigs_.
I'll explain its features in PR comments.

The original motivation for this was needing to migrate all existing _StorageConfigs_ to have a minimum value of _Hot.Ipfs.AddTimeout_. In the past, we had some really low default (10s), that's too low:
- If you're fetching the data from the IPFS network, 10s is almost negligible for any data fetching.
- If you're fetching from a directly connected peer, maybe is a good amount. But for big data (GiBs) might be too tight.
- If you're trying to fetch GiBs from the IPFS network... forget about it.

This PR comes with a unique transformation that does that exactly: bumps the value to a new reasonable 8min, but can be easily changed. We can keep adding more helpful transformations and have some battery of helpers that people might like to use.


cmd/powcfg/main.go
```diff
@@ -0,0 +1,96 @@
...
+	log.Infof("starting powcfg:\n%s", buildinfo.Summary())
+
+	if err := wireFlagsAndEnvs(); err != nil {
+		log.Fatalf("wiring flags/envs: %s", err)
+	}
+
+	mongoURI := config.GetString("mongouri")
+	mongoDB := config.GetString("mongodb")
+	mongoCollection := config.GetString("mongocollection")
+	badgerrepo := config.GetString("badgerrepo")
```
jsign, line 33: The user should configure where the go-datastore is (support both types).

cmd/powcfg/main.go
```diff
@@ -0,0 +1,96 @@
...
+	dryrun := config.GetBool("dryrun")
```
jsign, line 34: You can supply `--dryrun`, and that would mean that transformation would never write changes.
This is useful to know how many storage configs are going to be affected since this number gets printed when you run the program.

cmd/powcfg/ratelim/ratelim.go
```diff
@@ -0,0 +1,47 @@
+package ratelim
```
jsign, line 1: I decided to do a generic rate limiter. 
This is an idea that we applied in many places in Powergate, but never re-used as an internal library.
Eventually, we move this lib to main Powergate and use it.

cmd/powcfg/transform.go
```diff
@@ -0,0 +1,96 @@
+package main
```
jsign, line 1: TL;DR:
- Iterate all storage configs
- Call the provided transformation
- If the transformation changed the storage config, save the new value
- No writes are done if we're `--dryrun`.

cmd/powcfg/transform.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("running query: %s", err)
+	}
+	defer func() {
+		if err := res.Close(); err != nil {
+			log.Warnf("closing query result: %s", err)
+		}
+	}()
+
+	rl, err := ratelim.New(1000)
```
jsign, line 31: Example of `ratelim` usage: provide max limit.

cmd/powcfg/transform.go
```diff
@@ -0,0 +1,96 @@
...
+		parts := key.Namespaces()
+		if len(parts) < 6 {
+			continue
+		}
+		if parts[4] != "istore" || parts[5] != "cidstorageconfig" {
+			continue
+		}
+
+		r := r
+		rl.Exec(func() error {
```
jsign, line 54: Exec whatever you want. If no more slots are available, it will block until one is free.

cmd/powcfg/transform.go
```diff
@@ -0,0 +1,96 @@
...
+			}
+			if err := ds.Put(datastore.NewKey(r.Key), buf); err != nil {
+				return fmt.Errorf("put in datastore: %s", err)
+			}
+
+			return nil
+		})
+	}
+
+	errors := rl.Wait()
```
jsign, line 87: Wait for all things to finish. Done.

cmd/powcfg/transformations.go
```diff
@@ -0,0 +1,19 @@
+package main
```
jsign, line 1: Most of the code in this PR is structure.

Here is a transformation You receive a `*StorageConfig`, modify it if applies to whatever you want to do, and return if this entry should be re-saved (since was modified). The query, saving, etc is done by boilerplate code.

From now forward, whenever we want to make some massive change, we should only create a function here and point `applyTransformation` in `main.go` to use it.

go.mod
```diff
@@ -46,6 +46,7 @@ require (
 	github.com/olekukonko/tablewriter v0.0.4
 	github.com/ory/dockertest/v3 v3.6.2
 	github.com/oschwald/geoip2-golang v1.4.0
+	github.com/otiai10/copy v1.3.0
```
jsign, line 49: Used for copying folders (for tests).
In some other migration tests I have some custom code for copying folders that we can change in the future to simply use this lib too.

cmd/powcfg/ratelim/ratelim.go
asutula: Sweet

asutula (approved):

PR #734, opened 2020-12-18
Make embedded logging setup optional to support ipfs/go-log environment variables

### LINKED ISSUE #733: Add JSON logging support
It would be really useful to have JSON logging support in addition to pretty print so, it's better parsed in any logging platform such as ELK, DataDog, etc. 

It can be added in the config and/or added as a flag

kalambet:
In this PR I tried to fix #733 by making internal logging setup optional via new command line flag (the name is not ideal, but that is the best I was able to come up with).

`ipfs/go-log` that is used for the logging purposes executes setup through environment variables on package initialization: https://github.com/ipfs/go-log/blob/4f3460899491d70a650f7fd6b10217cae6ba3c41/setup.go#L16

That is why I moved original code under `if` statement, so the explicit setup of the logging subsystem here: https://github.com/textileio/powergate/blob/2a8087695f184b06632df58b34db7c760b9fde01/cmd/powd/main.go#L228 
will not override `go-log` settings gathered via such variables.

W/o `ipfslog` flag defined the behavior remains the same.

### DIFF
--- a/cmd/powd/main.go
+++ b/cmd/powd/main.go
@@ -273,7 +273,7 @@ func setupLogging(repoPath string) error {
 	// If at least one of them defined - do not override
 	// ipfs/go-log internal logging setup
 	for _, e := range os.Environ() {
-		if strings.HasPrefix(e, "GOLOG_LOG_") {
+		if strings.HasPrefix(e, "GOLOG_") {
 			ipfslog = true
 			break
 		}
--- a/cmd/powd/main.go
+++ b/cmd/powd/main.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/powd/main.go
```diff
@@ -276,12 +267,29 @@ func setupLogging(repoPath string) error {
...
 
-	// powd registered loggers get info level by default.
-	for _, l := range loggers {
-		if err := logging.SetLogLevel(l, "info"); err != nil {
-			return fmt.Errorf("setting up logger %s: %s", l, err)
+	ipfslog := config.GetBool("ipfslog")
+	// If ipfslog is defined, the output format, destination and level
+	// can be managed via native to ipfs/go-log library
+	// environment variables (e.g. GOLOG_LOG_*)
+	if !ipfslog {
```
jsign, line 274: @kalambet , what do you think of removing the needed new `ipfslog` flag, and deriving the behaviour by taking a look at the environment variables. If any `GOLOG_LOG_` is present, maybe we can assume the `ipfslog=true` behaviour.

Doing something like:
```
var ipfslog bool
for _, e := range os.Environ() {
	if strings.HasPrefix(e, "GOLOG_LOG_") {
		ipfslog = true
		break
	}
}
if !ipfslog { 
...
```
Maybe this would avoid some config burden. If the user is providing `GOLOG_LOG_*`, then most probably always would want the behaviour automatically without extra Powergate flags.

kalambet: nice! I'll update the PR. thank you!

kalambet:
Fixed, plus updated environment variable prefix because of
https://github.com/ipfs/go-log/blob/4f3460899491d70a650f7fd6b10217cae6ba3c41/setup.go#L32

jsign (approved):
LGTM

PR #736, opened 2020-12-22
Allow applying a storage-config without executing a job

jsign:
Before this change, applying a new _StorageConfig_ would create a Job that will ensure that the desired state holds.

In some situations, clients are interested in pushing a new configuration but not necessarily run it right away.
Examples:
- I want to increase the deal max price, but not necessarily create new deals if the current replication factor doesn't hold.
- I want to increase the replication factor, but not make deals if the current one doesn't match with the new desired one.
- I want to change the trusted miners.
- We want to import new DealIDs without re-evaluating the desired state (motivation for this PR).

In any of those cases, the user might be interested in changing the desired state and simply wait for a later stage to execute, or simply let the new configuration be executed by the repair/renew cron in the future (and not now).

To use this feature, the user would simply use a new option `WithNoExec(...)` which basically saves the new storage-config, but avoids creating the job (and thus forcing the execution).


asutula:
Just a first reaction to the PR description:

What exactly do you mean by:
> wait for a later stage to execute
?

asutula (approved):
Looks good!

jsign:
> Just a first reaction to the PR description:
> 
> What exactly do you mean by:
> 
> > wait for a later stage to execute
> 
> ?

Doing a series of config changes, and in the last one execute; so manual one.

PR #748, opened 2021-01-03
List storage jobs query API

asutula:
The `StorageJob`s API wasn't feeling right to me, and the main problem is that there was no way to query for a list of historical `StorageJobs`. This PR implements that new API and also takes a stab at the more general problem of data paging in APIs. This new `ListStorageJobs` API attempts to support query options for filter by API id, filter by cid, filter by job state, limit, ascending/descending, all while also supporting data paging. It's a bit of a bespoke solution for the particular use case of `StorageJobs`, but I feel like we can use similar ideas in other places in the API where data paging is required.

I want to get the bulk of the work out for preliminary PR feedback now, and will continue with the work still remaining: 

- [x] Incorporate this new API into the client and CLI
- [x] Write a needed data migration in the `sjstore`
- [x] Deleting redundant/old/useless APIs that this new one replaces
- [x] Finish writing through tests for `sjstore` queries.


ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -360,111 +369,169 @@ func (s *Store) GetStartedDeals(iid ffs.APIID, c cid.Cid) ([]cid.Cid, error) {
...
+	Executing
+	// Final lists final StorageJobs.
+	Final
+)
+
+// ListConfig controls the behavior for listing StorageJobs.
+type ListConfig struct {
+	// APIIDFilter filters StorageJobs list to the specified APIID. Defaults to no filter.
+	APIIDFilter ffs.APIID
+	// CidFilter filters StorageJobs list to the specified cid. Defaults to no filter.
```
jsign, line 390: Nit: list.

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -360,111 +369,169 @@ func (s *Store) GetStartedDeals(iid ffs.APIID, c cid.Cid) ([]cid.Cid, error) {
...
+				log.Errorf("extracting time from key b: %v", err)
+				return 0
+			}
+			if lTime > rTime {
+				return -1
+			} else if rTime > lTime {
+				return 1
+			} else {
+				return 0
 			}
```
jsign, line 438: At first, I thought that `if ascending { ` case should be `if !ascending {`, but I see here the conditions are reverted to what I consider the most common understanding of comparisons.

I think in general `left < right` is -1, equal 0, and `left > right` is 1.

Not a point to change code, just a funny comment.

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -360,111 +369,169 @@ func (s *Store) GetStartedDeals(iid ffs.APIID, c cid.Cid) ([]cid.Cid, error) {
...
-			for cid := range m[iid] {
-				cidsList = append(cidsList, cid)
+		// Filter out based on queued/executing etc
+		switch config.Select {
+		case All:
+		case Queued:
+			if _, queued := s.queuedIDs[jobID]; queued {
+				break
+			}
+			continue
```
jsign, line 503: Maybe simplify to?:
```
if _, queued := s.queuedIDs[jobID]; !queued {
   continue
}
```

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -360,111 +369,169 @@ func (s *Store) GetStartedDeals(iid ffs.APIID, c cid.Cid) ([]cid.Cid, error) {
...
+		job, err := s.get(jobID)
+		if err != nil {
+			return nil, false, "", fmt.Errorf("getting job: %v", err)
+		}
+		jobs = append(jobs, job)
+		nextPageToken = jobIDString
+		if len(jobs) == int(config.Limit) {
+			done = true
 		}
+		continue
```
jsign, line 527: I think this isn't needed? Maybe leftover from other thing?

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -360,111 +369,169 @@ func (s *Store) GetStartedDeals(iid ffs.APIID, c cid.Cid) ([]cid.Cid, error) {
...
+	} else if config.APIIDFilter == ffs.EmptyInstanceID && config.CidFilter.Defined() {
+		// use cid
+		prefix = prefixCid(config.CidFilter)
+	}
 
-	return mappedJobs(s.lastFinalJobs, iid, cids...)
-}
+	q := query.Query{
+		Prefix: prefix,
+		Orders: []query.Order{query.OrderByFunction(byTime(config.Ascending))},
```
jsign, line 457: Here's an important point that I mentioned it some time ago when discussing about how Query implementations work.

Whenever you set some `Orders`, what implementations will do is fetch all results from the datastore in memory, and then apply the sorting. So this means that all queries will load all things in memory, sort them, and then make them available in the for loop.

For some cases in the above if, the `Orders` could be avoided since by default keys are iterated in lexicographic order since you also added createdAt at the end of the key. At least, for Ascending ordering seems that `Orders` can be avoided, but descending ones we are out of luck (really unfortunate).

Does this make sense to you?

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -360,111 +369,169 @@ func (s *Store) GetStartedDeals(iid ffs.APIID, c cid.Cid) ([]cid.Cid, error) {
...
-	sort.Slice(res, func(a, b int) bool {
-		return res[a].CreatedAt > res[b].CreatedAt
-	})
+	if !more {
+		nextPageToken = ""
+	}
 
-	return res
+	return jobs, more, nextPageToken, nil
 }
```
jsign, line 535: Apart from the `Orders` point, I think this implementation makes sense.
This got quite complex considering how unfortunate making queries with go-datastore is.

jsign (commented):
Looks good, I left some notes.

api/client/admin/storagejobs.go
```diff
@@ -4,96 +4,98 @@ import (
...
-func WithUserID(userID string) StorageJobsOption {
-	return func(conf *storageJobsConfig) {
-		conf.UserID = userID
-	}
+// ListConfig controls the behavior for listing StorageJobs.
+type ListConfig struct {
+	// UserIDFilter filters StorageJobs list to the specified user ID. Defaults to no filter.
+	UserIDFilter string
+	// CidFilter filters StorageJobs list to the specified cid. Defaults to no filter.
+	CidFilter string
```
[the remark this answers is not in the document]
asutula: It's currently just one single cid. We previously had the ability to filter by many cids, but it seemed like something most people would probably never use, so I made it a single cid for now since that fits well with go-datastore prefix querying using those new indexes. Maybe this is too simple? Open to discussion here.

api/client/admin/storagejobs.go
```diff
@@ -4,96 +4,98 @@ import (
...
+	// Limit limits the number of StorageJobs returned. Defaults to no limit.
+	Limit uint64
```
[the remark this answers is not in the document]
asutula: Yes, I can set some default limit for remote clients. I think on the server side, the default should still be no limit so that in-process powergate calls can just get whatever data they want fast without having to page.

api/client/storagejobs.go
```diff
@@ -14,6 +14,48 @@ type StorageJobs struct {
...
+// SummaryOption configures a storageJobsConfig.
+type SummaryOption = func(*summaryConfig)
+
+// WithCid filters the results to the specified data cid.
+func WithCid(cid string) SummaryOption {
+	return func(conf *summaryConfig) {
+		conf.cid = cid
+	}
+}
+
```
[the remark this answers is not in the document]
asutula: They are already different... the admin client supports filtering by user id and cid, while the user client only supports cid. I agree I don't like the duplicated parts. Will think about the best way to share code here.

api/client/admin/storagejobs.go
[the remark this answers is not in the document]
asutula: Ooops, yea the comment is wrong, will correct.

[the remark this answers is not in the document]
asutula: No never mind, it is correct.

api/server/admin/jobs.go
```diff
@@ -8,126 +8,96 @@ import (
...
+		Ascending:     req.Ascending,
+		NextPageToken: req.NextPageToken,
+		Select:        selector,
 	}
-	jobs := a.s.ExecutingStorageJobs(ffs.APIID(req.UserId), cids...)
-	protoJobs, err := user.ToProtoStorageJobs(jobs)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "converting jobs to protos: %v", err)
+	if req.CidFilter != "" {
+		c, err := cid.Decode(req.CidFilter)
```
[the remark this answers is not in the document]
asutula: Think you just misread the comment.
```
CidFilter filters StorageJobs list to the specified cid.
```

api/server/admin/jobs.go
```diff
@@ -8,126 +8,96 @@ import (
...
-			LatestSuccessful: int32(len(latestSuccessfulJobs)),
-			Queued:           int32(len(queuedJobs)),
-		},
-		ExecutingStorageJobs:        protoExecutingJobs,
-		LatestFinalStorageJobs:      protoLatestFinalJobs,
-		LatestSuccessfulStorageJobs: protoLatestSuccessfulJobs,
-		QueuedStorageJobs:           protoQueuedJobs,
+		CountQueuedStorageJobs:    int32(len(queuedJobIDs)),
+		CountExecutingStorageJobs: int32(len(executingJobIDs)),
+		CountFinalStorageJobs:     int32(len(finalJobIDs)),
```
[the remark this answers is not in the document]
asutula: yea, I agree. will do.

api/server/user/service.go
```diff
@@ -63,11 +61,15 @@ func (s *Service) UserIdentifier(ctx context.Context, req *userPb.UserIdentifier
...
 	i, err := s.m.GetByAuthToken(token)
 	if err != nil {
-		return nil, err
+		code := codes.Internal
+		if err == manager.ErrAuthTokenNotFound {
+			code = codes.PermissionDenied
+		}
+		return nil, status.Errorf(code, "getting instance: %v", err)
 	}
 	return i, nil
```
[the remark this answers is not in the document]
asutula: Yea I wanted to share that error handing logic around getting the api instance from ctx. Sorry I didn't get a chance to comment all this to explain before you read through it :/

api/server/user/storagejobs.go
```diff
@@ -53,135 +50,99 @@ func (s *Service) StorageConfigForJob(ctx context.Context, req *userPb.StorageCo
...
-		LatestSuccessfulStorageJobs: protoLatestSuccessfulJobs,
-		QueuedStorageJobs:           protoQueuedJobs,
+		CountQueuedStorageJobs:    int32(len(queuedJobIDs)),
+		CountExecutingStorageJobs: int32(len(executingJobIDs)),
+		CountFinalStorageJobs:     int32(len(finalJobIDs)),
+		QueuedStorageJobs:         queuedJobIDs,
+		ExecutingStorageJobs:      executingJobIDs,
+		FinalStorageJobs:          finalJobIDs,
 	}, nil
 }
```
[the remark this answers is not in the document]
asutula: Yea, I agree. Will see if there is a good way to do it.

cmd/pow/cmd/admin/jobs/summary/summary.go
```diff
@@ -6,14 +6,14 @@ import (
...
-	"github.com/textileio/powergate/cmd/pow/cmd/admin/jobs/util"
+	"github.com/textileio/powergate/api/client/admin"
 	c "github.com/textileio/powergate/cmd/pow/common"
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
 func init() {
-	Cmd.Flags().StringP("user-id", "i", "", "optional instance id filter to apply")
-	Cmd.Flags().StringSliceP("cids", "c", nil, "optional cids filter to apply")
+	Cmd.Flags().StringP("user-id", "i", "", "optional user id filter to apply")
```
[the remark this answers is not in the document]
asutula: 👍

cmd/pow/cmd/storagejobs/list/list.go
```diff
@@ -0,0 +1,65 @@
...
+
+		res, err := c.PowClient.StorageJobs.List(c.MustAuthCtx(ctx), conf)
+		c.CheckErr(err)
+
+		json, err := protojson.MarshalOptions{Multiline: true, Indent: "  ", EmitUnpopulated: true}.Marshal(res)
+		c.CheckErr(err)
+
+		fmt.Println(string(json))
+	},
+}
```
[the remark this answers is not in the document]
asutula: Currently is not used. At the moment, the cli is just fetching all the data in one call. I was going to circle back to take advantage of paging asap, maybe in this pr, but maybe in an upcoming pr.

ffs/api/api_jobs.go
```diff
@@ -69,28 +69,45 @@ func (i *API) WatchJobs(ctx context.Context, c chan<- ffs.StorageJob, jids ...ff
...
+)
 
-// LatestFinalStorageJobs returns the most recent finished jobs for the specified cids.
-// If no cids are provided, data for all data cids is returned.
-func (i *API) LatestFinalStorageJobs(cids ...cid.Cid) []ffs.StorageJob {
-	return i.sched.LatestFinalStorageJobs(i.cfg.ID, cids...)
+// ListStorageJobsConfig controls the behavior for listing StorageJobs.
+type ListStorageJobsConfig struct {
+	// CidFilter filters StorageJobs list to the specified cid. Defaults to no filter.
+	CidFilter cid.Cid
```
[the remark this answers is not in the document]
asutula: The cli command is `list`, maybe I'm not understanding the problem you're describing.

api/client/admin/storagejobs.go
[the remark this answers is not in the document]
jsign: SGTM.

api/client/admin/storagejobs.go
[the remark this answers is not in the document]
jsign: Oh, I see now... misread/read_too_fast. :)

api/client/storagejobs.go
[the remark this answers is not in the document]
jsign: Yeah, not sure though. If only simplifies things. If you feel covering both cases is uglier, doesn't seem to worth it. (Since some fields doesn't make sense in other case).

api/server/admin/jobs.go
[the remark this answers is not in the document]
jsign: Yep.

ffs/api/api_jobs.go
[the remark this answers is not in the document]
jsign: Forget about it. It's the same confusion of misreading the comment above.

ffs/scheduler/internal/sjstore/sjstore.go
jsign: Dismiss this comment.

ffs/scheduler/internal/sjstore/sjstore.go
asutula: A couple thoughts there.

- This is a query of only the index data, so yes, to order it has to load all the data, but at least it's just simple index data and not the actual data.
- I wanted to make descending ordering the default since I think that is the most common way to look at the results. "show me what is going on recently" way of thinking. So to support that, no default lexicographic order would work unfortunately. Plus:
- I don' think these index keys actually sort ascending lexicographically by time because the api id and cid comes before the timestamp in the key like: `apiid/<apiid>/<cid>/<timestamp>. Right?

asutula: @jsign, I'm still considering using in memory maps instead of go-datstore indexes, but in general do you view either of those to be a workable solution or are you viewing the ordering issue or the way these indexes work to be a no-go thing that needs to change more fundamentally?

ffs/scheduler/internal/sjstore/sjstore.go
asutula: Ah yea, in Go an empty `switch` `case` will break out of they switch by default, right?

ffs/scheduler/internal/sjstore/sjstore.go
jsign: - 👍🏼 
- 👍🏼 
- But in some of your queries you're doing `prefix = prefixAPIIDAndCid(config.APIIDFilter, config.CidFilter)`, wouldn't those keys lexicographic order should only be determined by the last part of the key? (timestamp) Since APIID and Cid are fixed. Recall that I mentioned that in "some case" you could at least leverage this. Missing something?

ffs/scheduler/internal/sjstore/sjstore.go
asutula: right, leftover from some iteration of coding this up. removing.

ffs/scheduler/internal/sjstore/sjstore.go
asutula: Yea feels so hard to do something simple. I'd like to start thinking ahead more during the design phase of new code and start with indexes from the beginning in new features.

ffs/scheduler/internal/sjstore/sjstore.go
jsign: > @jsign, I'm still considering using in memory maps instead of go-datstore indexes, but in general do you view either of those to be a workable solution or are you viewing the ordering issue or the way these indexes work to be a no-go thing that needs to change more fundamentally?

I think your implemented idea is close to what I'd do with go-datastore. So I'm not much concerned about the design or implementation; the only ugly thing is that you'd need migration to re-create those indexes now. And also, I was most concerned about some code complexity and not sure about how bad would be to do this with go-datastore (`Order` field, etc).

So to be clear, I think the "cons" aren't really in your implementation or design (I think that's OK). But mostly to some unavoidable complexity or question-marks around the cost of all this (being in complexity or perf) caused by go-datastore API.

ffs/scheduler/internal/sjstore/sjstore.go
jsign: > Ah yea, in Go an empty `switch` `case` will break out of they switch by default, right?

[Yes](https://play.golang.org/p/I4BogTjX3Nl). 
If you want to merge cases, you can do: `case value1, value2:`.

ffs/scheduler/internal/sjstore/sjstore.go
jsign: Yeah, totally agree.

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -660,3 +752,27 @@ func makeStartedDealsKey(iid ffs.APIID, c cid.Cid) datastore.Key {
...
+	return dsBaseAPIID.ChildString(APIID.String()).String()
+}
+
+func prefixCidAndAPIID(cid cid.Cid, APIID ffs.APIID) string {
+	return dsBaseCid.ChildString(cid.String()).ChildString(APIID.String()).String()
+}
+
+func prefixAPIIDAndCid(APIID ffs.APIID, cid cid.Cid) string {
+	return dsBaseAPIID.ChildString(APIID.String()).ChildString(cid.String()).String()
+}
```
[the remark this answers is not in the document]
asutula: Sounds like a good idea. Will do.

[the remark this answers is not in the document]
asutula: Here's the comment I'm adding. It's actually a little different than what you typed, but let me know if you think the keys should change.

```
/**
There are many namespaces that are maintained in the contained datastore.
Use these descriptions to understand what information is stored where.
Might be useful in a future migration or datastore maintenance.

/job/<job-id>: Stores StorageJob data by job-id
/apiid/<api-id>/<cid>/<timestamp>: Index on api-id primarily, cid secondarily, with timestamp, values of job-id
/cid/<cid>/<api-id>/<timestamp>: Index on cid primarily, api-id secondarily, with timestamp, values of job-id
/starteddeals_v2/<instance-id>/<cid>: Stores StartedDeals data by instance-id and cid
*/
```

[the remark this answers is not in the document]
jsign: Cool! Yeah, most of what I wrote was a quick example. 
That looks great.

ffs/scheduler/internal/sjstore/sjstore.go
asutula: Re: Query prefixing and lexicographic order... Right, it would only work for a subset of queries that specify both api id and cid. If you specified only api id or only cid, you'd have to use `Order` still.

Thanks for clarifying what you think the "cons" are and what that all means.

asutula: > If you specified only api id or only cid or neither, you'd have to use Order still
Well, not quite true... I could also add two more indexes: `/apiid-only/<api-id>/timestamp` and `/cid-only/<cid>/timestamp`, but I was trying to find some optimal situation where the number of indexes wasn't crazy and there are still tools available to query the data in the required ways with acceptable performance. Plus the whole descending order by default thing removes value from making all possible indexes that will support ascending lexicographic order.

jsign: Yeah, all things can be solved with more indexes.
My point in my original thread comment was that there was a special case of your if-cases that can avoid the `Order`.

I'm totally against adding complexity in general (in this case, adding more indices). 
I just mentioned a special-case opportunity to make it faster without adding new indexes or wild stuff.

asutula: Gotcha, thanks.

api/client/admin/storagejobs.go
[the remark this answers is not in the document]
asutula: I'm going to hold of on adding a default limit to the client until a later pr where we actually add support for processing paged data to the cli. Don't feel like biting that off right now.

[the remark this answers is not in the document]
asutula: https://github.com/textileio/powergate/issues/749

migration/migration3.go
```diff
@@ -0,0 +1,53 @@
...
+type storageJob struct {
+	ID        string
+	APIID     string
+	Cid       cid.Cid
+	CreatedAt int64
+}
+
+// V3StorageJobsIndexMigration contains the logic to upgrade a datastore from
+// version 2 to version 3.
+var V3StorageJobsIndexMigration = Migration{
```
asutula, line 21: Giving this all the migration "3" naming so it doesn't conflict with migration 2 that already exists in `jsign/import`... will make my pending rebase easier.

migration/migration3.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/query"
+)
+
+type storageJob struct {
```
asutula, line 12: Includes only the needed parts of `ffs.StorageJob`.

migration/migration3.go
```diff
@@ -0,0 +1,53 @@
...
+
+			if err := ds.Put(apiidKey, []byte(job.ID)); err != nil {
+				return fmt.Errorf("putting apiid index record in datastore: %s", err)
+			}
+			if err := ds.Put(cidKey, []byte(job.ID)); err != nil {
+				return fmt.Errorf("putting cid index record in datastore: %s", err)
+			}
+		}
+		return nil
+	},
```
asutula, line 52: Very simple migration that simply reads all saved `StorageJob`s and writes the corresponding index records for them.

migration/migration3.go
```diff
@@ -0,0 +1,53 @@
...
+	ID        string
+	APIID     string
+	Cid       cid.Cid
+	CreatedAt int64
+}
+
+// V3StorageJobsIndexMigration contains the logic to upgrade a datastore from
+// version 2 to version 3.
+var V3StorageJobsIndexMigration = Migration{
+	UseTxn: true,
```
asutula, line 22: I don't think this is a "big" migration, so seems like this is a good idea.

migration/migration3_test.go
```diff
@@ -0,0 +1,23 @@
...
+	ds := tests.NewTxMapDatastore()
+
+	pre(t, ds, "testdata/v3_StorageJobs.pre")
+	txn, _ := ds.NewTransaction(false)
+
+	err := V3StorageJobsIndexMigration.Run(txn)
+	require.NoError(t, err)
+	require.NoError(t, txn.Commit())
+
+	post(t, ds, "testdata/v3_StorageJobs.post")
```
asutula, line 22: Not much to test here other than the data is what we expect according to the csv files.

migration/migration_test.go
```diff
@@ -155,6 +155,7 @@ func TestRealDataBadger(t *testing.T) {
 
 	migrations := map[int]Migration{
 		1: V1MultitenancyMigration,
+		2: V3StorageJobsIndexMigration,
```
asutula, line 158: Will merge this with `jsign/import` and this migration will be key `3`.

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -134,30 +154,39 @@ func (s *Store) Finalize(jid ffs.JobID, st ffs.JobStatus, jobError error, dealEr
...
 }
 
 // Dequeue dequeues a Job which doesn't have have another Executing Job
 // for the same Cid. Saying it differently, it's safe to execute. The returned
-// job Status is automatically changed to Executing. If no jobs are available to dequeue
+// job Status is automatically changed to Executing. If an instance id is provided,
+// only a job for that instance id will be dequeued. If no jobs are available to dequeue
 // it returns a nil *ffs.Job and no-error.
-func (s *Store) Dequeue() (*ffs.StorageJob, error) {
+func (s *Store) Dequeue(iid ffs.APIID) (*ffs.StorageJob, error) {
```
asutula, line 168: FYI @jsign the new aphid arg to `Dequeue`

ffs/scheduler/internal/sjstore/sjstore_test.go
```diff
@@ -1,21 +1,350 @@
...
+	executing []ffs.StorageJob
+	final     []ffs.StorageJob
+}
+type test struct {
+	name    string
+	apiJobs []apiJobs
+	cb      func(t *testing.T, s *Store, createdJobs ...createdJobs)
+}
+
+var testsList = []test{
```
asutula, line 35: Also FYI @jsign here's a declarative test setup I'm using to test `List`.

ffs/scheduler/internal/sjstore/sjstore_test.go
```diff
@@ -71,23 +400,39 @@ func TestDequeue(t *testing.T) {
...
+		require.NoError(t, err)
+
+		j2 := createJob(t, "apiid2", cid.Undef)
+		err = s.Enqueue(j2)
+		require.NoError(t, err)
+
+		j, err := s.Dequeue(ffs.APIID("apiid2"))
+		require.NoError(t, err)
+		require.Equal(t, ffs.APIID("apiid2"), j.APIID)
+	})
```
asutula, line 424: New test to test dequeuing a job for a specific api id.

PR #755, opened 2021-01-19
migrtest & V3 migrations improvements

jsign:
This PR:
- Creates a new tool `migrtest` to test migrations in a safe way with real data. See rendered [README](https://github.com/textileio/powergate/blob/96d7fb09c168fac868461b1d731cb3fbe25dab4e/cmd/migrtest/README.md) to have a gist of it.
- Does some improvements to V3 migration.

While doing this PR, `dscopy` from `go-thread` was extracted into https://github.com/textileio/dsutils (cc @sanderpick ). This new repo can have useful code to do this kind of tool in other repos. For example, I'd like to add some package to do some diffs comparing two databases there, so we can start having useful shared tools between repos.


migration/migration3.go
```diff
@@ -19,7 +20,7 @@ type storageJob struct {
 // V3StorageJobsIndexMigration contains the logic to upgrade a datastore from
 // version 2 to version 3.
 var V3StorageJobsIndexMigration = Migration{
-	UseTxn: true,
+	UseTxn: false,
```
jsign, line 23: Since this is parallelized now, it can't run in a txn. It was also quite slow to run it serially on big hosted-pows (e.g: landsatdata).

migration/migration3.go
```diff
@@ -28,26 +29,64 @@ var V3StorageJobsIndexMigration = Migration{
 		}
 		defer func() { _ = res.Close() }()
```
jsign, line 31: Nothing fancy, same idea I did in other existing migrations.

cmd/migrtest/main.go
```diff
@@ -0,0 +1,94 @@
+package main
```
jsign, line 1: I think it's quite clear.
It relies in a new repo as mentioned in PR description.

asutula (approved):

PR #774, opened 2021-02-12
Add Message CID to SendFilResponse

hoffmabc:
This would allow clients to retrieve the message CID for sending FIL.


asutula:
Great @hoffmabc, I've been wanting to make this addition as well. Just just last week started using the `develop` branch for all new features... Can you rebase your branch against `develop` and change the target branch for this PR then we can move forward? Thanks!

hoffmabc:
Alright @asutula I rebased and put it against /develop.

asutula:
Looking great! Only think missing I noticed is update the cli `send` commands in the user and admin parts of the cli `pow` subcommands. Those commands currently don't print any output... would be useful to print the cid, right?

hoffmabc:
Good suggestion. I'll add it.

asutula:
Quickest way is just to print out the json representation of the client response. You could steal that code from any number of other commands that do that.

asutula (approved):
Looks great. Thanks @hoffmabc!

PR #775, opened 2021-02-18
Update to Lotus v1.4.2

jsign:
- Uses new Lotus client.
- Uses newly created lotus-devnet also running v1.4.2
- Uses newly created Lotus docker image v1.4.2.


textileben (approved):
LGTM

PR #784, opened 2021-02-26
New index data & Lotus v1.5.0 updating & other improvements

jsign:
This PR includes:
- A new flag to indicate if the miner's index wants to be refreshed on Powergate start (default: false).
- Remove "Online" discovery for miners. This wasn't used and can add too much load.
- Re-enable location resolving for miners in the miner's metadata subindex.
- Include (Active/Live/Faulty) sector counts in miner's index, as to have a metric of sector counts of miners which is helpful to imagine their size and involvement in the network. 
- Include the new index data in the indices API.
- For storage-deal records information, the deal information is _always_ refreshed, not only when the deal becomes active. This can leave more useful data even if the deal fails.
- Bug fix regarding adding awareness of pending storage-deal records becoming final in the updated-at internal index.
- Removed CI workflows and dependabot referencing removed `develop` branch.
- Update Lotus client to v1.5.0
- Update localnet to newly created v1.5.0.
- Reference newly created Lotus v1.5.0 docker-image.
- Update dependencies.


asutula (approved):
Looks great!

PR #791, opened 2021-03-03
Move to not using multiaddr for config values

asutula:
Per a conversation today with @sanderpick and @textileben . So far, I've only update the lotus address to be plain old `host:port`, but we can take this all the way. Let me know what you think.

### DIFF
--- a/api/client/utils_test.go
+++ b/api/client/utils_test.go
@@ -32,7 +32,7 @@ func defaultServerConfig(t *testing.T) server.Config {
 	ipfsAddr := util.MustParseAddr(ipfsAddrStr)
 
 	devnet := tests.LaunchDevnetDocker(t, 1, 300, ipfsAddrStr, false)
-	devnetAddr := util.MustParseAddr("/ip4/127.0.0.1/tcp/" + devnet.GetPort("7777/tcp"))
+	devnetAddr := "127.0.0.1/tcp/" + devnet.GetPort("7777/tcp")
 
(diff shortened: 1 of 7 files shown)


api/client/utils_test.go
```diff
@@ -32,7 +32,7 @@ func defaultServerConfig(t *testing.T) server.Config {
 	ipfsAddr := util.MustParseAddr(ipfsAddrStr)
 
 	devnet := tests.LaunchDevnetDocker(t, 1, 300, ipfsAddrStr, false)
-	devnetAddr := util.MustParseAddr("/ip4/127.0.0.1/tcp/" + devnet.GetPort("7777/tcp"))
+	devnetAddr := "127.0.0.1/tcp/" + devnet.GetPort("7777/tcp")
```
jsign, line 35: I guess this should be:
`devnetAddr := "127.0.0.1:" + devnet.GetPort("7777/tcp")`

jsign (approved):
Sounds good.

But maybe we should hold merging this until a next major version bump since this in theory is a breaking change. Not in the APIs, but in the flag/envs that people might already have and are forced to change.

api/client/utils_test.go
asutula: 👍🏼

asutula:
@jsign I realized there were other places, like docs strings and readme that needed to be updated. May want to take another look. Sure, we can keep this on a branch for now. But once this is used in Hub, and we need to deploy hub (still might be many days or a week or two out), we'll have to release this.

Issue #793, opened 2021-03-05
Adding a small file fails. The error does not help much. 


felixniemeyer:
After setting up local net and the CLI-tool as described in the docs, adding ("applying config to") a small file as described in the docs fails.  

Reproduce: 

setup localnet and pow cli like described in the docs. Create a small file, create a user, stage the file and apply the config like this: 
```bash
echo "test" > testfile
response=`pow admin user create`
token=`jq -r '.[] | .token' <<< $response`
export POW_TOKEN=$token
response=`pow data stage testfile`
cid=`jq -r '.cid' <<< $response`
pow config apply --watch $cid
```

And it will fail with this job status: 
```
JOB_STATUS_FAILED executing cold-storage config:
 making deal configs:
 getting miners from minerselector:
 getting miners from reputation module:
 not enough miners satisfy the miner selector constraints 
```

When altering the code a bit to return an error as soon as one miner does not satisfy the constraints you learn that the piece size is too small: 
```
JOB_STATUS_FAILED executing cold-storage config:
 making deal configs:
 getting miners from minerselector:
 getting miners from reputation module:
 skipping cached miner, not replying to miner proposal query:
 miner doesn't satisfy piece size constraints:
 256<128<536870912       
```

When trying to add a bigger file, everything works. 

This may leave new users puzzled for a while.

Maybe adding a hint about the file size to the docs would be helpful enough. 
Or maybe doing something with this error in `reptop.go` [line 97](https://github.com/textileio/powergate/blob/a2048ee8d580d16b17698358d0ade3f1285150ee/ffs/minerselector/reptop/reptop.go#L98)
```go
if err != nil {
	continue // Cached miner isn't replying to query-ask now, skip.
}
```

The current handling might also be disadvantageous in other situations: if you have chosen impossible miner requirements, it would be good to know, what requirements exactly is hard to satisfy.

I have a solution in mind and will follow up with a pull request.

jsign:
Hey @felixniemeyer, thanks for the detailed explanation.

This is already explained in the docs [here](https://docs.textile.io/powergate/localnet/#localnet-features):
> Not using --bigsectors will limit you to storing files of around 700 bytes and deals will complete in 30-60 seconds. Using --bigsectors will allow you to store files anywhere from 1Mb to 400Mb, but deals will complete in 3-4 minutes. Be sure to choose the value that makes sense for your development scenario.

[...]

PR #794, opened 2021-03-05
better error message when no miner satisfies selector constraints

felixniemeyer:
solves this issue: 
https://github.com/textileio/powergate/issues/793


jsign (approved):
Thanks!

PR #795, opened 2021-03-08
Include verified-clients and remaining data-cap

jsign:
WIP


ffs/filcold/filcold.go
```diff
@@ -130,6 +134,19 @@ func (fc *FilCold) Store(ctx context.Context, c cid.Cid, cfg ffs.FilConfig) ([]c
...
+			return nil, nil, 0, fmt.Errorf("get address verified client info: %s", err)
+		}
+		if err == wallet.ErrNoVerifiedClient {
+			return nil, nil, 0, fmt.Errorf("wallet address isn't a verified client")
+		}
+		pieceSizeBig := big.NewInt(int64(pieceSize))
+		if pieceSizeBig.Cmp(vci.RemainingDatacapBytes) == 1 {
+			return nil, nil, 0, fmt.Errorf("the remaining data-cap %d MiB is less than piece-size %d MiB", pieceSize/1024/1024, vci.RemainingDatacapBytes.Uint64()/1024/1024)
+		}
+	}
```
jsign, line 149: Fail-fast logic to unverified or not-enough-data-cap situations.

proto/powergate/user/v1/user.proto
```diff
@@ -322,6 +322,11 @@ message AddrInfo {
   string address = 2;
   string type = 3;
   string balance = 4;
+  VerifiedClientInfo verified_client_info = 5;
+
+  message VerifiedClientInfo {
+    string remaining_datacap_bytes = 1;
```
jsign, line 328: This is `string` since the native type is `BigInt`, so in some extreme case, the user could have a lot of verified deal capacity.

PR #811, opened 2021-03-25
OpenTelemetry metrics & Miner Index fix for document max-size limit & new deal-watcher

jsign:
This PR includes multiple improvements:
- The Miner Index was refactored, and new flags/envs can be tuned to: refresh the index on Powergate start, tune the level of parallelization to speed up rebuilding the index, and the frequency of index refresh.
- Any `go-datastore` implementation is wrapped to a cusotmized `go-ds-measure` to publish Prometheus metrics of Get, Put, Query, Commit, and other usual operations of `go-datastore`. We forked the original repo to add Txn based metrics.
- Opencensus was removed, and was replaced with Opentelemetry.
- The miner index saving switched from being saved as a complete snapshot, to being saved per miner. This was to fix a limitation of MongoDB backed `go-datastore` not allowing to save documents bigger than 16MB.
- A new `DealWatcher` component was created to provide an instant-based solution to detect deal state changes. The existing poll-style is kept because we aren't totally sure how reliable this new way is. In fact, this was done originally and we switched to poll-based since long-running websocket connections of `go-jsonrpc` for the Lotus client were unreliable. Keeping both allows to get the benefits of instant updating, but if for some reason that breaks, we still do polling once in a while. (Apart from trying to recover that broken watcher).

A new set of application metrics are exposed:
- General metrics about the build:
  - Build date
  - Git summary
  - Git branch
  - Git commit
Datastore:
  - All mentioned operations of `go-datastore` latencies.
- Deal watcher metrics:
  - How many deal updates we get from the chain
  - How many times the watcher got in bad health and had to be re-created. This helps to understand how reliable this new mechanism is.
- Deal module:
  - A counter of how many deals are being actively tracked (in-progress)
  - Same as above but for retrievals
  - A counter of the number of finalized deals, tagged with success or error.
  - A counter of the number of bytes of finalized deals, tagged with success or error.
- Cold-storage:
  - A gauge to understand the data pre-processing queue (deal size + CommP calculation), tagged with "waiting" and "in-progress". (Remember that we rate-limit by default to 2 "in-progress" since it's a heavy process to avoid killing the Lotus node)
- Scheduler:
  - A counter to show the Jobs in final status, each tagged with "success", "failed", "canceled".
  - A gauge to show Jobs in "queued" and "executing" status.
- Indices:
  - Ask and Miner index progress gauge. (0%/100%
- Wallet module:
  - A counter of the number of created wallet addresses
  - A histogram of wallet transfer amounts.

I believe there's a lot of room to continue thinking if these are the right metrics or other ones that are useful to be gathered. The main point here was doing some kickoff and we can keep making some incremental improvements not that we have things wired.

Here's a quick screenshot of the Grafana dashboard. It has some rough edges, but we'll eventually also publish in [Grafana dashboards](https://grafana.com/grafana/dashboards):
image

While doing this work, some secondary repos were forked+edited and created:
- https://github.com/textileio/go-ds-measure/: Forked from original and added txn related metrics.
- https://github.com/textileio/go-metrics-opentelemetry: New repo that provides an implementation of http://github.com/ipfs/go-metrics-interface, which is the interface used in go-ds-measure (since it uses a generic interface for metrics collection).


index/miner/lotusidx/onchain.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return fmt.Errorf("getting chain head: %s", err)
+	}
+	newIndex.LastUpdated = int64(chainHead.Height())
+
+	mi.lock.Lock()
+	mi.index.OnChain = newIndex
+	mi.lock.Unlock()
+
+	if err := mi.store.SaveOnChain(ctx, newIndex); err != nil {
```
jsign, line 49: Some code moving here, and removing som old style of saving on-chain data which had problems to saving when using MongoDB as the backing store for go-datastore.

Before, we simply saved the whole miner index as a single JSON in go-datastore. But MongoDB has a 16MB limit on document sizes, so Powergate failed to saved the index. This PR created a new `store` for the miner-index which saves Miner Index information per-miner, so avoiding big documents.

index/miner/lotusidx/store/store.go
```diff
@@ -0,0 +1,189 @@
+package store
```
jsign, line 1: So here's the new `store` that the Miner Index uses. Before, there wasn't a separate `store` package because the persistence was made in the same package as the miner index. So I took advantage and moved the persistence logic to this own package (as we usually do for better organization).

index/miner/lotusidx/store/store.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		if err := s.ds.Put(key, buf); err != nil {
+			return fmt.Errorf("saving metadata in store: %s", err)
+		}
+	}
+	return nil
+}
+
+// SaveOnChain creates/updates on-chain information of miners.
+func (s *Store) SaveOnChain(ctx context.Context, index miner.ChainIndex) error {
```
jsign, line 52: The miner index has a lot of miners. I used a `datastore.Batching` interface to allow to persist things in batches since single updates were too slow.

index/miner/lotusidx/store/store.go
```diff
@@ -0,0 +1,189 @@
...
+		i++
+		if i%1000 == 0 {
+			if err := b.Commit(); err != nil {
+				return fmt.Errorf("committing batch: %s", err)
+			}
+			b, err = s.ds.Batch()
+			if err != nil {
+				return fmt.Errorf("creating batch: %s", err)
+			}
+		}
```
jsign, line 68: Every 1000 miner index entries, we commit the batch and start a new one.

migration/migration5.go
```diff
@@ -0,0 +1,33 @@
...
+		for v := range res.Next() {
+			if err := ds.Delete(datastore.NewKey(v.Key)); err != nil {
+				return fmt.Errorf("deleting miner chainstore key: %s", err)
+			}
+			count++
+		}
+		log.Infof("deleted %d chainstore keys", count)
+
+		return nil
+	},
```
jsign, line 32: The new miner-index store saves the data in a new namespace. The only thing this migration is doing is removing the old persisted miner-index data. If wasn't removed it would stay there forever. Isn't really that big, but doing it as to keep things clean.

eightysteele:
> Any go-datastore implementation is wrapped to a cusotmized go-ds-measure to publish Prometheus metrics of Get, Put, Query, Commit, and other usual operations of go-datastore. We forked the original repo to add Txn based metrics.

So good!

Curious if any of the refactoring here makes it easier to eventually add in some rolling averages stuff (textileio/textile#519).

jsign:
> > Any go-datastore implementation is wrapped to a cusotmized go-ds-measure to publish Prometheus metrics of Get, Put, Query, Commit, and other usual operations of go-datastore. We forked the original repo to add Txn based metrics.
> 
> So good!
> 
> Curious if any of the refactoring here makes it easier to eventually add in some rolling averages stuff ([textileio/textile#519](https://github.com/textileio/textile/issues/519)).

The Hub Miner Index is built with `mindexd` [daemon](https://github.com/textileio/textile/tree/master/api/mindexd). This was done to allow the Hub to decide its own path on what the Hub Miner Index should look like. (Also, to gather deal-making information from all our hosted powergates, so it has more information than any powergate too. But this fact is unrelated to that feature).

That daemon is already querying Powergate on regular intervals about prices offered by miners, so it can already do further aggregations (being rolling averages, or other things) on its own. It would need to keep some short history of the last ask prices, and do more calculation on tops or similar.

eightysteele:
Gotcha, that makes sense 👍🏻 (thanks!)

wallet/lotuswallet/wallet.go
```diff
@@ -18,6 +18,7 @@ import (
 	logger "github.com/ipfs/go-log/v2"
```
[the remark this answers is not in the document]
asutula: Yea very interesting.

asutula (approved):
Looks great. No issues at all, just some general questions for my own curiosity.

deals/module/retrieve.go
```diff
@@ -103,6 +103,9 @@ func (m *Module) retrieve(ctx context.Context, lapi *apistruct.FullNodeStruct, l
 	go func() {
 		defer lapiCls()
 		defer close(out)
+		m.metricRetrievalTracking.Add(ctx, 1)
+		defer m.metricRetrievalTracking.Add(ctx, -1)
+
```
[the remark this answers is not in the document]
jsign: Yeah, that could be possible. I'd vote to let the daemon itself publish the metric instead of being an external component, just at to make metrics a baked-in feature of the daemon.

index/miner/lotusidx/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+
+	_ = metric.Must(meter).NewFloat64ValueObserver("powergate.index.miner.progress", mi.progressValueObserver, metric.WithDescription("Miner index refresh progress"), metric.WithUnit(unit.Dimensionless))
+	mi.meterRefreshDuration = metric.Must(meter).NewInt64ValueRecorder("powergate.index.miner.refresh.duration", metric.WithDescription("Refresh duration"), metric.WithUnit(unit.Milliseconds))
+}
+
+func (mi *Index) progressValueObserver(ctx context.Context, result metric.Float64ObserverResult) {
+	mi.metricLock.Lock()
+	defer mi.metricLock.Unlock()
+	result.Observe(mi.onchainProgress, onchainSubindex)
+	result.Observe(mi.metaProgress, metaSubindex)
```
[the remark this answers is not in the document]
jsign: This metric is defined as:
```
_ = metric.Must(meter).NewFloat64ValueObserver("powergate.index.ask.progress", ai.progressValueObserver, metric.WithDescription("Ask index refresh progress"), metric.WithUnit(unit.Dimensionless))
```
So what happens in the `NewXXXXObserver` style of metrics, is that the OpenTelemetry library will call the callback (`progressValueObserver`) on defined intervals.
This has some benefits of not needing to record quickly changing metrics, and also registering a ton of observations within the same timestamp.

In this case, since the update progress is a gauge, I let this metric be updated on every collection interval.

There are quite a few things to learn, and keep improving. I think [this](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#interpretation) is the best reference as to understand how different measurements work. I still need to give a full read, but that's the best source of documentation for the metrics API.

PR #820, opened 2021-04-07
Offline data preparation tools

jsign:
This PR adds new Powergate CLI commands to prepare data to make offline deals. 

We'll soon add offline-deal support in the Powergate APIs, but these preparing tools are useful to prepare data for this stage without having to run a Lotus or go-ipfs daemon. Also, the prepared data can be used with any Filecoin client.

Might be helpful to read the docs PR that explains how to use these new features: https://github.com/textileio/docs/pull/281

Also, it updates the Lotus client, devnet, and docker-image to v1.6.0.


go.mod
```diff
@@ -7,40 +7,51 @@ require (
...
 	github.com/grpc-ecosystem/go-grpc-middleware v1.2.2
 	github.com/improbable-eng/grpc-web v0.14.0
+	github.com/ipfs/go-blockservice v0.1.4
+	github.com/ipfs/go-car v0.0.4
 	github.com/ipfs/go-cid v0.0.7
 	github.com/ipfs/go-datastore v0.4.5
+	github.com/ipfs/go-ds-badger v0.2.6
 	github.com/ipfs/go-ds-badger2 v0.1.1-0.20200708190120-187fc06f714e
+	github.com/ipfs/go-graphsync v0.7.0 // indirect
+	github.com/ipfs/go-ipfs v0.8.0
```
jsign, line 38: I might regret of adding this dependency... mostly because of indirect dependencies that also Louts brings in.
If at some point in the future this is too much of a mess, we might need to extract the subcommand into some other binary with it's own mod. 

Doesn't seems like a problem now.

sanderpick: At some point, the adder logic could probably be pulled out of `coreunix.NewAdder`, I think that's what I did for the local bucket repo stuff... but yeah, nbd.

sanderpick (approved):
Nice! LGTM

go.mod
jsign: Yeah, actually go-ipfs end up using that package so can be narrowed down.

PR #848, opened 2021-07-02
pow: support offline Filecoin batch preparation

jsign:
This PR adds support for `pow offline prepare --aggregate` which takes a folder as an argument and:
- Dagifies all the files
- Batches them in a Filecoin batch as described by [this]() spec.
- Calculates the PieceCid and PieceSize.

And outputs:
- Prints in standard output (can use `--json` for json formatting; highly recommended):
   - The PayloadCid, PieceCid, and PieceSize.
   - Each processed file with their corresponding Cid.
- Generates the following assets in the current folder:
   - The CAR file of the UnixFS batch.
   - The manifest file that was generated and lives inside the batch as described by the spec.

### DIFF
--- a/cmd/pow/cmd/prepare/prepare.go
+++ b/cmd/pow/cmd/prepare/prepare.go
@@ -12,8 +12,12 @@ import (
 	"sync"
 	"time"
 
+	files "github.com/ipfs/go-ipfs-files"
+	unixfile "github.com/ipfs/go-unixfs/file"
+
 	"github.com/cheggaaa/pb/v3"
 	"github.com/dustin/go-humanize"
+	aggregator "github.com/filecoin-project/go-dagaggregator-unixfs"
 	bsrv "github.com/ipfs/go-blockservice"
 	"github.com/ipfs/go-car"
 	"github.com/ipfs/go-cid"
@@ -36,6 +40,7 @@ func init() {
 	prepare.Flags().String("tmpdir", os.TempDir(), "path of folder where a temporal blockstore is created for processing data")
 	prepare.Flags().String("ipfs-api", "", "IPFS HTTP API multiaddress that stores the cid (only for Cid processing instead of file/folder path)")
 	prepare.Flags().Bool("json", false, "avoid pretty output and use json formatting")
+	prepare.Flags().Bool("aggregate", false, "aggregates a folder of files")
 
 	commp.Flags().Bool("json", false, "avoid pretty output and use json formatting")
 	commp.Flags().Bool("skip-car-validation", false, "skips CAR validation when processing a path")
@@ -86,7 +91,7 @@ If a Cid is provided, an extra --ipfs-api flag should be provided to connect to 
 		if err != nil {
 			c.Fatal(fmt.Errorf("parsing json flag: %s", err))
 		}
-		dataCid, dagService, cls, err := prepareDAGService(cmd, args, quiet)
+		dataCid, dagService, _, cls, err := prepareDAGService(cmd, args, quiet)
 		if err != nil {
 			c.Fatal(fmt.Errorf("creating dag-service: %s", err))
 		}
@@ -147,7 +152,7 @@ You can use the --skip-car-validation, but usually shouldn't be done unless you 
 			c.Fatal(fmt.Errorf("parsing json flag: %s", err))
 		}
 		if jsonFlag {
-			printJSONResult(pieceSize, pieceCID)
+			printJSONResult(pieceSize, cid.Undef, pieceCID, nil)
 			return
 		}
 		c.Message("Piece-size: %d (%s)", pieceSize, humanize.IBytes(pieceSize))
@@ -183,7 +188,7 @@ The piece-size and piece-cid are printed to stderr. For scripting usage, its rec
 		if err != nil {
 			c.Fatal(fmt.Errorf("parsing json flag: %s", err))
 		}
-		dataCid, dagService, cls, err := prepareDAGService(cmd, args, json)
+		payloadCid, dagService, aggrFiles, cls, err := prepareDAGService(cmd, args, json)
 		if err != nil {
 			c.Fatal(fmt.Errorf("creating temporal dag-service: %s", err))
 		}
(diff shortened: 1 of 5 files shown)


cmd/pow/cmd/prepare/prepare.go
```diff
@@ -36,6 +41,7 @@ func init() {
 	prepare.Flags().String("tmpdir", os.TempDir(), "path of folder where a temporal blockstore is created for processing data")
 	prepare.Flags().String("ipfs-api", "", "IPFS HTTP API multiaddress that stores the cid (only for Cid processing instead of file/folder path)")
 	prepare.Flags().Bool("json", false, "avoid pretty output and use json formatting")
+	prepare.Flags().Bool("aggregate", false, "aggregates a folder of files")
```
jsign, line 44: New flag that indicates the files in the provided folder should be aggregated with the aggregator library.

cmd/pow/cmd/prepare/prepare.go
```diff
@@ -183,7 +189,19 @@ The piece-size and piece-cid are printed to stderr. For scripting usage, its rec
...
+		aggregate, err := cmd.Flags().GetBool("aggregate")
+		if err != nil {
+			c.Fatal(fmt.Errorf("aggregate flag: %s", err))
+		}
+		ipfsAPI, err := cmd.Flags().GetString("ipfs-api")
+		if err != nil {
+			c.Fatal(fmt.Errorf("get api addr flag: %s", err))
+		}
+		if ipfsAPI != "" && aggregate {
+			c.Fatal(errors.New("the --aggregate flag can't be used with a remote go-ipfs node"))
```
jsign, line 201: At least for now... we could keep working to support it.

cmd/pow/cmd/prepare/prepare.go
```diff
@@ -248,64 +266,99 @@ The piece-size and piece-cid are printed to stderr. For scripting usage, its rec
...
+
+			}
+			defer func() {
+				if err := fManifest.Close(); err != nil {
+					c.Fatal(fmt.Errorf("closing manifest file: %s", err))
+				}
+			}()
+			if _, err := io.Copy(fManifest, manifest); err != nil {
+				c.Fatal(fmt.Errorf("writing manifest file: %s", err))
+			}
```
jsign, line 297: This code section is to retrieve from the generated batch the manifest file that lives inside.
The manifest file lives as the first file in the root folder (defined in the spec), thus `Links()[0]` on 275 will always succeed and give us the file.

The rest is pulling the file and writing it to the current folder.

cmd/pow/cmd/prepare/prepare.go
```diff
@@ -395,10 +454,44 @@ func dagify(ctx context.Context, dagService ipld.DAGService, path string, quiet
...
+				RootCid: dcid,
+			})
+			aggregatedFiles = append(aggregatedFiles, aggregatedFile{Name: p, Cid: dcid.String()})
+			return err
+		})
+		dataCid, err = aggregator.Aggregate(ctx, dagService, lst)
+		if err != nil {
+			return cid.Undef, nil, fmt.Errorf("aggregating: %s", err)
+		}
 	}
```
jsign, line 489: If `--aggregate` is provided, we walk the folder, ingest each file, and later aggregate it.

cmd/pow/cmd/prepare/prepare.go
```diff
@@ -248,64 +266,99 @@ The piece-size and piece-cid are printed to stderr. For scripting usage, its rec
...
+			manifestNd, err := dagService.Get(ctx, manifestLnk.Cid)
+			if err != nil {
+				c.Fatal(fmt.Errorf("get manifest node: %s", err))
+			}
+			manifestF, err := unixfile.NewUnixfsFile(context.Background(), dagService, manifestNd)
+			if err != nil {
+				c.Fatal(fmt.Errorf("get unifxfs manifest file: %s", err))
+			}
+			manifest := manifestF.(files.File)
+			fManifest, err := os.Create(args[1] + ".manifest")
```
merlinran, line 285: Is it possible that we generate the manifest file on the fly in `dagify` after walked the directory, without writing anything to the local directory?

jsign: Actually, I'm writing it to the local directory as a feature. 
The manifest file is generated on the fly by the aggregator library and included in the batch UnixFS DAG.

Also, the `dagify` is per file and the manifest is a _batch scoped_ concept.

go.mod
```diff
@@ -81,4 +83,6 @@ require (
 	nhooyr.io/websocket v1.8.6 // indirect
 )
 
+replace github.com/ipfs/go-unixfs => github.com/ipfs/go-unixfs v0.2.2
```
merlinran, line 86: Why this instead of setting the version directly?

jsign: Becase a lot of libs in this ecosystem use v0. That means that the dependency graph is very brittle when you include a new dependency that ends up using a newer version of an indirect dependency considering how Go MVS version resolving works.

A lot of libs in the IPFS ecosystem that are v0 make breaking changes in further v0.. and that doesn't play well with Go MVS (which selects the maximum referenced minor version of a major version). 

Yes... horrible. Usually you have some luck by trying to upgrade other deps hoping that they will stop reference the old v0 version that is causing this needed downgrade, but that's not usually an easy task.

cmd/pow/cmd/prepare/prepare.go
merlinran: Ah gotcha. I confused the order ;)

go.mod
jsign: Long story short, if you don't use strict semantic versioning in the Go ecosystem.. you'll create nightmares for your clients.

merlinran: Okay I see this is to force its direct and indirect dependencies to use this specific version instead of relying on MVS. Maybe worth a comment? Or maybe I'm ignorant enough to not understanding the intention in the first place ;)

merlinran (approved):
👍

go.mod
jsign: Yes, that's exactly the intention. We usually only use `replace` to solve MVS problems on dependencies that don't follow strict semantic versioning. So that's the default correct interpretation.

cmd/pow/cmd/prepare/prepare.go
donatocayurin: Thank you ❤️🙏

PR #884, opened 2022-05-26
Webhook notifications

vmanilo:
Draft


ffs/types.go
```diff
@@ -342,6 +343,36 @@ func (cc ColdConfig) Validate() error {
...
+type WebhookConfiguration struct {
+	Events []string
+	Alerts []*WebhookAlert
+}
+
+type WebhookAlert struct {
+	Type      string
+	Threshold string
+}
+
```
asutula: Cool, I Like the way this is all modeled.

ffs/types.go
```diff
@@ -342,6 +343,36 @@ func (cc ColdConfig) Validate() error {
...
+type WebhookConfiguration struct {
+	Events []string
+	Alerts []*WebhookAlert
+}
+
+type WebhookAlert struct {
+	Type      string
+	Threshold string
```
asutula, line 406: Curious to see what you're thinking for `Threshold` use cases. Sounds interesting.

asutula (commented):
Nice work so far. The way you've written the code is easy to follow and understand, so I don't have many comments on the code for now (this is a good thing). 

I see you injected the Notifier into the deals module. That is a fine place to start, but I'm guessing there will be a different place we may want to leverage the Notifier, or at least an additional place or two. 

If I'm remembering things correctly, the Scheduler will be a very good place to use Notifier. The Scheduler manages all things related to storing, retrieving, renewing and repairing deals. For this reason, it will be a source for many notifications. I'd imagine those events might be higher level with less detailed information, so maybe we augment those notifications with finer grained notifications from the deals module or any other module that makes sense. 

So, sounds like we should dig into the Scheduler and understand how it works and how notifications fit in.

ffs/scheduler/internal/astore/astore.go
```diff
@@ -38,6 +38,7 @@ type RetrievalAction struct {
 	Miners        []string
 	WalletAddress string
 	MaxPrice      uint64
+	Notifications []*ffs.NotificationConfig
```
asutula, line 41: I'm curious why this addition is only in `RetreivalAction` and not `StorageAction`, but maybe I'll learn this later in my review.

ffs/scheduler/internal/sjstore/sjstore.go
```diff
@@ -78,7 +81,7 @@ type watcher struct {
 }
 
 // New returns a new JobStore backed by the Datastore.
-func New(ds datastore.TxnDatastore) (*Store, error) {
+func New(ds datastore.TxnDatastore, notifier notifications.Notifier) (*Store, error) {
```
asutula, line 84: Nice. This looks like a really clean integration of notifications into the relevant places.

notifications/events.go
```diff
@@ -0,0 +1,49 @@
...
+	StorageAuctionCanceledEvent  = storageAuction + separator + canceled
+
+	// Data retrieval events
+	AllDataRetrievalEvents      = dataRetrieval + separator + all
+	DataRetrievalCreatedEvent   = dataRetrieval + separator + created
+	DataRetrievalCompletedEvent = dataRetrieval + separator + completed
+	DataRetrievalRetriedEvent   = dataRetrieval + separator + retried
+	DataRetrievalFailedEvent    = dataRetrieval + separator + failed
+	DataRetrievalCanceledEvent  = dataRetrieval + separator + canceled
+)
```
asutula, line 43: Cool way of building up this vocabulary of events. Very exciting to see them all listed here.

notifications/events.go
```diff
@@ -0,0 +1,49 @@
...
+	AllStorageDealEvents      = storageDeal + separator + all
+	StorageDealCreatedEvent   = storageDeal + separator + created
+	StorageDealCompletedEvent = storageDeal + separator + completed
+	StorageDealRetriedEvent   = storageDeal + separator + retried
+	StorageDealFailedEvent    = storageDeal + separator + failed
+	StorageDealCanceledEvent  = storageDeal + separator + canceled
+	StorageDealExpiredEvent   = storageDeal + separator + expired
+	StorageDealSlashedEvent   = storageDeal + separator + slashed
+
+	// Storage auction events
```
asutula: Auctions aren't relevant in Powergate. We have another project that uses auctions to make filecoin deals. I think at some point there was the idea of integrating that into Powergate, but it won't happen anymore. In fact, we are going to sunset the auctions project in favor of similar functionality that now exist natively in the Filecoin network.

notifications/final_job_status.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (f FinalJobStatus) FinalUpdates() bool {
+	return true
+}
+
+func (f FinalJobStatus) MatchNotificationAlert(alert *ffs.WebhookAlert) bool {
+	return false
+}
+
+func (f FinalJobStatus) MatchNotificationEvent(event string) bool {
```
asutula: Maybe we need to implement some Go linter in this project, but would be really useful to see comments on any exported type, func, var, const. Some thing are self descriptive, but some definitely are not, so comments are helpful.

notifications/notifier.go
```diff
@@ -0,0 +1,170 @@
...
+	NotifyJobUpdates(job JobUpdates)
+}
+
+type notifier struct {
+	ctx           context.Context
+	configs       *configStore
+	updates       chan JobUpdates
+	toDelete      chan ffs.JobID
+	notifications chan *notification
+}
```
asutula: Very nice design of the notifier with workers. I like how the code puts the job updates through a funnel of logic to figure out exactly who to notify. Cool stuff.

notifications/store.go
```diff
@@ -0,0 +1,42 @@
...
+	"sync"
+
+	"github.com/textileio/powergate/v2/ffs"
+)
+
+type configStore struct {
+	sync.RWMutex
+
+	configs map[ffs.JobID][]*ffs.NotificationConfig
+}
```
asutula, line 15: Will this need to write to a persistent store (database) at some point?

docker/docker-compose-localnet.yaml
```diff
@@ -8,14 +8,15 @@ services:
...
-      - 7000:7000
+      - 7001:7001
     depends_on:
       - ipfs
       - lotus
     environment:
       - POWD_DEVNET=true
       - POWD_LOTUSHOST=/dns4/lotus/tcp/7777
       - POWD_IPFSAPIADDR=/dns4/ipfs/tcp/5001
+      - POW_GATEWAYHOSTADDR=0.0.0.0:7001
```
asutula, line 19: Just curious why these changes popped up (and in `docker-compose.yaml` as well). Maybe something you needed for local testing? If it's generally useful, feel free to keep the changes.

asutula (commented):
Coming along really nicely. It seems like everywhere you're integrating the notifier into the system makes sense to me. I like the implementation of the notifier and the way you define notification types, process events, and calculate notification to send. Nice work.

ffs/scheduler/internal/astore/astore.go
vmanilo: `StorageAction` already contains `ffs.StorageConfig`, which has `Notifications []*NotificationConfig`

https://github.com/textileio/powergate/pull/884/files#diff-c9cfa7a6e3661251bc9aec6067691ec59f7f9b166127bd8757502e3b8223d0e4R141

notifications/events.go
vmanilo: ok, I'll remove auction events

notifications/final_job_status.go
vmanilo: ok, will work on adding comments

notifications/store.go
vmanilo: yes, I think at some point would be better to put it to persistent store, for the beginning I guess we can keep it in memory

docker/docker-compose-localnet.yaml
vmanilo: yes, it's just for local testing, would need to revert it back
when I tested locally found that on my laptop some battery management uses this port :)

notifications/notifier.go
vmanilo: thank you :)

ffs/scheduler/internal/sjstore/sjstore.go
vmanilo: cool :)

Issue #882, opened 2022-05-12
Unable to build powergate CLI


chaitanyaprem:
Followed the steps mentioned in https://docs.textile.io/powergate/ to clone and build Powergate CLI. 
I am facing following errors and unable to proceed. 

My environment is M1 macbook Pro with golang version `go version go1.18 darwin/amd64`

On running `make install-pow` following error is displayed. 
```
go: downloading github.com/marten-seemann/qtls-go1-17 v0.1.0-rc.1
go: downloading github.com/matttproud/golang_protobuf_extensions v1.0.1
# github.com/lucas-clemente/quic-go/internal/qtls
../../../go/pkg/mod/github.com/lucas-clemente/quic-go@v0.21.2/internal/qtls/go118.go:5:13: cannot use "quic-go doesn't build on Go 1.18 yet." (untyped string constant) as int value in variable declaration
make: *** [install-pow] Error 2
```

Thought i found a workaround mentioned here https://github.com/ipfs/go-ipfs/issues/8819 . 
Tried to proceed with it but I am running into the below error. Looks like go-ipfs version used is 0.8 which is very old and using some flags which have been removed in later dependency (quic-go).
```
CGO_ENABLED=0 go install -ldflags="-X github.com/textileio/powergate/v2/buildinfo.GitSummary=v2.6.2-6-gb82b011-dirty -X github.com/textileio/powergate/v2/buildinfo.Version=git -X github.com/textileio/powergate/v2/buildinfo.BuildDate=2022-05-12T00:08:11Z -X github.com/textileio/powergate/v2/buildinfo.GitCommit=b82b011 -X github.com/textileio/powergate/v2/buildinfo.GitBranch=master -X github.com/textileio/powergate/v2/buildinfo.GitState=dirty" ./cmd/pow
# github.com/ipfs/go-ipfs/core/node/libp2p
../../../go/pkg/mod/github.com/ipfs/go-ipfs@v0.8.0/core/node/libp2p/transport.go:21:7: undefined: quic.RetireBugBackwardsCompatibilityMode
make: *** [install-pow] Error 2
```

jsign:
The newest libp2p has a limitation (I guess is temporal) for people compiling with Go v1.18.x, which is a release from some weeks ago.

You need to compile with v1.17.x or older. (i.e: I tried with v1.17.8 now and didn't have an issue compiling)
Be sure that your git repo is clean (no changes).

chaitanyaprem:
[...]

PR #893, opened 2022-08-23
Specify which version of Go is supported

jsolly:
Add a note to the home page about installing Go. Once we have support for newer versions of Go, we will need to modify this again.

Related to https://github.com/textileio/powergate/issues/882 but does not solve it. This only updates the doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Join us on our [public Slack channel](https://slack.textile.io/) for news, discu
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


jsign:
Thanks @jsolly, do you mind taking a look at [DCO](https://github.com/textileio/powergate/pull/893/checks?check_run_id=7979218702) to fix it?

jsolly:
@jsign I fixed the DCO issue!

jsign (approved):
LGTM!

